PR #3, opened 2015-05-26
Fixed endless cycle on Windows

Bobris:
in defaultExternalResolver


Rich-Harris:
Excellent, thanks! At some point I'll get round to setting up AppVeyor so these things get caught automatically...

Will be releasing a new version soon, this fix will be included

PR #6, opened 2015-05-27
Partially fix tests on Windows

Bobris:
There are still unfortunate tests which uses path.relative and asserting result :-( which of course also fails. One of them:
rollup/test/function/shadowed-external-export

### DIFF
--- a/test/function/custom-path-resolver-async/_config.js
+++ b/test/function/custom-path-resolver-async/_config.js
@@ -1,7 +1,7 @@
 var assert = require( 'assert' );
 
 module.exports = {
-	description: 'uses a custom path resolver (synchronous)',
+	description: 'uses a custom path resolver (asynchronous)',
 	options: {
 		resolvePath: function ( importee, importer ) {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Ah, nuisance! Sorry about that. Will of course merge this, but before I do can I ask about the whitespace differences in the test files? I see that it's intentional, I just want to make sure that it's not a change that's going to go back and forth depending on the committer's OS :-)

Bobris:
Ouch these pesky new lines. Second commit is really bad. I still don't understand newlines at end of files as I didn't saw code which would do it. Please commit only first one. I will study more code to write always "\n" and not platform dependent new lines.

Bobris:
Removed second commit so it is now easy to merge.

Rich-Harris:
Huh, thanks for closing this - I thought it had auto-closed via merge but apparently I don't know how to git

Bobris:
It took me a while understand that you actually already merged it and cursed git saying me nothing to commit after resolving merge conflict :-)

PR #8, opened 2015-05-28
Remove unused imports and use one that should be used.

eventualbuddha:


Rich-Harris:
:+1: Thanks, both for the tidy up and the reminder that I intend to use `blank()` (creates object from `null`) rather than doing the `hasOwnProperty` dance everywhere

PR #9, opened 2015-05-29
Fix for malformed output.

mbostock:
Without this change, you can get malformed stuff like this:

```js
var foo = function() {
  return "foo";
}var bar = function() {
  return "bar";
}
```

(Uncaught SyntaxError: Unexpected token var.)

By appending a newline, a semicolon is implicitly inserted between statements when necessary:

```js
var foo = function() {
  return "foo";
}
var bar = function() {
  return "bar";
}
```

The indentation as a result of this commit is off, but I’m not sure how to fix it because the build fails for me. (I tested this by editing dist directly.)

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -44,7 +44,7 @@ export default class Module {
 			});
 
 			this.statements = ast.body.map( node => {
-				const magicString = this.magicString.snip( node.start, node.end );
+				const magicString = this.magicString.snip( node.start, node.end ).append( '\n' );
 				return new Statement( node, magicString, this );
 			});
 		} catch ( err ) {
--- a/src/Module.js
+++ b/src/Module.js
@@ -44,7 +44,7 @@ export default class Module {
 			});
 
 			this.statements = ast.body.map( node => {
-				const magicString = this.magicString.snip( node.start, node.end ).append( '\n' );
+				const magicString = this.magicString.snip( node.start, node.end ).prepend( '\n' ).append( '\n' );
 				return new Statement( node, magicString, this );
 			});
 		} catch ( err ) {
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -248,7 +248,7 @@ export default class Statement {
 	}
 
 	replaceIdentifiers ( names ) {
-		const magicString = this.magicString.clone().trim();
+		const magicString = this.magicString.clone();
 		const replacementStack = [ names ];
 		const nameList = keys( names );
 
--- a/src/finalisers/umd.js
+++ b/src/finalisers/umd.js
@@ -16,7 +16,7 @@ export default function umd ( bundle, magicString, exportMode, options ) {
 
 	if ( exportMode === 'named' ) {
 		amdDeps.unshift( `'exports'` );
-		cjsDeps.unshift( `'exports'` );
+		cjsDeps.unshift( `exports` );
 		globalDeps.unshift( `(global.${options.moduleName} = {})` );
 
 		args.unshift( 'exports' );


mbostock:
The build no longer fails for me (inexplicably), but the tests don’t pass. Hopefully you can figure out the right solution? :grin:

mbostock:
FWIW, I found another bug, but don’t know what the solution is yet. It generated malformed output when I had a default export at the top of a file rather than at the bottom. The file that generated the error looked like this:

```js
// The leaf groups of the selection hierarchy are initially NodeList,
// and then lazily converted to arrays when mutation is required.
export default function(selection) {
  return selection._root = arrayifyNode(selection._root, selection._depth);
};

function arrayifyNode(nodes, depth) {
  var i = -1,
      n = nodes.length,
      node;

  if (--depth) {
    while (++i < n) {
      if (node = nodes[i]) {
        nodes[i] = arrayifyNode(node, depth);
      }
    }
  }

  else if (!Array.isArray(nodes)) {
    var array = new Array(n);
    while (++i < n) array[i] = nodes[i];
    array._parent = nodes._parent;
    nodes = array;
  }

  return nodes;
}
```

And the fixed file that worked around the error:

```js
function arrayifyNode(nodes, depth) {
  var i = -1,
      n = nodes.length,
      node;

  if (--depth) {
    while (++i < n) {
      if (node = nodes[i]) {
        nodes[i] = arrayifyNode(node, depth);
      }
    }
  }

  else if (!Array.isArray(nodes)) {
    var array = new Array(n);
    while (++i < n) array[i] = nodes[i];
    array._parent = nodes._parent;
    nodes = array;
  }

  return nodes;
}

// The leaf groups of the selection hierarchy are initially NodeList,
// and then lazily converted to arrays when mutation is required.
export default function(selection) {
  return selection._root = arrayifyNode(selection._root, selection._depth);
};
```

The malformed output was:

```js
    function arrayifyNode(nodes, depth) {
      var i = -1,
          n = nodes.length,
          node;

      if (--depth) {
        while (++i < n) {
          if (node = nodes[i]) {
            nodes[i] = arrayifyNode(node, depth);
          }
        }
      }

      else if (!Array.isArray(nodes)) {
        var array = new Array(n);
        while (++i < n) array[i] = nodes[i];
        array._parent = nodes._parent;
        nodes = array;
      }

      return nodes;
    }
// The leaf groups of the selection hierarchy are initially NodeList,
    // and then lazily converted to arrays when mutation is required.var arrayify = function(selection) {
      return selection._root = arrayifyNode(selection._root, selection._depth);
    }
```

Rich-Harris:
The non-deterministic test results might possibly be related to https://github.com/Rich-Harris/magic-string/issues/12 (magic-string is basically a hack for manipulating strings in a sourcemap-aware way - saves a bunch on AST manipulation and code generation but there are some rough edges around indentation), will look into it.

Any chance you still have the code that was generating the first malformed output? I can't seem to reproduce it, though I've definitely noticed it before in passing (albeit harmlessly up till now), and want to make sure we can guard against regressions.

Rich-Harris:
Sorry, I misunderstood you when you said the tests were failing - it's because they're expecting particular whitespace. The idea was to try and preserve the rhythm of the original code and keep comments attached to the right lines as far as possible, like this...

```js
// these statements are on consecutive lines
one();
two();
three();

thenWeHaveABreak();

/**
 * JSDoc comment
 */
andAnotherBreak(); // trailing line comments are kept in place too
```

...so padding each statement with newlines creates different output. I'm probably being too anal, though I might try an approach that simply guarantees at least one newline before each statement, and otherwise sticks to the existing method. That should fix the comment bug as well

mbostock:
I’m all for preserving the original formatting of the code as much as possible. I think that’s a lovely feature. I was just bashing it with a hammer to get it to work and look forward to your more-correct fix.

Rich-Harris:
This is fixed and released in 0.7.0

mbostock:
Works on my end! Thank you.

I did notice a few missing newlines still, but everything seemed to run okay. I only noticed it when there were comments (maybe comments at the start of a file?). For example, see the comments here:

```js
    var emptyOf = function(selection) {
      return new Selection(emptyNode(arrayify(selection), selection._depth), selection._depth);
    }// Lazily constructs the exit selection for this (update) selection.
    // Until this selection is joined to data, the exit selection will be empty.

    var selection_exit = function() {
      return this._exit || (this._exit = emptyOf(this));
    }// Lazily constructs the enter selection for this (update) selection.
    // Until this selection is joined to data, the enter selection will be empty.

    var selection_enter = function() {
      if (!this._enter) {
        this._enter = emptyOf(this);
        this._enter._update = this;
      }
      return this._enter;
    }
```

If you clone the [d3-selection repo](https://github.com/d3/d3-selection) you should be able to see this in build/d3.js after running `npm run test`.

Rich-Harris:
@mbostock comment handling is improved in 0.7.3 - I've tried it out with d3-selection and it behaves a bit better

Issue #30, opened 2015-06-18
Custom naming systems


guybedford:
Really loving this project! The APIs are great to work with.

Just been playing around with hooking it in to SystemJS builder and I'm even defining a custom loader to manage everything in-memory which is great.

It would be nice to be able to use the same naming system as SystemJS without having to translate into file paths, so I'm trying to set both resolvePath and resolveExternal, but it seems the entry point still gets resolved to a path.

Would it be possible to have the above resolution run through the same hooks as above?

The ES specification defines module loading through two hooks:

* HostNormalizeModuleName(ID, parentID)
* HostGetSource(normalizedID)

It would be nice to maintain some consistency with this by allowing any naming system.

I know it's very much my use case, but let me know if it's a possibility.

Rich-Harris:
Great! This seems like it might be relevant to #25 (getting Rollup to work in a browser environment). If I understand you correctly, the goal is to replace [this line](https://github.com/rollup/rollup/blob/master/src/Bundle.js#L18) with something that loosely equates to HostNormalizeModuleName (i.e. `resolvePath`, but where [`importer`](https://github.com/rollup/rollup/blob/master/src/utils/resolvePath.js#L6) is `undefined`), so that a subsequent HostGetSource (a.k.a. `load`) operation can load the entry module in the absence of a filesystem. Is that right?

If so, that seems both doable and desirable.

guybedford:
Yes, exactly!

Rich-Harris:
Have to run now, but I've made a start on this in #31 - will pick it up soon, but in the meantime checking out that branch and `npm link`ing it will allow you to try it out if needs be.

Rich-Harris:
Fixed as of 0.9.0. #31 for details. @guybedford let me know if this behaves as you expect!

Issue #25, opened 2015-06-16
Browser support


Rich-Harris:
It would be useful to be able to run rollup in-browser, for the sake of custom builders (cc @mbostock) and interactive demos.

There are four external dependencies - `path`, `sander`, `magic-string` and `acorn`. `acorn` and `magic-string` both work in-browser (and `magic-string` and its only dependency use `jsnext:main`, so they be bundled with Rollup). The `path` functions can be replaced with internal helpers without too much difficulty.

`sander` is a little trickier, since it's a wrapper around `fs`. But if there were a neat way to exclude it from the build there'd be no problem - it's possible to supply a custom `load` function which returns the contents of the sought module (it just so happens that the default `load` uses `sander` to read from disk).

Rich-Harris:
This is implemented in 0.11.0. Will try and get a demo site up soon

donmccurdy:
Is there an example or demo? I'm wondering about doing this in the browser:

```js
rollup.rollup({
  entry: '/foo/bar.js',
  plugins: [ ??? ]
}).then( bundle => {
  let { code, map } = bundle.generate({
    format: 'cjs'   // amd, es6, iife, umd
  });
  console.log(code);
});
```

.. where `/foo/bar.js` is path relative to the current page, which might depend on other files modules.

PR #31, opened 2015-06-19
Apply resolve hook to entry module

Rich-Harris:
**not ready for merge**

This enables #30, and by extension #25 (eventually), by applying the `resolvePath` hook to `options.entry`. `resolvePath` is now a misnomer; it should become `resolveId` or something similar, and there are a few other bits of internal terminology that will probably need updating as well.

Something to note: `bundle.entryPath` and `bundle.base` no longer exist. These were used for the purposes of generating non-conflicting identifiers by restricting the 'available' path parts to those within the source directory (i.e. `_foo` to deconflict with `foo`, rather than `Users_rharris_code_someProject_src_foo`) - for now, that's being worked around, but it'll need a more robust approach eventually.

* [ ] fix internal terminology
* [ ] alternative approach to `bundle.base`
* [ ] update docs

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -1,4 +1,4 @@
-import { basename, dirname, extname, relative, resolve } from 'path';
+import { basename, dirname, extname, relative } from 'path';
 import { Promise } from 'sander';
 import MagicString from 'magic-string';
 import { blank, keys } from './utils/object';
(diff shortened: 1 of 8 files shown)


guybedford:
I guess `bundle.base` is something used by the default id resolver? Surely it can still exist in that way though?

Perhaps the `resolveId` name change can be a later breaking change, and a first step can get the functionality in minimally breaking?

Rich-Harris:
Wow, was it really three weeks ago that I opened this PR? :dizzy_face: Anyway, this is released as 0.9.0 - `resolveId` has taken the place of `resolvePath` (I'd already made this change locally - figured it's better to try and do it right in one go), and is applied to all modules including the entry module, so this should all work now. I'm glossing over the `bundle.base` stuff for the time being.

guybedford:
Awesome! Will test it out very soon. It should even allow self-hosting SystemJS's build in future. :dog: :pizza:

PR #35, opened 2015-07-08
Allow reassignments

Rich-Harris:
Quick hack in service of an experiment, which may turn out to be more generally useful so I'm leaving it here for future me to consider merging.

The spec disallows this sort of thing...

```js
import { uid } from './foo';

export class Bar {
  constructor () {
    this.id = uid++; // can't update or assign to an imported binding
  }
}
```

...but the equivalent ES5 is used in the wild in libraries like three.js (`THREE.GeometryIdCount++` and so on). Allowing reassignments may be the easiest way to help transition existing codebases to ES6.

### DIFF
--- a/bin/help.md
+++ b/bin/help.md
@@ -5,15 +5,16 @@ Usage: rollup [options] <entry file>
 
 Basic options:
 
--v, --version            Show version number
--h, --help               Show this help message
--i, --input              Input (alternative to <entry file>)
--o, --output <output>    Output (if absent, prints to stdout)
--f, --format [umd]       Type of output (amd, cjs, es6, iife, umd)
--e, --external           Comma-separate list of module IDs to exclude
--n, --name               Name for UMD export
--u, --id                 ID for AMD module (default is anonymous)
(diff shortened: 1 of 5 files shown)


caridy:
that will be problematic. imagine the scenario where browsers implements modules (hopefully soon) and you have `TREE` authored in ES6, and GeometryIdCount can be mutable because you're using a transpiler that diverge from the specifications, which means the original source will never work in a browser.

Rich-Harris:
Yeah, on reflection I'm not wild about it either. I was trying to build [three-jsnext](https://github.com/rollup/three-jsnext), and this seemed like the easiest interim solution. In the end I went with an alternative - turning this...


```js
// exporter
THREE.GeometryIdCount = 0;

// importer
this.id = THREE.GeometryIdCount ++;
```

...into this...

```js
// exporter
var count = 0;
export function THREE$GeometryIdCount () { return count++ }

// importer
this.id = THREE$GeometryIdCount();
```

...which renders `allowReassignments` unnecessary. (Unfortunately there are some other issues related to cyclical dependencies, so nothing works anyway yet, but that's a separate issue.) So I'll close this.

Issue #39, opened 2015-07-13
Missing side-effects


Rich-Harris:
Given this situation...

```js
// main.js
import Foo from './Foo';
var foo = new Foo();
var baz = new Foo.Baz();

// Foo.js
import { Bar } from './Bar';
import { Baz } from './Baz';

Bar.Baz = Baz; // this statement is ignored

export default Bar;

// Bar.js
export function Bar () {
  alert( 'bar' );
}

// Baz.js
export function Baz () {
  alert('baz');
}
```

...we'd expect `Foo` to have a `Baz` property, but Rollup goes straight to the original `Foo` definition, ignoring the `Bar.Baz = Baz` line:

```js
function Bar () {
  alert( 'bar' );
}

var Foo = Bar;

var foo = new Foo();
var baz = new Foo.Baz();
```

We need to catch any statements that mutate imported bindings, wherever they are.

eventualbuddha:
Is this fixed now? If so, mind releasing a build with the fix?

Rich-Harris:
Yep, fixed - sorry, got sidetracked, will push out a release momentarily

Rich-Harris:
Released as 0.11.0

eventualbuddha:
Thanks!

PR #40, opened 2015-07-13
Add a failing test that illustrates a problem with RSVP.

eventualbuddha:
I ran into this trying to use rollup with [LGTM](https://github.com/square/lgtm) where I bundle only the parts of RSVP that I need. The problem is that rollup seems not to consider the side effect of setting `config.async = asap;` worth preserving because the file that contains is pretty much just passes an import directly into an export.

Perhaps rollup should consider the statement referenced above worth preserving if it uses something that is also used by any of its exports, namely `defer`.

For reference, this is the output that rollup generates:

```js
'use strict';

const config = {};

function defer() {
  config.async();
}

defer();
```

### DIFF
--- a/test/function/top-level-side-effects-are-preserved/_config.js
+++ b/test/function/top-level-side-effects-are-preserved/_config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  description: 'top level side effects are preserved'
+};
--- a/test/function/top-level-side-effects-are-preserved/asap.js
+++ b/test/function/top-level-side-effects-are-preserved/asap.js
@@ -1,0 +1,1 @@
+export default function asap() {}
--- a/test/function/top-level-side-effects-are-preserved/config.js
+++ b/test/function/top-level-side-effects-are-preserved/config.js
@@ -1,0 +1,1 @@
+export const config = {};
--- a/test/function/top-level-side-effects-are-preserved/defer.js
+++ b/test/function/top-level-side-effects-are-preserved/defer.js
@@ -1,0 +1,5 @@
+import { config } from './config';
+
+export default function defer() {
+  config.async();
+}
--- a/test/function/top-level-side-effects-are-preserved/main.js
+++ b/test/function/top-level-side-effects-are-preserved/main.js
@@ -1,0 +1,3 @@
+import { defer } from './rsvp';
+
+defer();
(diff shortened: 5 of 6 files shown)


Rich-Harris:
Thanks - I *think* this might be the same as #39, so hopefully this test case can cover both issues. I think I have an idea about how to fix it, as part of some wider refactoring/tweaking made possible by 0.10.

eventualbuddha:
@Rich-Harris It seems likely. What's your thought on fixing it?

Rich-Harris:
@eventualbuddha In 0.9 and below, Rollup works by 'expanding' statements, which looks something like this:

```js
expand = statement => ([
  ...dependenciesOf(statement).map(expand), // e.g. `var foo = {bar}` depends on `bar`
  statement,
  ...statementsThatModify(statement).map(expand) // e.g. `foo.bar = 1` modifies `foo`
]);
```

Obviously it's a good deal more complicated in practice because of asynchrony and cyclicality, etc, but that's the gist. The theory was that expanding all the statements in the entry module would be sufficient to generate a correctly-ordered list of statements from the entire bundle. That theory is now disproven - it turns out that the sorting needs to happen at the module level (using the same sorting algorithm as Esperanto - start off with a topological sort, then shake the result to ensure cyclical dependencies are in the right order), after which we just need to include whichever statements in each module were 'touched' by the expansion.

So that's what happens in 0.10, but other than the correct sorting algorithm the implementation hasn't really changed - we're still building up `bundle.statements` via recursive expansion. Aside from the volume of garbage arrays this process generates, there's a flaw - [a statement is responsible for finding other statements that modify its own value, and it only looks in its own module](https://github.com/rollup/rollup/blob/master/src/Statement.js#L263), when we actually need to inspect the entire bundle.

My thinking is that the whole thing could be simplified - we find all the dependencies of the entry module's statements, recursively, then inspect the entire bundle to find statements that modify any of the names that have been defined along the way. If that yields any *new* definitions (e.g. `foo.bar = baz` means we need to go away and define `baz`), we repeat the process until it settles. Then we simply concatenate all the statements that have been touched by the process. Result is exactly the same as at present except that side-effects like the one above are included. At least, that's the idea.

Probably a bit clearer in my head than I was articulate but hopefully that makes some sense - going to try and do this today.

PR #43, opened 2015-07-16
Test that top-level side effects involving imports are preserved.

eventualbuddha:
With the current implementation of rollup, this test fails. To make it pass, you can replace `import obj from './obj;` in lib.js with `var obj = {};`, suggesting that it's the import itself that somehow confuses rollup.

### DIFF
--- a/test/function/top-level-side-effect-on-imported/_config.js
+++ b/test/function/top-level-side-effect-on-imported/_config.js
@@ -1,0 +1,3 @@
+module.exports = {
+	description: 'top level function calls are preserved'
+};
--- a/test/function/top-level-side-effect-on-imported/lib.js
+++ b/test/function/top-level-side-effect-on-imported/lib.js
@@ -1,0 +1,5 @@
+import obj from './obj';
+
+obj.value = 'augmented';
+
+export default obj;
--- a/test/function/top-level-side-effect-on-imported/main.js
+++ b/test/function/top-level-side-effect-on-imported/main.js
@@ -1,0 +1,3 @@
+import lib from './lib';
+
+assert.strictEqual(lib.value, 'augmented');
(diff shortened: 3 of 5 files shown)


eventualbuddha:
This is the output from rollup:

```js
var obj = { value: 'original' };

var lib = obj;

assert.strictEqual(lib.value, 'augmented');
```

And this is what esperanto outputs (removing the AMD wrapper and "use strict"):

```js
var obj = { value: 'original' };

obj.value = 'augmented';

var lib = obj;

assert.strictEqual(lib.value, 'augmented');
```

Rich-Harris:
Looks like it's the default export that's tripping it up - this works fine:

```js
// main.js
import lib from './lib';
assert.strictEqual( lib.value, 'augmented' );

// lib.js
import { obj } from './obj';

obj.value = 'augmented';

export default obj;

// obj.js
const obj = { value: 'original' };
export { obj };
```

Don't know the solution off the top of my head but at least we know where to look...

eventualbuddha:
It looks like it might be because there is no `definingStatement` associated with `obj` in the `Statement` representing `obj.value = 'augmented';`, that is both `definingStatement` and `exportDeclaration` [here](https://github.com/rollup/rollup/blob/ac1a0ef85223bfb709c81ce3e731784e84993ecc/src/Bundle.js#L419-L423) are `undefined`, so `shouldMark` is falsy:

```js
const definingStatement = module.definitions[ name ];
const exportDeclaration = module.exports[ name ];

const shouldMark = ( definingStatement && definingStatement.isIncluded ) ||
                   ( exportDeclaration && exportDeclaration.isUsed );
```

You probably already knew this, so this is mostly just showing my work for debugging. Not familiar enough with the rest of it to know what approach should be used to fix it.

Rich-Harris:
Think I've got this bug squashed - have opened #48

Issue #44, opened 2015-07-16
rollup fails on "export * from ..."


Victorystick:
When I attempt to `rollup` code like that below, I get this error:
```
Module wat.js does not export wat (imported by main.js)
```

I assume this is due to "export * from ...". Minimal example:
```javascript
// --- main.js
import { wat } from './wat';

wat();

// --- wat.js
export * from './wat-impl';

// --- wat-impl.js
export function wat() {}
```

Do you have plans to support this pattern in the pipeline?

Rich-Harris:
Ah. Yep, this definitely needs to be supported, thanks for raising it

Victorystick:
I guess I'll close this since the PR got merged. :+1:

PR #46, opened 2015-07-16
Implemented export * from ...

Victorystick:
Discovering esperanto and then rollup has been very exciting. I always felt that it should be possible bundle ES modules more efficiently than compiling them to CommonJS and bundling with Browserify. I'm really drawn towards this project.

I gave implementing `export * from '...'` a go, as per #44. I added the example I presented there as a test, and got it passing. I'm not sure I entirely understand the bundling process, but I'm sure you can oblige me with some pointers if something's amiss.

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -6,7 +6,7 @@ import Statement from './Statement';
 import walk from './ast/walk';
 import analyse from './ast/analyse';
 import { blank, keys } from './utils/object';
-import { sequence } from './utils/promise';
+import { first, sequence } from './utils/promise';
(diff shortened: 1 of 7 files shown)


Rich-Harris:
This is awesome! Thanks, this PR looks spot on, I would have spent a long time figuring out how to implement this. Will push a release out shortly

Rich-Harris:
Released as 0.11.1. I invited you to be a github collaborator in case that's of interest, totally up to you :-)

Victorystick:
Awesome! I look forward to seeing rollup usage go, nay roll, up!

PR #47, opened 2015-07-17
Dynamically defined object properties

Victorystick:
When I bundled some code with a few functions of a modern, ES build of Lodash (similar to the README example), I encountered a ReferenceError at runtime.

It is caused by Lodash defining a number of variables with string tags, which are used to populate an object. Extract from Lodash [here](https://github.com/lodash/lodash/blob/master/lodash.js#L49) and [here](https://github.com/lodash/lodash/blob/master/lodash.js#L168):

```js
/** `Object#toString` result references. */
  var argsTag = '[object Arguments]',
      arrayTag = '[object Array]',
      boolTag = '[object Boolean]',
      ...

// Later...

/** Used to identify `toStringTag` values supported by `_.clone`. */
  var cloneableTags = {};
  cloneableTags[argsTag] = cloneableTags[arrayTag] =
  cloneableTags[arrayBufferTag] = cloneableTags[boolTag] =
  cloneableTags[dateTag] = cloneableTags[float32Tag] =
  cloneableTags[float64Tag] = cloneableTags[int8Tag] =
  cloneableTags[int16Tag] = cloneableTags[int32Tag] =
  cloneableTags[numberTag] = cloneableTags[objectTag] =
  cloneableTags[regexpTag] = cloneableTags[stringTag] =
  cloneableTags[uint8Tag] = cloneableTags[uint8ClampedTag] =
  cloneableTags[uint16Tag] = cloneableTags[uint32Tag] = true;
  cloneableTags[errorTag] = cloneableTags[funcTag] =
  cloneableTags[mapTag] = cloneableTags[setTag] =
  cloneableTags[weakMapTag] = false;

```
The assignment statements to `clonableTags` do not recognize the tag variables as dependencies, so the definitions of the tags are not included resulting in a ReferenceError. This PR solves this issue, by checking for reads in computed member expressions.

Since I could not find any references to `BinaryExpression` in the sources, I was unsure whether expressions like
```js
var n = 'Number';

clonableTags['[object ' + n + ']'] = true;
```
would work for the same reason, but they appear to work just fine even without the fix. Could you point out how it is handled?

Bundle for tests before PR:
```js
// Only the code below is included in the bundle.

var tags = {};

tags[objTag] = tags[arrTag] =
tags[fnTag] = tags[dateTag] = true;

tags[strTag] = tags[numTag] = false;

assert.equal(tags['[object Object]'], true);
assert.equal(tags['[object Number]'], false);
assert.equal(Object.keys(tags).length, 6);
```
After:
```js
// Emulates a piece of Lodash source code.
// The tag strings are not imported, and a runtime error ensues.

var objTag = '[object Object]';
var arrTag = '[object Array]';
var fnTag = '[object Function]';
var dateTag = '[object Date]';

var strTag = '[object String]';
var numTag = '[object Number]';


// Only the code below is included in the bundle.

var tags = {};

tags[objTag] = tags[arrTag] =
tags[fnTag] = tags[dateTag] = true;

tags[strTag] = tags[numTag] = false;

assert.equal(tags['[object Object]'], true);
assert.equal(tags['[object Number]'], false);
assert.equal(Object.keys(tags).length, 6);
```

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -175,10 +175,24 @@ export default class Statement {
 	}
 
 	checkForWrites ( scope, node ) {
+		const parent = node;
+
 		const addNode = ( node, isAssignment ) => {
 			let depth = 0; // determine whether we're illegally modifying a binding or namespace
 
 			while ( node.type === 'MemberExpression' ) {
+
+				// In a situation like that below, make sure the assignments
+				// depend on `a` and `b`.
+				//
+				//    var a = 1, b = 0;
+				//
+				//    arr[a] = arr[b] = true;
+				//
+				if ( node.computed ) {
+					this.checkForReads( scope, node.property, parent, true );
+				}
+
 				node = node.object;
 				depth += 1;
 			}
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -250,7 +250,7 @@ export default class Statement {
 	}
 
 	mark () {
-		if ( this.included ) return; // prevent infinite loops
+		if ( this.isIncluded ) return; // prevent infinite loops
 		this.isIncluded = true;
 
 		const dependencies = Object.keys( this.dependsOn );
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -134,7 +134,7 @@ export default class Statement {
 					}
 
 					this.checkForReads( scope, node, parent, !depth );
-					this.checkForWrites( scope, node, !depth );
+					this.checkForWrites( scope, node );
 				},
 				leave: ( node, parent ) => {
 					if ( node._scope ) {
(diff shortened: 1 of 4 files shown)


eventualbuddha:
This looks good to me. I'll give @Rich-Harris a chance to review before merging since my familiarity with the project is still not strong.

Rich-Harris:
Just getting a chance to look at this now - thanks @Victorystick. I *think* we can actually do it more simply - the failure seems to be that it's treating all `MemberExpression` properties as non-computed. I've  pushed up a new branch based off this one that adds a `!parent.computed` check [here](https://github.com/rollup/rollup/compare/master...rollup:object-dynamic-properties-alt#diff-2964822a0500976c7a3115375c1d3245R156,) and the test passes, does that seem correct? If so will merge and release

Victorystick:
Yes. My changes has quite bothered me today. I just wasn't satisfied. It took me a while to find the missing `!parent.computed` check. You had remembered it in the renaming section.

Your commit is _just_ what I was looking for. I say merge it! :+1:

Rich-Harris:
Released as 0.11.2

PR #48, opened 2015-07-19
Fix #43

Rich-Harris:
Fix for #43. A couple of things were going wrong:

* the code that scans statements to see if they're used elsewhere in the bundle wasn't handling default exports
* no export declarations were actually being marked as `isUsed`. Not quite sure how that happened, must have borked a commit

### DIFF
--- a/test/function/top-level-side-effects-are-preserved/_config.js
+++ b/test/function/top-level-side-effects-are-preserved/_config.js
@@ -1,3 +1,3 @@
 module.exports = {
-	description: 'top level side effects are preserved'
+	description: 'top level side effects on imports are preserved'
 };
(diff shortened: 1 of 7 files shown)


Victorystick:
Looks good to me.

test/function/top-level-side-effects-are-preserved/_config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-	description: 'top level side effects are preserved'
+	description: 'top level side effects on imports are preserved'
```
eventualbuddha: I think this change was an accidental one by me and should be left out. I think I force-pushed my branch to remove it.

eventualbuddha:
Seems good to me!

Issue #37, opened 2015-07-08
Functions are included unnecessarily


Rich-Harris:
Since functions are assumed to be capable of mutating their arguments, `mutate` is correctly included when you import `foo`:

```js
let foo = {};
mutate( foo );
export default foo;
```

But the same logic includes `irrelevant`:

```js
let foo = {};

function irrelevant () {
  mutate( foo );
}

export default foo;
```

Only top-level call expressions need be included. (As mentioned previously, it would eventually be nice to  analyse function definitions and determine whether they *are* capable of mutating their arguments.)

Rich-Harris:
This is long since [fixed](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20foo%20from%20'.%2Ffoo.js'%5Cnconsole.log(foo)%22%7D%2C%7B%22name%22%3A%22foo.js%22%2C%22code%22%3A%22let%20foo%20%3D%20%7B%7D%3B%5Cn%5Cnfunction%20irrelevant%20()%20%7B%5Cn%20%20mutate(%20foo%20)%3B%5Cn%7D%5Cn%5Cnexport%20default%20foo%3B%22%7D%5D%7D) – closing

PR #49, opened 2015-07-19
exclude unnecessary functions containing assignments/updates to included names (#37)

Rich-Harris:
This addresses #37, by tracking whether we're inside a function declaration (or a `var foo = function () {...}`), which is effectively the same thing) before marking a statement as being a modifier of a particular name. This means that `unused` is excluded in the example below...

```js
export let foo = {};

function unused () {
  foo.bar = 'baz';
}
```

...but something like this would still be included:

```js
export let foo = {};

[ 'a', 'b', 'c' ].forEach( function ( letter, i ) {
  foo[letter] = i;
});
```


Rich-Harris:
Hmm. Haven't quite got this right yet - not ready for merge. Still a few false positives.

Victorystick:
Isn't the issue centered around the difference between statements that are executed, and function definitions (declarations and function expressions)? Don't you think it might be better to keep track of functions separately from other statements?

A function declaration (kind of statement) defines a function and binds it to a variable, effectively modifying it.

Rich-Harris:
We could treat function declarations (or quasi-declarations, i.e. `var foo = function () {...}`) differently, yeah – the `writeDepth` thing is *supposed* to be doing exactly that, but it might be too much. Though we also have to accommodate this sort of thing:

```js
let fn;
if ( someCondition ) {
  fn = () => { /* do one thing */ };
} else {
  fn = () => { /* do another thing */ };
}
```

I've found it gets even trickier with methods like this...

```js
var obj = {
  mutate: () => foo.mutated = true
};
```

...which shouldn't be included just because `foo` is. I haven't quite figured out how to differentiate between that and this:

```js
({
  mutate: () => foo.mutated = true
}).mutate();
```

I'm leaving open the possibility that we're already at the limits of what can be done with static analysis in JavaScript.

PR #54, opened 2015-07-22
Fix #53

### LINKED ISSUE #53: Rollup hangs on unknown external modules
When rollup tries to bundle an external module which doesn't exist. It gets stuck in an infinite loop in [defaultExternalResolver](https://github.com/rollup/rollup/blob/master/src/utils/resolveId.js#L55), where `dir` was modified like so

```js
'/home/user/repo' -> 
'/home/user' -> '/home' -> '' -> '.' -> '.' -> ...
```
never reaching the terminating condition `dir === '/'`.

The issue was mentioned in #52. Thanks for bringing it up @petulla!

Victorystick:
See issue #53 for explaination. This PR fixes `dirname` to handle the root (empty string) case.

### DIFF
--- a/src/utils/path.js
+++ b/src/utils/path.js
@@ -11,15 +11,19 @@ export function basename ( path ) {
 }
 
 export function dirname ( path ) {
-	const match = /(\/|\\)[^\/\\]+$/.exec( path );
+	const match = /(\/|\\)[^\/\\]*$/.exec( path );
 	if ( !match ) return '.';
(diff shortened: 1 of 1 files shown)


eventualbuddha:
@Victorystick Nice! Tests?

Rich-Harris:
Have added a test and merged this - going to push out a new release momentarily

Rich-Harris:
'momentarily', he said... released as 0.12.0

PR #56, opened 2015-07-22
allow importing external modules for side effects, fixes #55

### LINKED ISSUE #55: Importing external module for side effects fails
works:
```sh
echo "import foo from \"foo\"" > rollup.js && rollup -e foo rollup.js
```

fails:
```sh
echo "import \"foo\"" > rollup.js && rollup -e foo rollup.js
```

Swatinem:


Victorystick:
This looks good! :+1:

PR #58, opened 2015-07-22
Fix #57

### LINKED ISSUE #57: Fragile dependency resolution pattern in UMD format
Bundling the following to UMD, where 'factory' is external
```js
import factory from 'factory';

factory(null);

// yields:

(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(require('factory')) :
	typeof define === 'function' && define.amd ? define(['factory'], factory) :
	factory(factory);
}(this, function (factory) { 'use strict';

	factory(null);

}));
```
As you can see, the UMD factory function is passed itself instead of it's dependency. Mitigated by prefixing dependency name with `global.`.

```diff
	typeof define === 'function' && define.amd ? define(['factory'], factory) :
-	factory(factory);
+	factory(global.factory);
}(this, function (factory) { 'use strict';
```

Victorystick:
Prefix globalDeps with `global.`, to fix #57.

### DIFF
--- a/src/finalisers/umd.js
+++ b/src/finalisers/umd.js
@@ -11,7 +11,7 @@ export default function umd ( bundle, magicString, { exportMode, indentString },
 	let amdDeps = bundle.externalModules.map( quoteId );
 	let cjsDeps = bundle.externalModules.map( req );
 	let globalDeps = bundle.externalModules.map( module => {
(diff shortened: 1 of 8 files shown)


Rich-Harris:
:+1:

PR #60, opened 2015-07-24
Add globals option to rollup executable.

Victorystick:
Also:
* Sorted minimalist aliases alphabetically
* Updated help.md to include correct format default (ES6).
* Removed unused requires and variables from `bin/runRollup.js`.
* Added missing definition of `map` for outputting inline source maps.
* Grabs external names from globals option. These are equivalent:

```
rollup -f iife --globals lodash:_,jquery:jQuery
rollup -f iife --globals lodash:_,jquery:jQuery --external lodash,jquery
```


Rich-Harris:
:+1: :sunglasses:

PR #64, opened 2015-07-25
Move rendering logic into modules

Rich-Harris:
Not ready for merge - there's a couple of failing tests, and some tidying up still to do.

This PR takes advantage of the fact that sorting now happens at the module level rather than the statement level. It means that we don't need to have a `MagicString` instance per statement (with all the additional cloning and memory consumption that entails), but rather one per module, and each module can take care of itself. It also means that handling things like comments gets a good bit easier.

In time, it may allow some other improvements, e.g. to the way canonical names are determined.

All the tests are passing except some new ones I just added, around multiple var declarations, because these now need to be handled a bit differently.


Rich-Harris:
All tests now passing.

Victorystick:
I'll take a proper look at this tomorrow. Seems good, looking at the tests and overall changes. Don't forget to update the README etc. when it's merged, since the API has changed.

Rich-Harris:
The public API is the same - it's just the internal method name that's changed from `generate` to `render`

Victorystick:
Right. That's me not yet taking a proper look.

Victorystick:
:+1: Merging.

Issue #61, opened 2015-07-24
Shorthand properties


Rich-Harris:
Unable to check right now, but I'd strongly expect [this Esperanto issue](https://github.com/esperantojs/esperanto/issues/183) to apply to Rollup as well:

```js
import foo from 'foo';

export default {
  // `foo` might be rewritten `_foo`, which means the
  // shorthand property no longer works
  foo 
}
```

Victorystick:
This ought to apply to all object shorthands, right? Not just those with exports. We'll just have to expand the shorthands, if we rename the variables.

PR #65, opened 2015-07-27
Revert "expand shorthand properties (#61)"

Victorystick:
Reverts rollup/rollup#63.

After merging #64 I re-ran the tests on Travis. I thought they'd take the changes for #64 into account, which they didn't. The result was test that failed when testing locally. Had #63 been rebased on current master, the failing test `unused-var-d` would have been detected.

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -327,28 +327,18 @@ export default class Statement {
 					replacementStack.push( newNames );
 				}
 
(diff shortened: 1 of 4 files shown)


Victorystick:
Removed this completely. I'm a fool. :anguished: I forgot to update my node_modules. Travis did the right thing apparently. All is good with the world.

Rich-Harris:
Heh, been there done that

PR #66, opened 2015-07-27
Implement features needed by Esperanto

Rich-Harris:
Esperanto is continuing to rack up issues, but it doesn't make sense to maintain two projects doing basically the same thing. I'm planning to remove all the bundling logic from Esperanto and just use Rollup instead. Before that's possible, we need to reach feature parity (and fix some bugs exposed by Esperanto):

* [x] replace `exports.default` with `exports['default']`
* [x] include `imports` and `exports` in exposed bundle object
* [x] prevent module from importing itself
* [x] fix typo with AMD module IDs (`['myModule']` -> `'myModule'`)
* [x] make `'use strict'` optional (`options.useStrict = false`)
* [x] support `banner` and `footer` options
* [ ] fix various bugs, yet to catalogue


Victorystick:
By "remove all bundling logic from Esperanto", what do you intend the main use case for Esperanto to be?

Rich-Harris:
Ultimately, it doesn't really need to exist at all. The bundling has been superseded by Rollup, and the one-to-one conversion can be done by Babel (Esperanto is faster, because it does less, but the reality is most people using ES6 modules are probably using other ES6 features anyway, so would need Babel regardless). But enough people depend on Esperanto that I'd prefer not to completely shut off support – this is just a way of achieving that, by using Rollup internally rather than Esperanto's existing bundler (the API for Esperanto users would be pretty much unaffected).

Rich-Harris:
@Victorystick how far along do you reckon #71 is? We have merge conflicts unfortunately... trying to work out whether it's better to merge this, or wait for that. I've opened a new [bundle-scopes](https://github.com/rollup/rollup/tree/bundle-scopes) branch based off of yours which resolves those conflicts, in case that's useful

Victorystick:
I'm officially on vacation, and probably won't be able to get as much work done this week as I would like. I think it might be best to merge what you've got, and let me continue as soon as I can.

Your work on fixing the current set of merge conflicts is much appreciated. :smiley: Do you see how I intend for BundleScope to work?

Issue #68, opened 2015-07-29
Trying to looking up named import from external module


Swatinem:
We have a quite big (400K source) project that I would like to bundle with rollup, but it it currently fails when I use a named import from an external dependency.

The problem is here:
https://github.com/rollup/rollup/blob/master/src/Bundle.js#L238
`TypeError: Cannot read property 'mat4' of undefined`

I will try to reduce this to a smaller testcase and fix it in the process, but it might take me some time…

Rich-Harris:
Closed via #70

PR #70, opened 2015-07-30
Add failing testcase for #68

Swatinem:
I managed to reduce the failure from #68, with some comments about some observations I made trying to reduce the testcase.

In summary:
you have a relative default import that exports a typical constructor/prototype pattern, where one of the prototype methods is calling a property on a named external import.

* if we do not call the property of the named import but the named import itself, it compiles but generates wrong code
* if we do not define the prototype method but call the property of the named import in the constructor function it compiles but generates wrong code
* if we do not use a relative local import but copy-paste the code right into the `main.js` it compiles but generates wrong code

The compile error is like this:
```
TypeError: Cannot read property 'bar' of undefined
 at .gobble-build/01-rollup/1/rollup.js:1975:48
```

### DIFF
--- a/test/function/named-external-method-in-prototype/_config.js
+++ b/test/function/named-external-method-in-prototype/_config.js
@@ -1,0 +1,15 @@
+module.exports = {
+	description: 'method of external named import used inside prototype method',
+	context: {
+		// override require here, making "foo" appear as a global module
+		require: function ( name ) {
+			if ( name === 'bar' ) {
+				return require( './bar' );
+			}
+			return require( name );
+		}
+	},
+	options: {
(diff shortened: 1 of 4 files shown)


Rich-Harris:
Thanks @Swatinem! There were a couple of schoolboy errors in the code – the fixes were uncontroversial so I've gone ahead pushed a new version straight out (0.12.1)

Swatinem:
Thanks. One step further but not quite there yet.

Now I get the following:
```
TypeError: Cannot read property 'localName' of undefined
    at .gobble-build/01-rollup/1/rollup.js:1979:62
```

I will try to create a minified testcase tomorrow…

PR #71, opened 2015-08-01
Bundle scopes

Victorystick:
Work in progress...

Use a BundleScope which keeps track of all top-level variables in use by a bundle, and deconflicts them all, once and for all. Each module has a ModuleScope that links its variables to those in the BundleScope.


Victorystick:
This has gotten more complex than it should, and diverged too far from master. I'm gonna take the lessons I've learned and try do to something similar.

PR #73, opened 2015-08-03
add a failing test for re-exporting and renaming a 'default' export

Swatinem:


Swatinem:
I have amended the testcase, now using multiple renamed default exports.

Strangely enough, in my own codebase, this fails to compile with
```
TypeError: Cannot read property 'exports' of undefined
    at node_modules/rollup/dist/rollup.js:2072:38
```

But as a reduced testcase, this does compile, but generates wrong code.

Rich-Harris:
Good news, this is fixed as of 0.14.1 (possibly earlier, but I suspect it was that release)! Thanks for the test case, apologies for not confirming sooner

Issue #16, opened 2015-06-02
Export from should not create a local binding


Rich-Harris:
Unused test [here](https://github.com/rollup/rollup/tree/master/test/function/export-from-no-local-binding)

Rich-Harris:
Fixed in 0.14.1

PR #86, opened 2015-08-17
Reexports

Rich-Harris:
Not ready for merge. This adds two failing tests to cover corner cases around re-exported bindings (see #16 and comments on https://github.com/rollup/rollup/pull/78/files#r37185935)

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -336,7 +336,7 @@ export default class Bundle {
 		});
 
 		// prepend bundle with internal namespaces
-		const indentString = magicString.getIndentString();
+		const indentString = getIndentString( magicString, options );
 		const namespaceBlock = this.internalNamespaceModules.map( module => {
 			const exports = keys( module.exports )
 				.concat( keys( module.reexports ) )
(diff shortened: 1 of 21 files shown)


Rich-Harris:
Not feeling great about the direction this is heading in to be honest... >100 additional LOC (some of it not very DRY, admittedly) and it touches the codebase in more places than I'd anticipated. Am in two minds about whether to start again from e4fd1ef172afac994c0a7b20344d9f5a1989bfb1 or push forward and try to tidy it up, or abandon this line of thinking altogether!

Rich-Harris:
This is in better shape now, and it also covers another tricky case, where a single binding is exported in multiple ways:

```js
// main.js
import { foo } from './foo';
export { foo as a };
export { foo as b };
export { foo as c };

// foo.js
export var foo = 1;
foo = 2;
```

The only realistic way to keep `exports.a`, `exports.b` and `exports.c` in sync is to use getters, so that's what happens (previously, `exports.a` and `exports.b` just wouldn't have been assigned to). Could be optimised in future to cover cases where no reassignment happens, or where ES3 support is more important than spec compliance.

Is a fairly modest increase (~30-40 LOC, net) over 0.14.0. @Victorystick what are your thoughts? With re-exports handled correctly, does this cover all of what #71 was working towards, or would it be better to keep going until all the tests pass in that branch?

Victorystick:
I suppose it's pretty much covers it. One big goal of #71 was to simplify the conflicting logic, but that can be done later. :smile: I think I'd try to make smaller, incremental changes based on this branch instead, and see where that goes.

One thing it would have allowed beyond this, was static resolution of namespaces within the bundle. I don't think that's been done yet, right? So that `ns.other.foo` below would bind directly to `foo`.

```js
// entry.js
import * as ns from './mid';
ns.other.foo();

// mid.js
export * as other from './top';

// top.js
export function foo () {}

Victorystick:
Perhaps the `export * as other ...` is part of the re-exports extensions (#86), but you get where I'm going with it.

Rich-Harris:
Cool - have released this as 0.14.1. 

> static resolution of namespaces within the bundle

Ah, I see. Interesting. You're right, this doesn't do that.

PR #87, opened 2015-08-20
Better readme

Rich-Harris:
([Rendered](https://github.com/rollup/rollup/blob/better-readme/README.md))

This PR tries to do a better job of explaining the value of Rollup, and removes all the outdated 'you probably shouldn't use this' warnings. The CLI and API examples have been moved to the wiki, so that there's a central place for documentation.

It also adds an MIT license file.

Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ Rollup can be used via a [JavaScript API](https://github.com/rollup/rollup/wiki/
 
 ## A next-generation ES6 module bundler
 
-When you're developing software, it's much easier to break your library or application apart into separate pieces that you can work on separately. It's also very likely that you'll have dependencies on third party libraries. The result is lots of small files – but that's bad news for browsers, which get slowed down by having to make many requests.
+When you're developing software, it's much easier to break your library or application apart into separate pieces that you can work on separately. It's also very likely that you'll have dependencies on third party libraries. The result is lots of small files – but that's bad news for browsers, which get slowed down by having to make many requests. (It's also [bad news for Node!](https://kev.inburke.com/kevin/node-require-is-dog-slow/))
 
-The solution is to write you code as **modules**, and use a **module bundler** to concatenate everything into a single file. [Browserify](http://browserify.org/) and [Webpack](http://webpack.github.io/) are examples of module bundlers.
+The solution is to write your code as **modules**, and use a **module bundler** to concatenate everything into a single file. [Browserify](http://browserify.org/) and [Webpack](http://webpack.github.io/) are examples of module bundlers.
 
 So far, so good, **but there's a problem**. When you include a library in your bundle...
 
-```
+```js
 var utils = require( 'utils' );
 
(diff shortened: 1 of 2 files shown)


Victorystick:
Wow. Rollup's all grown up. :)

The Licence section of the old README was particularly outdated. (Not that there is any code here...)

Some comments aside, I see the point you were trying to make in your linked Medium article, and I hope you'll write that follow-up soon.

README.md
```diff
@@ -1,161 +1,87 @@
...
+
+  foo();
+})();
+```
 
-* Support for legacy module formats
-* Respect for original formatting and code comments
+A minifier can detect that `foo` gets called, but that `bar` doesn't. When we remove `bar`, it turns out that we can also remove `baz`.
 
+Unfortunately, **traditional modules – CommonJS and AMD – make this kind of optimisation next to impossible**. Rather than *excluding dead code*, we should be *including live code*. That's only possible with ES6 modules.
```
eventualbuddha: You and I know why traditional modules make this next to impossible, but the reader of this may not. Should we explain or link to an explanation? It probably wouldn't take more than a sentence or two to explain.

README.md
```diff
@@ -1,161 +1,87 @@
...
-  bundle.write({
-    dest: 'bundle.js', // also writes sourcemap
-    format: 'amd'
-  });
-});
-```
+Most libraries that you depend on aren't written as ES6 modules, so Rollup can't work with them directly. (You *can* bundle your own app or library code with Rollup as a CommonJS module, then pass the result over to Webpack or Browserify, of course.)
 
-The duplication (`rollup.rollup`) is intentional. You have to say it like you're in the circus, otherwise it won't work.
+**You can help!** It's possible to write libraries as ES6 modules while still making it easy for other developers to use your code as they already do, using the [jsnext:main](https://github.com/rollup/rollup/wiki/jsnext:main) field in your package.json. You'll be writing your code in a more future-proof way, and helping to bring an end to the [dark days of JavaScript package management](https://medium.com/@trek/last-week-i-had-a-small-meltdown-on-twitter-about-npms-future-plans-around-front-end-packaging-b424dd8d367a).
```
eventualbuddha: This call to action is good, but I think we need an example project that shows how this should be done. I created one that you can link to if you like: https://github.com/eventualbuddha/rollup-starter-project. If you have suggestions on how to improve that starter project (i.e. use `rollup-babel`?) please open an issue or PR!

README.md
Victorystick: An example with the kinds of dynamic exports made possible by those module formats would be nice. Eg.
```js
// What does this module export?
functionCall().forEach( [ name, value ] => exports[ name ] = value );
```

eventualbuddha: Yeah, that's a good illustration.

Rich-Harris:
Thanks, both good points – will update to reflect those shortly

PR #89, opened 2015-08-21
add minifier example and link to starter project

Rich-Harris:
Follows #87. I thought it might be better *not* to include an example of dynamic exports, because that's the kind of bad habit that ES6 modules forbid but which some developers maintain is an advantage of legacy formats!


eventualbuddha:
:+1:

PR #90, opened 2015-08-21
Exclude unused function declarations...

Rich-Harris:
...that happen to mutate things that are used. This is a more conservative version of #49, which poses slightly harder challenges. No sense in letting perfect be the enemy of good, etc etc.

Essentially, if ensures that `unused` doesn't get included in situations like this:

```js
var obj = {};

function unused () { obj.mutated = true; }

// or alternatively
var unused = function () { obj.mutated = true; }
```


eventualbuddha:
I don't pretend to understand exactly what `readDepth` and `writeDepth` do, but this seems good based on the tests passing.

Rich-Harris:
Yeah, I'm not great at naming things :grimacing:

`readDepth` allows us to distinguish between strong and weak dependencies (i.e. `class A extends B` indicates `A` has a strong dependency on `B`, whereas `function foo () { return bar }` indicates `foo` has a weak dependency on `bar`, because the code doesn't run immediately), and `writeDepth` determines whether `foo.bar = baz` should be deemed to mutate `foo` based on whether it's at the top level or inside a function.

Very much open to any suggestions on how to clarify this stuff in the code!

Issue #92, opened 2015-08-21
Re-writing `foo` as `exports.foo` inline is often wasteful


Rich-Harris:
To keep bindings live, Rollup rewrites this...

```js
export let count = 0;
export function incr () { count++; }
```

...as this:

```js
exports.count = 0;
function incr () { exports.count++; }

exports.incr = incr; 
```

But in most cases, that's unnecessary – if an exported variable is never updated or reassigned, there's no need to turn `veryLongDescriptiveName` into `exports.veryLongDescriptiveName` every time it's encountered. Instead it would be better to keep it as a (mangleable) variable name and export it at the end.

PR #94, opened 2015-08-22
Prevent function expressions being incorrectly renamed

### LINKED ISSUE #91: Shadowed exported variables
Given a situation like this...

```js
var foo;

foo = (function () {
  return function foo () {
    return 42;
  };
})();

export { foo };
```

...we end up with this, which is a syntax error:

```js
exports.foo = (function () {
  return function exports.foo () {
    return 42;
  };
})();
```

The offending line is [here](https://github.com/rollup/rollup/blob/f1d35aad7e9b0567cf54cb55e339a097bfa3ac07/src/Statement.js#L328-L331). I suspect the answer is to be smarter about creating new scopes, i.e. doing it at the `BlockStatement` level rather than the `FunctionDeclaration`/`FunctionExpression` level.

Rich-Harris:
This fixes #91, but also #92 because it includes PR #93 (had to merge `gh-92` into `gh-91` because the fix for #92 also 'fixed' #91, but only superficially, and I needed to figure out a better test case).

Essentially it simplifies how scopes are attached to the AST. A scope is always attached to a `BlockStatement` node, regardless of whether it's a function scope. We need to jump through some slightly different hoops to prevent e.g. function arguments and function expression names from being rewritten, but on balance it ends up being slightly less work than what we had before.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 [these people](https://github.com/rollup/rollup/graphs/contributors)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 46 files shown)


Victorystick:
Didn't you say that we should create getters for the bundle exported names? You mentioned using getters, to ensure correctly updated bindings. I expect the issue is something like this?

```js
// main.js
var a = 0;
export function bump() { a++; }
export { a, a as b, a as c };

// generated-without-getters.js
exports.a = exports.b = exports.c = 0;

exports.bump = function bump() {
  exports.a = exports.b = exports.c = ( exports.a++ );
};

// generated-with-getters.js
var a = 0;
exports.bump = function bump() { a++; };

var aGetterDesc = { get: function () { return a; } };

Object.defineProperties(exports, {
  a: aGetterDesc,
  b: aGetterDesc,
  c: aGetterDesc
});
```
I guess that could be a separate task.

Rich-Harris:
It's related but turned out to be a separate thing – essentially when you have a named function expression...

```js
var fn = function factorial ( x ) {
  return x <= 2 ? x : x * factorial( x - 1 );
}
```

...`factorial` is considered part of the function's scope (i.e. the function name and the reference to `factorial` in the function body should always be the same thing), but Rollup was incorrectly attaching it to the parent scope. This PR fixes that, but also includes #93 which ensures that `export var foo = 1` gets rewritten as `exports.foo = 1` *only* if the value of `foo` could change after the module is evaluated.

The thing with the getters was purely about multiple exports sharing a binding. I much prefer your without-getters example (for one thing it works in old IE...) – there might be some places where it gets tricky (e.g. `var a = b++`, where `b` is exported) but it would be great to do if it's possible without going down the more costly transform/codegen route.

Anyway, this should be good to merge if it looks okay?

Victorystick:
Go ahead. :+1:

Issue #96, opened 2015-08-22
Missing side-effects when rolling up acorn


Rich-Harris:
This follows #88 and #95. Rollup doesn't account for this pattern:

```js
import {Parser} from './state';

/* some code happens... */

const pp = Parser.prototype

pp.initialContext = function() {
  return [types.b_stat]
}

/* a bunch more methods get added to pp... */
```

Rollup is smart enough to recognise that changes to `Parser.prototype` should be included, if we're dependent on `Parser`, but not smart enough to recognise that changes to `pp` should also be included. So we need to expand the search for 'modifier statements' to include aliases (where `pp` is treated as an alias of `Parser` by virtue of the fact that `Parser` is one of the one-or-more names contained in the `right` node of an `AssignmentExpression` where the `left` node is an `Identifier` whose `name` is `pp`, or the `init` node of a `VariableDeclarator` whose `id.name` is `pp`), and aliases of aliases. Tricky but doable (he said confidently).

It might yield false positives unless later on we distinguish between reassignments and mutations – at the moment they're essentially regarded as the same thing. In other words:

```js
// if this is included
var a = something;

var b = a;

// this should be included, because `b` is an alias of `a`, and is mutated
b.prop = 'whatever';

// but this shouldn't be, because we can't indirectly mutate `a` like this
b = c;
```

But that can be solved later; false positives are a lower priority than false negatives.

This isn't a complete solution, because (and I suppose deep down I always knew this...) it's possible for imported-but-ignored modules to have side-effects:

```js
// main.js
import { foo } from './foo';
import { bar } from './bar';

assert.ok( foo.wasMutated );

// foo.js
export var foo = {};

// bar.js
import { foo } from './foo';

var f = foo;
f.wasMutated = true;

export var bar = 'whatever';
```

Rollup would ignore `bar.js` because the `bar` binding is never used in `main.js`, so it would never 'discover' the side-effect even if it knew what to do with it. This might be a rare (and ill-advised) pattern, but if it exists in the wild we have to deal with it: the entire dependency graph as described by `import` statements will need to be loaded and parsed, regardless of whether the bindings described by those imports are actually used. (This would only increase bundle size in situations where side-effects do exist; the reason it's regrettable is purely that it will slow Rollup down a bit in cases where it was previously able to get away with not parsing some code.)

[...]

PR #97, opened 2015-08-23
Load all modules and syncify

Rich-Harris:
This is the first part of the plan to address #96 – loading *all* modules specified by `import` statements, regardless of whether a given module is destined to end up in the bundle. This is a prerequisite for identifying certain side-effects.

As a consequence we can rewrite a number of methods to be synchronous (and therefore shorter and easier to read), since we only need the async stuff during the initial loading phase.


Victorystick:
This seems great! I've wanted a change like this but been afraid to suggest it. Thought there might have been some clever reason for the lazy-loading.

PR #99, opened 2015-08-29
RFC: Use Scopes for Bundles and Modules

Victorystick:
*Not ready for merge*
(Sorry for so many commits. Please ignore JSONModule for now.)

My second attempt at using Scopes. This time, both the implementation and interface is much simpler.

All tests pass except ~10, including a new optimization for internal namespace resolution.

```js
// if foo, bar and quux are a chain of namespaces
// and a is a function in quux
foo.bar.quux.a();

// becomes just
a();
```

There are a lot of changes, but bare with me. Some of the changes may be introduced incrementally, and it might be best to break the changes down into smaller pieces. I thought it'd be possible to do in a single push, but maybe not.

Additionally, I'm interested in specifying an interface for Modules, so that a plugin/transformer can create specialized modules for different types of input. By combining the namespace optimization described above, with a specialized module for e.g. JSON, it would be possible to bundle parts of JSON files along with a bundle. (I'm thinking of the version field of package.json in particular. It could be inlined into the code during bundling.) If such a Module interface is defined, other types of files could be bundled in the same way as JSON. Templates, YAML, you name it.

Please comment.


Rich-Harris:
> I thought it'd be possible to do in a single push, but maybe not.

I know what you mean, these sorts of changes are frustratingly difficult to do piecemeal but somewhat overwhelming to do in a single go. I guess my question is what the explicit goal is here? Are there functional improvements beyond static namespace resolution (which I agree is definitely worthwhile)? Might help us figure out the best plan of attack.

> Additionally, I'm interested in specifying an interface for Modules, so that a plugin/transformer can create specialized modules for different types of input

How does this differ from the existing transform hook? A JSON transformer could look like this:

```js
import { rollup } from 'rollup';
import { extname } from 'path';

function jsonTransformer ( source, id ) {
  if ( extname( id ) !== '.json' ) return source;

  const data = JSON.parse( source );
  return Object.keys( data ).map( key => {
    return `export const ${key} = ${JSON.stringify(data[key])};`
  }).join( '\n\n' );
}

// alternatively, depending on usage
function jsonDefaultTransformer ( source, id ) {
  return extname( id ) === '.json' ? `export default ${source}` : source;
}

rollup.rollup({
  entry: 'main.js',
  transform: [ jsonTransformer ]
}).then(...);
```

Admittedly the transform hook needs to be a bit more sophisticated – a transformer deciding whether it applies or not based on `id` is brittle if you have a chain of them, and it currently breaks sourcemaps.

Victorystick:
The Scope (perhaps it should be called virtual scope) simplifies the bindings between names. See [test/testScope.js](https://github.com/Victorystick/rollup/blob/scope2/test/testScope.js) for some examples of how it works. The point is to allow several different scopes to exist without conflicts in the same real scope. Both the exports and locals of all internal and external modules live together in the bundle scope. It also simplifies deconfliction by handling it at one central location. The deconflict method can be supplied an optional renaming function, and could be used for minification.

Doing a `lookup` for some name yields an Identifier,  which is just an object with the properties `name` and `originalName`. By making modules implement the Identifier interface, we can `lookup` modules as well. In the case of a member expression, if we lookup an internal module, we can continue doing lookups to optimize the access.

Victorystick:
I actually implemented a JSON transformer, like the first one you just presented, to feel my way around the transformer API. It works alright. (I was thinking of adding some code to make transforms accesible from the CLI, similarly to Browserify plugins. `rollup -t json-transform`, but decided not to rush it.)

The biggest advantage of doing it as a JSONModule (or any kind of data module) is together with namespace optimization. Your approach works well for fields at the top level, where unused fields are excluded from the bundle. But the approach I'm suggesting with modules and namespace optimization works equally well, even for deeply nested data.

Consider this `package.json`:
```json
{
  "name": "example",
  "contributors": [
    {
      "name": "Dev Eloper"
    },
    {
      "name": "Ed Itor"
    },
    {
      "name": "Gal O'Code"
    }
  ]
}
```
and this script:
```js
// main.js
import package from './package.json';

console.log( package.contributors[2].name );

// Becomes, with NSO:
var name = "Gal O'Code";

console.log( name );
```
Additionally, we can check that the field being accessed actually exists at bundling-time!

Victorystick:
I'm passing all but `top-level-side-effect-on-imported`and `top-level-side-effects-are-preserved` locally.

Skipped:
* Since the renaming algorithm has changed, `internal-conflict-resulution` fails because it tests the exact resulting names. In this case the `bar`s and `_bar`s are simply reversed.
* `shorthand-properties` fails for the same reason. Since `foo` is never rewritten, `{ foo }` stays as it is. But the test expects the result to be `{ foo: _foo }`.

Victorystick:
I've rebased the branch on top of master to include your changes. It works really well. :)

I've encountered an issue you might help me resolve. Statements can only depend on names in the current scope, which means that for a module the statement depends on to be included, the Statement's Module must import the Identifier directly. The Statement then marks this name by calling `module.mark(name)`.

I've tried to change this so that individual identifiers are marked instead: `module.locals.lookup( 'name' ).mark();`. This works well enough. The statement in module defining the binding referred to locally as 'name' get included, but the module itself is never included in the bundle.

Perhaps it'd be easiest for a Statement to add the id of the module defining the binding to its own modules strong-/weak dependencies straight away?

Please, take a look.

Rich-Harris:
Cool! We could just swap `bar` and `_bar` around in that failing `form` test. We should reinstate the `shorthand-properties` test another way if possible.

> Perhaps it'd be easiest for a Statement to add the id of the module defining the binding to its own modules strong-/weak dependencies straight away?

Sounds sensible. I don't know off-hand what it would entail, but I can't think of an obvious reason not to do that. Maybe that's what we'd have done in the first place if not for all the late/lazy loading of modules.

> The biggest advantage of doing it as a JSONModule (or any kind of data module) is together with namespace optimization.

I'll be honest, this sounds like a mythical use case! I've never personally encountered a situation where I was inlining data from a JSON file and I didn't want all of it (either because I was using all of it, or because I needed to access stuff dynamically).

Victorystick:
Woo!! Passing! :tada:

Victorystick:
Fixed missing optimizations for namespaces. :+1:

@Rich-Harris It's finally time for a review.

Victorystick:
@eventualbuddha, thoughts?

src/Scope.js
```diff
@@ -0,0 +1,152 @@
+import { blank, keys } from './utils/object';
+
+// A minimal `Identifier` implementation. Anything that has an `originalName`,
+// and a mutable `name` property can be used as an `Identifier`.
+function Identifier ( name ) {
```
eventualbuddha: This is more a question of style, but should this be a `class`? I know it's essentially acting as a struct, so I could get behind a rule that says "use `function` for struct-like objects and `class` for class-like objects".

Victorystick: I'll adapt the code to whatever convention is preferred for Rollup's source code. I chose `function` since I thought it'd be clearest.

src/Bundle.js
```diff
@@ -193,7 +107,15 @@ export default class Bundle {
 				this.modules.push( module );
 				this.moduleById[ id ] = module;
 
-				return this.fetchAllDependencies( module ).then( () => module );
+				return this.fetchAllDependencies( module ).then( () => {
+					// Analyze the module once all its dependencies have been resolved.
+					// This means that any dependencies of a module has already been
+					// analysed when it's time for the module itself.
+					//
+					// FIXME: Except for cyclic dependencies...
```
Rich-Harris: is this still a FIXME?

src/Scope.js
Rich-Harris: I'm happy either way, so we may as well keep it as is

Rich-Harris:
I don't pretend to have completely wrapped my head round the whole thing, but it looks good. `JSONModule` is still in the PR but doesn't appear to be used anywhere, should that be removed?

Other than that, and the comment above re FIXME, I say :shipit:

Victorystick:
Yea, the `JSONModule` shouldn't be there, and the FIXME shouldn't be there. It works. :+1:

Your call to merge. :smile:

PR #109, opened 2015-09-11
Update umd.js to fix undefined global error

JonathanWolfe:


Rich-Harris:
Thanks @JonathanWolfe – can I ask where you're encountering this error? Am wondering if this risks creating a situation where neither `this` nor `window` exist, in which case the `window` reference would cause another ReferenceError. Though I suppose if you fall all the way down that rabbit hole then you end up with [this](https://github.com/lodash/lodash/blob/38a37f1b18979f3fb687b3a903a60e5591ed9905/lodash.js#L268-L295)...

eventualbuddha:
When does this happen? What environments have `this` undefined globally but `window` defined?

JonathanWolfe:
Uglify is the culprit for me. it renames the `this` in the IIFE call to undefined. This fixes that potential compile error.

eventualbuddha:
@JonathanWolfe that's because Babel assumes all input are ES6 modules and rollup assumes UMD output is for scripts, not modules. Would [rollup-babel](https://github.com/rollup/rollup-babel) work for you?

JonathanWolfe:
Didn't know that existed. I'll try it.

Rich-Harris:
> rollup-babel doesn't run babel on the files.

That's... weird! Could you share your configuration?

An alternative solution, if we can't get this to work, would be to use Rollup to generate an ES6 module then hand that off to Babel.

JonathanWolfe:
@Rich-Harris: Sorry, I nuked it already.

I still think the PR is good for safety reasons.

Rich-Harris:
@JonathanWolfe Cool. Git wouldn't let me apply your patch to a new branch so I opened a new PR, #127, which does the same thing and updates the tests. I'm still not totally sold on the idea, so I'll give other people a chance to weigh in

Issue #107, opened 2015-09-11
Sourcemap-aware transform sequence


Rich-Harris:
Via https://github.com/rollup/rollup-babel/issues/1:

> the `transform` sequence should always supply a string to the transformer and accept either a string or a `{ code, map }` object in return, and Rollup should take care of combining sourcemaps

At the moment it's a [dumb reduce](https://github.com/rollup/rollup/blob/b8894dd9014c2a6d54a66673f321b3acfb9821d2/src/utils/load.js#L7-L9), which leaves individual transformers at the mercy of whatever came before them. It's also tied to the `load` function, meaning that custom loaders (e.g. in-browser, from cache, whatever) have to reimplement `transform` (rollup-babel [suffers from this](https://github.com/rollup/rollup-babel/blob/2e0d91567aff40ce17eb76c472c670004fe72e12/index.js#L63-L82)).

Couple of other idle thoughts while I'm on the subject:

* it'd be nice if transformers were configurable. E.g. we might want to use an off-the-shelf transformer (not that such things currently exist) but only on files that match a certain minimatch pattern
* Long-term it'd be nice to be able to specify transforms via the CLI, a la Browserify transforms

callumlocke:
Is this issue only about combining sourcemaps coming from 'transforms'?

What about when I'm not using rollup's transforms feature, but I'm using rollup as part of a sequence of build steps:

```
$ babel original.js | rollup | browserify > bundle.js
```

(I haven't bothered to write all the flags in but you get the point...) I've tried a variation of the above, and the final sourcemap seemed to take me back to the *output* of Babel, but not all the way back to the original file (when I debug in a browser, the code looks like Babel output). So it seems like browserify correctly notices the incoming sourcemap from rollup and combines it, but rollup doesn't do this.

Am I correct this hasn't been implemented or am I doing something wrong? If it's not implemented, should there be a separate issue for it?

Rich-Harris:
[...]

PR #111, opened 2015-09-12
RFC: Resolver fallbacks

### LINKED ISSUE #108: RFC: use fallbacks for resolveId and resolveExternal
As also discussed in #107, custom resolvers must re-implement the default resolve behavior for ids that they do not handle. It'd be better to be able to delegate to another implementation, and use it as a fallback.

```js
rollup.rollup({
  resolveId: function ( id, importer ) {
    if ( specialcase ) return 'special-special';

    // result is undefined, i.e. the resolver can't handle this...
  },
  ...
}).then( ... );
```

Proposal:
1. If a resolver returns a string or an eventually resolved promise, that value is used.
2. If a resolver returns undefined, or an eventually rejected promise, we fallback to another implementation.

This would allow for an API a bit like the one discussed in #1.
```js
rollup.rollup({
  resolveId: [ jspmResolver, npmResolver ],
  ...
}).then( ... );
```

We try to do things the JSPM way, if that fails the NPM way, and finally Rollup's implicit default.

Victorystick:
It would be really nice to be able to "fall back" to some other implementation of `resolveId`, `resolveExternal` etc. This PR (which should be considered an RFC) includes one possible way to implement this in a backwards-compatible manner.

Let the resolvers handle what they can, and delegate everything else to some other function. It may not do anything to help #107, but could be something in the right direction. Thought?

Fixes #108

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -1,5 +1,6 @@
 import { Promise } from 'sander';
 import MagicString from 'magic-string';
+import attempt from './utils/attempt.js';
 import { blank, keys } from './utils/object';
 import Module from './Module';
 import ExternalModule from './ExternalModule';
@@ -17,12 +18,16 @@ export default class Bundle {
 		this.entry = options.entry;
 		this.entryModule = null;
 
-		this.resolveId = options.resolveId || defaultResolver;
-		this.load = options.load || defaultLoader;
+		this.resolveId = ensureArray( options.resolveId )
+			.reduce( attempt, defaultResolver );
+
+		this.load = ensureArray( options.load )
+			.reduce( attempt, defaultLoader );
 
 		this.resolveOptions = {
 			external: ensureArray( options.external ),
-			resolveExternal: options.resolveExternal || defaultExternalResolver
+			resolveExternal: ensureArray( options.resolveExternal )
+				.reduce( attempt, defaultExternalResolver )
 		};
 
 		this.loadOptions = {
--- a/src/utils/attempt.js
+++ b/src/utils/attempt.js
@@ -1,0 +1,11 @@
+// Given a `fallback` and a `preferred` function,
+// return a function that attempts to call the `preferred`.
+// If it fails, or returns undefined, call the fallback and return its value.
+export default function attempt ( fallback, preferred ) {
+	return function ( ...args ) {
+		const boundFallback = () => fallback( ...args );
+
+		return Promise.resolve( preferred( ...args ) )
+			.then( res => res === undefined ? boundFallback() : res, boundFallback );
+	};
+}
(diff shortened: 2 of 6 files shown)


Rich-Harris:
Nice. I think this is the right direction. And yeah, re #107, transforms are probably a different matter – for one thing they're synchronous unlike `resolveId`, `resolveExternal` and `load` which can be sync or async, and for another *all* transforms need to be applied, whereas in this case we're going with the first successful result. Will add some inline comments

src/utils/attempt.js
```diff
@@ -0,0 +1,11 @@
+// Given a `fallback` and a `preferred` function,
+// return a function that attempts to call the `preferred`.
+// If it fails, or returns undefined, call the fallback and return its value.
+export default function attempt ( fallback, preferred ) {
+	return function ( ...args ) {
```
Rich-Harris: doesn't this mean that later results take priority over earlier ones? Surely we want to use the first successful result? If I do...

```js
one = () => 1;
two = () => 2;
fallback = () => 'fallback';

fn = [ one, two ].reduce( attempt, fallback );
fn().then( result => console.log( result ) );
```

...it logs `2`, when I'd want it to log `1`. What if it was this instead?

```js
function first ( candidates ) {
  return function ( ...args ) {
    return candidates.reduce( ( promise, candidate ) => {
      return promise.then( result => result || Promise.resolve( candidate( ...args ) ) );
    }, Promise.resolve() );
  }
}
```

src/utils/attempt.js
```diff
@@ -0,0 +1,11 @@
...
+		const boundFallback = () => fallback( ...args );
+
+		return Promise.resolve( preferred( ...args ) )
+			.then( res => res === undefined ? boundFallback() : res, boundFallback );
```
Rich-Harris: I don't think we should treat an error the same as no result, since then there'd be no way for a function in the middle to indicate there was a problem – I think errors should fall through

src/Bundle.js
```diff
@@ -17,12 +18,16 @@ export default class Bundle {
 		this.entry = options.entry;
 		this.entryModule = null;
 
-		this.resolveId = options.resolveId || defaultResolver;
-		this.load = options.load || defaultLoader;
+		this.resolveId = ensureArray( options.resolveId )
```
Rich-Harris: using `first` (see below) instead of `attempt`, this'd presumably look something like this:

```js
this.resolveId = first( ensureArray( options.resolveId ).concat( defaultResolver ) );
```

(or the fallback could be a second argument, and the `first` function would do the concatting...)

Victorystick:
Hmm. Yes, it should be the other way around; using `first` like you suggest. It'd probably be nice to be able to abort a resolver chain.

```
rollup --resolveExternal npm,jspm --resolveId relativeToBase,fail
```

A configuration like this requires all module imports to be relative to the project base, and that it should reject any id that isn't.

We could do like gobble and load packages like `rollup-external-jspm` etc. when we specify command line arguments like above.

I also wonder if we could, or maybe should, use an ES6 version of [node-browser-resolve](https://github.com/defunctzombie/node-browser-resolve) or similar, since it's been battle-tested for quite some time.

Issue #112, opened 2015-09-12
Functions that mutate exports in closure are removed


trbrc:
If I have two files that look like this, `src/index.js`:
```
import array from './module';
array[0]();
```

and `src/module.js`:
```
const array = [];
for (let i = 0; i < 10; i++) {
    array.push(() => i);
}
export default array;
```

then run
```mkdir -p tmp && babel src -b es6.modules,strict -d tmp && rollup -f iife -i tmp/index.js -o bundle.js```

Now, `tmp/module.js` looks like this:
```
var array = [];

var _loop = function (i) {
    array.push(function () {
        return i;
    });
};

for (var i = 0; i < 10; i++) {
    _loop(i);
}
export default array;
```

But the output in `bundle.js` does not include the loop at all:
```
(function () { 'use strict';

    var array = [];

    array[0]();

})();
```

If I change the loop in `module.js` to simply `push(i)`, it works though. It seems that Rollup is not able to detect that the function Babel creates to get correct scoping for `i` will mutate `array`.

The only way I found to deal with this is to do something like this:

```
const array = [];
(function (array) {
    for (let i = 0; i < 10; i++) {
        array.push(() => i);
    }
})(array);
export default array;
```

Any chance that Rollup will be able to handle this without workarounds?

**Comment by Victorystick** [comment]
That's definitely a bug. Odd. I thought this'd already be taken care of. You shouldn't need any workarounds.

I'll put together a failing test for it, unless Rich gets there first!

**Comment by trbrc** [comment]
Ok, good to hear. If it helps, here's a minimal example to trigger the bug, without the incidental Babel stuff.

`index.js`
```javascript
import value from './module';
value;
```

`module.js`
```javascript
var value = 'original';
function change() {
    value = 'changed';
}
change();
export default value;
```

`$ rollup index.js`
```javascript
var value = 'original';

value;
```

Victorystick:
Oops! My bad. Seems like [we don't handle](https://github.com/rollup/rollup/blob/master/src/Statement.js#L136-L146) this particular case yet. Will have to think through a solution for how to approach this pattern.

Rich-Harris:
Thanks for raising this and posting a test. I suppose we should include
top-level CallExpression statements (top-level including inside blocks and
IIFEs - I think we already have a convenient way of tracking this), and
assignments to globals. That should catch these sorts of side effects. Am
on vacation at the moment though so have limited bandwidth, both
figuratively and literally :grinning: Will take a look in a few days if
[...]

PR #113, opened 2015-09-13
Added a failing test for #112

Victorystick:


lukeapage:
Have you considered having a safe mode that doesn't remove unused code? It makes me a little nervous on a large project that bugs like this exist.. and any future bugs cannot be worked around.

Rich-Harris:
@lukeapage it's crossed my mind, yeah. I've resisted it so far because I think we're close to identifying all the various rules you need to follow to safely include the minimum code necessary, and because it's easier to find the rough edges when they're forcibly exposed! But if it comes to it then I don't think it'd be too difficult to add a more conservative mode.

lukeapage:
Okay, good to know. Once this is fixed I might see if I can see anything not covered by tests.

PR #114, opened 2015-09-14
Export all from bundle

Victorystick:
I noticed today that it wasn't possible to from the entry `export * from './internal.js'`. This PR fixes that.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -49,7 +49,6 @@ export default class Bundle {
 
 		this.statements = null;
 		this.externalModules = [];
-		this.internalNamespaceModules = [];
 	}
 
 	build () {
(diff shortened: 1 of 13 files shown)


caridy:
is `default` export from `./internal.js` going to be re-experted or not when transforming `export * from './internal.js'` with this PR?

Rich-Harris:
Yes, default exports are re-exported – I've updated the tests to reflect that

caridy:
@Rich-Harris that's the problem, that's why I was asking for this, because the specification is very clear about this. Default export SHOULD NOT be reexported when using `export * from ...`. Details are in step 6.c here: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-resolveexport

Rich-Harris:
d'oh! Have raised #123

Victorystick:
I'll address this when I get back from my vacation.

PR #127, opened 2015-09-21
guard against undefined `this` in post-processed UMD bundle - fixes #109

Rich-Harris:
Similar to #109, but a slightly more compact form.

For:

* keeps things working if a UMD bundle is subsequently processed with Babel, or something else that rewrites `this` as `undefined`.

Against:

* It adds extra bytes to the bundle
* Rollup isn't actually doing anything wrong here (in an ES5 environment, `this` is always the global object at the top level AFAIK), so it's really just masking a problem rather than fixing it

Any thoughts?

### DIFF
--- a/src/finalisers/umd.js
+++ b/src/finalisers/umd.js
@@ -40,7 +40,7 @@ export default function umd ( bundle, magicString, { exportMode, indentString },
 			typeof exports === 'object' && typeof module !== 'undefined' ? ${cjsExport}factory(${cjsDeps.join( ', ' )}) :
 			typeof define === 'function' && define.amd ? define(${amdParams}factory) :
 			${defaultExport}factory(${globalDeps});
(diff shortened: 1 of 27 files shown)


Victorystick:
It seems like a fair addition to make transpilation work, but it is a kludge.

Can't Babel just be told to compile a non-module, where global `this` is defined?

Rich-Harris:
It [turns out you can](https://babeljs.io/docs/faq/), yes:

> Babel assumes that all input code is an ES2015 module. ES2015 modules are implicitly strict mode so this means that top-level `this` is not `window` in the browser nor is it `exports` in node.

> If you don't want this behaviour then you have the option of disabling the `strict` transformer:

```bash
$ babel --blacklist strict script.js
```

```js
require("babel").transform("code", { blacklist: ["strict"] });
```

So I'll close this.

PR #129, opened 2015-09-22
add bundle.modules - fixes #128

### LINKED ISSUE #128: export "modules" as well as "externalModules" in the user-facing Bundle object.
I'm writing a plugin to perform incremental builds to multiple targets using rollup, and it'd be very handy if the "modules" property was exported along with the "externalModules" property. Currently "externalModules" is exported as "imports".

This is handy when trying to identify if a modified file will affect a specific target (i.e. is it in the dependency tree for my given target) - if so, I want to rebuild that target, but ignore unaffected targets.

I'm not sure what an appropriate name would be - internalImports/internalDependencies?

I'd be happy to make this change myself :)

Rich-Harris:
This PR adds a `bundle.modules` property, where each module is an object with an `id` property (which in normal circumstances is the module's resolved path), and room for future growth. See #128

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -17,6 +17,9 @@ export function rollup ( options ) {
 		return {
 			imports: bundle.externalModules.map( module => module.id ),
 			exports: keys( bundle.entryModule.exports ),
+			modules: bundle.orderedModules.map( module => {
+				return { id: module.id };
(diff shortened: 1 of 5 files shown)


Victorystick:
Seems good. :+1:

PR #130, opened 2015-09-22
Add code coverage with Istanbul and Codecov

Rich-Harris:
https://codecov.io/github/rollup/rollup?branch=istanbul

This PR adds code coverage reports via [Istanbul](https://gotwarlost.github.io/istanbul/) and [codecov.io](https://codecov.io). Couple of questions/things to consider before merging:

* It instruments the built file (dist/rollup.js), not the src files, which is ignored in this repo – meaning we can't actually *see* on codecov.io which bits aren't covered
* Is there some sourcemap magic that'll fix that? I don't know
* Should dist/rollup.js be checked in? Should the coverage reports be generated on `npm test` automatically and checked in, so they can be viewed on GitHub? Or is that daft?
* `npm install` inexplicably runs `prepublish` scripts, meaning that Travis CI ends up having to run the build and tests twice. This is an inexplicably bad decision. I really hope npm fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 .gobble*
 dist
 _actual
+coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,5 @@ node_js:
 env:
   global:
     - BUILD_TIMEOUT=10000
+install: npm install --ignore-scripts
+script: npm run ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
   - "0.12"
   - "iojs"
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ node_js:
 env:
   global:
     - BUILD_TIMEOUT=10000
-install: npm install --ignore-scripts
+install: npm install
 script: npm run ci
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -8,7 +8,8 @@ var node = src
 		entry: 'rollup.js',
 		dest: 'rollup.js',
 		format: 'cjs',
-		external: [ 'sander', 'acorn' ]
+		external: [ 'sander', 'acorn' ],
+		sourceMap: true
 	})
 	.transform( 'babel' );
 
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
     "rollup": "./bin/rollup"
   },
   "scripts": {
+    "pretest": "npm run build",
     "test": "mocha",
+    "pretest-coverage": "npm run build",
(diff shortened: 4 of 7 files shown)


Swatinem:
You currently run rollup+babel for the dist build.
An alternative would be to run babel-istanbul + rollup, to get instrumentation into the built file.
Then you need to capture the `__coverage__` variable, write it to a .json file and run the reporting step of babel-istanbul.

It is ugly and inconvenient, I know :-(

lukeapage:
> Is there some sourcemap magic that'll fix that? I don't know

https://github.com/gotwarlost/istanbul/issues/212#issuecomment-139366862

Rich-Harris:
Thanks both! I've rigged up remap-istanbul, and it works like a charm. (It forced me to get Rollup's sourcemap working as well, which was long overdue.) Awesome!

Swatinem:
Hm, `magic-string` is covered as well.
And might I suggest uploading the lcov report via
```yaml
after_success:
  - bash <(curl -s https://codecov.io/bash) -f ./coverage/lcov.info
```
the reason is that with the lcov upload, codecov can generate branch&function coverage stats.

Swatinem:
> @Swatinem I don't think they have it yet! I think the newest version they have is iojs 3.3.1

travis uses nvm for node version selection, and that already has 4:
https://travis-ci.org/Swatinem/yavd/builds/81485964#L79-L86 (don’t mind the broken build, haha)

Rich-Harris:
Ah, okay – will update .travis.yml with node 4. I saw the magic-string thing and removed `node_modules/*`, so coverage is now showing as 94%: https://codecov.io/github/rollup/rollup – hooray for sourcemaps making that possible.

I don't quite understand the difference with lcov – does istanbul need to generate an lcov report that remap-istanbul then remaps, or can remap-istanbul generate an lcov file from `coverage-final.json`? Forgive my ignorance, a bit new to this stuff

Victorystick:
This is fantastic! With this we should be able to add another shield and instill more confidence in Rollup.

Rich-Harris:
The best part is that we can find bits of code that aren't used any more – I've found a few blocks that were necessary at one point, then the design changed slightly and suddenly they're sitting there unused, but in a way that would be very hard to detect otherwise. See #135

Swatinem:
haven’t looked at remap-istanbul too much. but it can create all the reports that istanbul can. So just let it create lcov files…

lukeapage:
@Rich-Harris 

> npm install inexplicably runs prepublish scripts, meaning that Travis CI ends up having to run the build and tests twice. This is an inexplicably bad decision. I really hope npm fixes that.

I doubt they will do that.. I tracked down this issue and discussion:

https://github.com/npm/npm/issues/3059#issuecomment-12282746

Rich-Harris:
@lukeapage yeah, I've come across that issue. Makes me weep. I mean, when *everyone* is telling you that you're wrong, maybe it's time to reconsider your position (which at this point seems to be 'fixing it would confuse people who rely on the broken behaviour')?

PR #138, opened 2015-09-24
Changed the default deconfliction function to behave more like Babel.…

Victorystick:
… Should be much faster for many identical identifiers.

Deconflicting five `foo`s, yields
```
foo
_foo
_foo1
_foo2
_foo3
```
instead of
```
foo
_foo
__foo
___foo
____foo
```

### DIFF
--- a/src/Scope.js
+++ b/src/Scope.js
@@ -27,9 +27,23 @@ function isntReference ( id ) {
 	return !( id instanceof Reference );
 }
 
-// Prefix the argument with '_'.
-function underscorePrefix ( x ) {
-	return '_' + x;
+// Returns a function that will prefix its argument with '_'
+// and append a number if called with the same argument more than once.
+function underscorePrefix () {
+	function number ( x ) {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Not sure I get why it's faster? But it's certainly nicer

Victorystick:
It's much faster in the worst case, where there are lots and lots of identifiers with the same name.

Given _n_ `foo`s:
1. the first `foo` has no conflicts so it can be kept as is.
2. the second `foo` conflicts only with the first, so it is changed to `_foo`.
3. the third conflicts first with `foo`, so the renaming function changed it to `_foo` which in turn conflicts with the second `foo` that was renamed. So it must become `__foo`.

So the time to deconflict _n_ `foo`s was previously O(_n<sup>2</sup>_), while it should be O(_n_) now. The renaming function was changed to always generate a unique identifier name.

Rich-Harris:
Oh, right! D'oh.

PR #141, opened 2015-09-24
Polish optimization of namespace lookups

### LINKED ISSUE #139: Bad generated code (missing import * as x)
This is a reproduction case from a real failing...

entry.js
```
import a from './foo';

a();
```

foo.js
```
import * as a from './bar';

function b() {
  console.log(a.foo, a.bar);
}

export default b;
```

bar.js
```
function foo() {

}
function bar() {

}
export { foo, bar };
```

Victorystick:
* Explain better how optimization of namespace lookups are done in comments.
* Consider `hasReplacements` to be true if there exists any namespace replacements.
* Improve error messages slightly

I have added a test-case based on #139, but I'm not sure what to call it.

Fixes #139

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -25,12 +25,19 @@ function isFunctionDeclaration ( node, parent ) {
 	if ( node.type === 'FunctionExpression' && parent.type === 'VariableDeclarator' ) return true;
 }
 
+// Extract the property access from a MemberExpression.
+function property ( node ) {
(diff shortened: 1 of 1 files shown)


lukeapage:
awesome, thanks for the quick fix

Rich-Harris:
> but I'm not sure what to call it.

If in doubt, 'gh-139' #yolo

Issue #132, opened 2015-09-23
Incorrect module order


Rich-Harris:
Modules are ending up in a funny order following some recent changes – [this test](https://github.com/rollup/rollup/tree/module-order/test/function/module-sort-order) has the details. Essentially it looks as though modules are claiming a dependency on other modules that mutate their values, which is technically accurate but which breaks the module sorting algorithm.

The offending code is [here](https://github.com/rollup/rollup/blob/c903bb2e4e7f7b4fc0378945f4478b58020e29ff/src/Module.js#L414-L417), but it can't removed without re-introducing a bug that was previously handled [here](https://github.com/rollup/rollup/blob/0df5d9b1775d617512f59238b7e26841a3b119c4/src/Bundle.js#L228-L268) – this method included all the necessary code but without impacting module order. @Victorystick any idea how to fix?

Victorystick:
I'll look in to it. :+1:

Rich-Harris:
@Victorystick have you had any thoughts about how to solve this? I've tried to come at it a few different ways, to no avail (I don't think I quite grok all the new stuff around the `Scope` and `Identifier` classes yet!). At the moment we're shipping a broken library, so hopefully we can figure it out between us... :grimacing:

Victorystick:
Hmm. In `bundle.sort`, when I log the `strongDeps` structure I get the same structure no matter if I include [this](https://github.com/rollup/rollup/blob/c903bb2e4e7f7b4fc0378945f4478b58020e29ff/src/Module.js#L414-L417) or not.

```
main.js
  needs -> z.js
z.js

c.js
 
b.js
 
a.js
  needs -> z.js
```
The sort order shouldn't be effected by `weakDependencies` anyway, right?

Victorystick:
The `weakDependencies` helps to decide what modules _are_ included, and not _the order_ they're included, right?

The issue seems to be related to the `ordered` array.
> unordered is actually semi-ordered, as [ fewer dependencies ... more dependencies ]

This used to hold before, but the previous step doesn't seem to preserve the invariant anymore. The module order is `[ 'a', 'b', 'c', 'z', 'main' ]` after the first step, compared to `[ 'b', 'c', 'z', 'a', 'main' ]`.

But as you can see, `a` actually _has_ a strong dependency on `z`, unlike `c`, `b` and `z`. But it gets sorted first despite of this.

Victorystick:
[...]

PR #151, opened 2015-09-27
Contingency plan for #132

Rich-Harris:
Since 0.16.0, [Ractive.js is failing to build](https://github.com/ractivejs/ractive/issues/2159) because modules are in the wrong order. Given that there's been a number of fixes in Rollup since then, and more to come soon, pinning an old version isn't really a viable option.

Have tried to solve #132 in various ways, but am ashamed to say I've admitted defeat. This PR represents the contingency plan – reinstating the various recent fixes and new tests on top of 0.15.0 instead. The one thing it *doesn't* yet reinstate is static namespace optimisations (internal `import *` statements will continue to work, they just won't be as neat).

I know this is drastic, but unless there's a better way of fixing it – @Victorystick here's hoping you have something up your sleeve! – we might need to do this, at least temporarily, as #132 is a bug with no possible workaround.


Victorystick:
I understand. I didn't realize the severity of the situation; shipping broken code. I should have prioritized #132 over the other issues. I'll make it top priority. Can't work on it today, but I'll do what I can tomorrow.

Good of you to prepare a contingency plan.

Rich-Harris:
Cool, thanks. It's my bad for not managing to wrap my head round the new stuff. That said, now that I've faced my first real test with it, I *do* think it adds quite a bit of complexity – e.g. we have two `Scope` classes doing conceptually related but different things, [monkey-patching](https://github.com/rollup/rollup/blob/master/src/Module.js#L93-L116), 'virtual' scopes, and more indirection. I totally understand the goals, I just think it necessitates too many extra moving parts. What concerns me is that potential future contributors might have the same problem understanding it all as I did.

The 0.15 model is that you have a single `Bundle` object made of a number of `Module` objects which contain `Statement` objects, which only care about their immediate surroundings and are mediated by the higher levels – adding things like `Identifier` and `Reference` into the mix is arguably a lot more elegant insofar as it means there's a way to represent a value present throughout a bundle by different names, but (as far as I can tell) you lose that straightforward* hierarchical data structure (or at the very least you duplicate it). That's the reason it's tying my brain in knots! (Of course you might feel exactly the same way looking in the other direction...) I don't know if there's a way to get the best of both worlds?

*<i>may not actually be 'straightforward' – there's some irreducible complexity in what we're doing, no matter how you slice and dice it</i>

Victorystick:
The monkey patching is unfortunate, and I want to remove it at the first opportunity. This was the easiest way to get the required behavior without introducing any additional classes. (Although maybe in this case I should have. It wouldn't have been that complex regardless.)

I understand that you want to keep Rollup's code structure simple and easy to understand, ideally with only a few classes. I agree with that goal. However, when the project complexity grows, I think it's best to add a few more classes, to mange the complexity, so long as they have clear interface boundaries.

The transition to the new Scope class hasn't been painless, and I regret that, but I do believe in the end result.

I think the idea of virtual scopes isn't all too complicated, but I really should try to explain it better. I'll get back to you on that.

Rich-Harris:
I'm having a little trouble reducing the [issue referenced here](https://github.com/rollup/rollup/pull/156#issuecomment-143848707) to something self-contained – should we publish a new version from this branch for the time being?

Victorystick:
I think that might be best. Those who take Rollup for a spin, or indeed depend upon it, should get something that works.

Rich-Harris:
I've published 0.17.0 from this branch

Victorystick:
Closing since the release of 0.18.x.

Issue #148, opened 2015-09-26
incorrect error on double import


lukeapage:
I think this might be from the recent fix

```
// bar.js
export default function bar() {
}

// foo.js
export { default as bar } from "./bar";

// zoo.js
import * as foo from './foo';

function wah() {
  foo.bar();
}

wah();

// entry.js
import * as foo from './foo';
import './zoo';

export default {
  foo: foo
};
```

error is..

```
Error: Module "c:/git/grunt-rollup/test/fixtures/foo.js" doesn't export "bar"!
    at walk.leave (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1881:14)
    at visit (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1517:3)
    at visit (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1512:4)
    at visit (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1512:4)
    at visit (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1509:5)
    at visit (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1512:4)
    at walk (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1463:2)
    at Statement.analyse (c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:1808:4)
    at c:\git\grunt-rollup\node_modules\rollup\dist\rollup.js:2499:14
    at Array.forEach (native)
```

I can't see a problem with it though ? unless its the wrong error message and there is something else I am missing.

again stumbled on this whilst trying to reproduce another bug but it could be related so I'll wait on reporting that till this is fixed.

lukeapage:
If I change entry.js to..
```
import './zoo';
```

then it all compiles fine...

oh and this is on master, not 1.16.4

Victorystick:
It's a recent change designed so that we can better track when an undefined export is referenced. I'll investigate the issue today, and see what I can do.

Victorystick:
The issue is caused by the order in which the modules are analysed. Due to how the modules depend on each other (and how they are loaded), `zoo` is analysed before `foo`, meaning that `foo`'s exports are undefined when `zoo` attempts to access them. This could be avoided by splitting the analysis into two stages:

1. Where all imports/exports are defined and bound.
2. Where all statements are analysed with respect to these bindings.

Rich-Harris:
[...]

PR #152, opened 2015-09-27
Optimise multiple namespaces in a function

lukeapage:
I hinted in #148 that I was getting many namespaces not optimised. I thought I'd take a look after @Victorystick 's work in #150 (which didn't fix the issue).

I've cherry-picked that PR into here - I suggest that you merge that first and then I'll rebase so you can see just my changes.

I wasn't 100% sure it was the right change - at first I thought the `topNode` logic was in reverse and it should be `node.object !== topNode` but that breaks an existing test and actually seems correct - the problem then appeared to be that it was assumed that a "statement" would only have one namespace optimisation in it - except a statement can be a function declaration, so as in the test any member expressions in the function causes it to not optimise.

### DIFF
--- a/src/optimise/namespace-lookup.js
+++ b/src/optimise/namespace-lookup.js
@@ -39,7 +39,10 @@ export default function ( statement ) {
 					const id = statement.module.locals.lookup( localName );
 
 					// It only counts if it exists, is a module, and isn't external.
-					if ( !id || !id.isModule || id.isExternal ) return;
+					if ( !id || !id.isModule || id.isExternal ) {
+						topNode = null;
+						return;
+					}
 
 					namespace = id;
 				}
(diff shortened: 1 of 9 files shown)


lukeapage:
Added a test enhancement and then changed to a simpler fix, that seems to work, but I'm not 100% sure why..
Will await your feedback on which one is correct!

Victorystick:
I honestly didn't think of resetting `topNode` when we couldn't find a namespace. That's just what we need. Thanks for noticing and fixing it!

lukeapage:
ok, rebased and squashed.

Victorystick:
LGTM! :+1:

lukeapage:
@Rich-Harris this doesn't apply any more.. do you want the tests?

Rich-Harris:
@lukeapage never any harm in having more tests – have added these ones. Thanks!

Issue #154, opened 2015-09-27
namespace optimisation ignores variable scopes


lukeapage:
```
// entry
import * as bar from './bar';

function car() {
  var bar = {
    foo: function() {
    }
  };
  bar.foo();
}

// bar.js
export function foo() {};
```

output:
```
function foo() {}

function car() {
  var bar = {
    foo: function() {
    }
  };
  foo();
}
```

expected output:
```
function foo() {}

function car() {
  var bar = {
    foo: function() {
    }
  };
  bar.foo();
}
```

lukeapage:
similarly (and worse/more likely)
```
// entry.js
import * as bar from './bar';

function car() {
  var foo = function() {};
  bar.foo();
}

// bar.js
export function foo() {};
```

results in..
```
function foo() {}

function car() {
  var foo = function() {};
  foo();
}
```

lukeapage:
@Rich-Harris have you considered refactoring rollup into a babel plugin?

advantages:
1. a variable scope management and renaming system for free
2. take advantage of existing plugins, e.g. namespace inlining is a special case of the babel inline plugin, in fact so is inlining the contents of files - you could focus on getting es6 modules into a 100% correct ast and then use inlining plugins to inline and dead code elimination to remove the bits of modules not used
3. most people using es6 need to transpile already. this would avoid reading the ast twice
4. automatic support for every file format babel supports

Sorry if you find the suggestion insulting, rollup is a great package and IMO this technique is the future of dependency management. You are probably well past this stage, but I thought I'd suggest it as even now it could save some work.

Rich-Harris:
@lukeapage not an insult! Though I think you're underestimating the complexity of it :grinning: My sense is that if it was doable, it would have been done, or at least someone would be working on it.

To my knowledge, no-one is. The evidence is that it comes up on Babel's issue tracker repeatedly – https://github.com/babel/babel/issues/495, https://github.com/babel/babel/issues/1118, https://github.com/babel/babel/issues/1681, https://github.com/babel/babel/issues/1806. In fact Sebastian [recommends using Rollup here](https://github.com/babel/babel/issues/1681#issuecomment-136581780).

[...]

PR #155, opened 2015-09-27
test for #154 (shadowed namespaces)

Rich-Harris:
Test for #154


lukeapage:
I've been thinking about this and I think the 2 test cases I provided might have to be fixed in different ways,  so it will be worth adding both of them..

Rich-Harris:
Thanks @lukeapage – this works in 0.18

PR #156, opened 2015-09-28
Module order side effects

### LINKED ISSUE #132: Incorrect module order

Victorystick:
I've been all over with this. As I said in [one of the comments](https://github.com/rollup/rollup/issues/132#issuecomment-143752767), I'm not entirely sure what should be included in `weakDependencies`, but I think this should do it.

Adding the `dependencies` (the imported/re-exported modules) of the current module as weak dependencies, seems to do the trick. We make sure to include the modules we depend on (makes sense), and so on recursively. Any statements that have been marked within that traversal will be included.

The tests pass, but as we've discovered, they're not yet all that thorough at this point.
@Rich-Harris Does this get the Ractive build passing again? :pray: 

Fixes #132

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -281,7 +281,7 @@ export default class Bundle {
 		let seen = blank();
 
 		let ordered = [];
-		let hasCycles;
+		let hasCycles = false;
 
 		// Map from module id to list of modules.
 		let strongDeps = blank();
--- a/src/Module.js
+++ b/src/Module.js
@@ -41,14 +41,15 @@ class Id {
 }
 
 class LateBoundIdPlaceholder {
-	constructor ( module, name ) {
+	constructor ( module, name, local ) {
 		this.module = module;
 		this.name = name;
 		this.placeholder = true;
+		this.local = local || false;
 	}
 
 	mark () {
-		throw new Error(`The imported name "${this.name}" is never exported by "${this.module.id}".`);
+		throw new Error( `The name "${this.name}" is never ${ this.local ? 'defined' : 'exported' } by "${this.module.id}".` );
 	}
 }
 
@@ -104,7 +105,6 @@ export default class Module {
 				}
 			}
 
-			// throw new Error( `The name "${name}" is never exported (from ${this.id})!` );
 			this.exports.define( name, new LateBoundIdPlaceholder( this, name ) );
 			return reference.call( this.exports, name );
 		};
@@ -203,6 +203,13 @@ export default class Module {
 				node.specifiers.forEach( specifier => {
 					const localName = specifier.local.name;
 					const exportedName = specifier.exported.name;
+
+					// HACK: Fix Ractive builds.
+					// If we try to export something that doesn't exists in the current
+					// local scope, create a placeholder for it.
+					if ( !this.locals.defines( localName ) ) {
+						this.locals.define( localName, new LateBoundIdPlaceholder( this, localName, true ) );
+					}
 
 					this.exports.bind( exportedName, this.locals.reference( localName ) );
 				});
--- a/src/Module.js
+++ b/src/Module.js
@@ -359,7 +359,7 @@ export default class Module {
 	}
 
 	consolidateDependencies () {
-		let strongDependencies = blank();
+		const strongDependencies = blank();
 
 		function addDependency ( dependencies, declaration ) {
 			if ( declaration && declaration.module && !declaration.module.isExternal ) {
@@ -397,7 +397,7 @@ export default class Module {
 			}
 		});
 
-		let weakDependencies = blank();
+		const weakDependencies = blank();
 
 		this.statements.forEach( statement => {
 			keys( statement.dependsOn ).forEach( name => {
@@ -407,19 +407,22 @@ export default class Module {
 			});
 		});
 
+		// Make us weakly depend on all our dependencies.
+		// If we don't, these modules (and their side-effects) won't be included.
+		this.dependencies.forEach( source => {
+			const module = this.getModule( source );
+
+			if ( !module.isExternal ) {
+				weakDependencies[ module.id ] = module;
+			}
+		});
+
 		// Go through all our local and exported ids and make us depend on
-		// the defining modules as well as
-		this.exports.getIds().concat(this.locals.getIds()).forEach( id => {
+		// the defining modules.
+		this.exports.getIds().concat( this.locals.getIds() ).forEach( id => {
 			if ( id.module && !id.module.isExternal ) {
 				weakDependencies[ id.module.id ] = id.module;
 			}
-
-			if ( !id.modifierStatements ) return;
-
-			id.modifierStatements.forEach( statement => {
-				const module = statement.module;
-				weakDependencies[ module.id ] = module;
-			});
 		});
 
 		// `Bundle.sort` gets stuck in an infinite loop if a module has
--- a/test/function/import-of-unexported-fails/_config.js
+++ b/test/function/import-of-unexported-fails/_config.js
@@ -4,7 +4,7 @@ module.exports = {
 	description: 'marking an imported, but unexported, identifier should throw',
 
 	error: function ( err ) {
-		assert.equal( err.message.slice( 0, 50 ), 'The imported name "default" is never exported by "' );
+		assert.equal( err.message.slice( 0, 41 ), 'The name "default" is never exported by "' );
 		assert.equal( err.message.slice( -10 ), 'empty.js".' );
 	}
 };
--- a/test/function/module-sort-order/_config.js
+++ b/test/function/module-sort-order/_config.js
@@ -1,5 +1,3 @@
 module.exports = {
-	// solo: true,
-	// show: true,
 	description: 'module sorter is not confused by top-level call expressions'
 };
--- a/test/function/module-sort-order/_config.js
+++ b/test/function/module-sort-order/_config.js
@@ -1,0 +1,5 @@
+module.exports = {
+	// solo: true,
+	// show: true,
+	description: 'module sorter is not confused by top-level call expressions'
+};
--- a/test/function/module-sort-order/a.js
+++ b/test/function/module-sort-order/a.js
@@ -1,0 +1,20 @@
+import { b } from './b';
+import z from './z';
+
+z();
+
+const p = {
+	q: function () {
+		b.nope();
+	}
+};
+
+(function () {
+	const p = {
+		q: function () {
+			b.nope();
+		}
+	};
+})();
+
+export default 42;
--- a/test/function/module-sort-order/a.js
+++ b/test/function/module-sort-order/a.js
@@ -3,14 +3,14 @@ import z from './z';
 
 z();
 
-const p = {
+var p = {
 	q: function () {
 		b.nope();
 	}
 };
 
 (function () {
-	const p = {
+	var p = {
 		q: function () {
 			b.nope();
 		}
--- a/test/function/module-sort-order/b.js
+++ b/test/function/module-sort-order/b.js
@@ -1,1 +1,1 @@
-export const b = function () {};
+export var b = function () {};
--- a/test/function/module-sort-order/b.js
+++ b/test/function/module-sort-order/b.js
@@ -1,0 +1,1 @@
+export const b = function () {};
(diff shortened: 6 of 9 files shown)


Victorystick:
Now that I think about it a bit more, this should handle the situation you described in #96, where side-effects of _imported-but-ignored modules_ get left out. Due to the crudeness of this patch, that issue just might sort itself out.

Rich-Harris:
> Does this get the Ractive build passing again? :pray:

I'm afraid not! With 0.16.x it would build, but into non-sensical JavaScript – now it won't build at all (Cannot reference undefined identifier "isArray"). Looking into it

Victorystick:
This morning I cloned Ractive and got to work on the build error. Once I'd figured out what caused them, it wasn't too difficult to fix. The Ractive build started _working_.

```
> ractive@0.8.0-edge build /Users/oskar/github/ractive
> sh ./scripts/build.sh

> linting...
> emptying tmp dir...
> building Ractive...
gobble: build started                                                                                                  
gobble: 03-rollupBabel done in 3845ms                                                                          
gobble: 04-rollupBabel done in 3231ms                                                                          
gobble: 05-rollupBabel done in 2590ms                                                                          
gobble: 06-rollupBabel done in 2783ms                                                                          
gobble: 07-merge done in 23ms                                                                                  
gobble: 14-replace done in 7ms                                                                                 
gobble: 08-moveTo done in 4ms                                                                                  
gobble: 09-merge done in 2ms                                                                                   
gobble: 10-babel done in 3274ms                                                                                
gobble: 02-babel done in 4502ms                                                                                
gobble: 11-merge done in 5ms                                                                                   
gobble: 12-bundleTests done in 1919ms                                                                          
gobble: 15-moveTo done in 1ms                                                                                  
gobble: 16-include done in 6ms                                                                                 
gobble: 17-babel done in 423ms                                                                                 
gobble: 18-moveTo done in 1ms                                                                                  
gobble: 19-merge done in 9ms                                                                                   
gobble: 20-moveTo done in 11ms                                                                                 
gobble: 21-merge done in 5ms                                                                                   
gobble: built to tmp in 23851ms                                                                      
> running tests...
> running node.js-specific tests. working directory is /Users/oskar/github/ractive

  [․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․]%cRactive.js %c0.8.0-edge %cin debug mode, %cmore... color: rgb(114, 157, 52); font-weight: normal; color: rgb(85, 85, 85); font-weight: normal; color: rgb(85, 85, 85); font-weight: normal; color: rgb(82, 140, 224); font-weight: normal; text-decoration: underline;
You're running Ractive 0.8.0-edge in debug mode - messages will be printed to the console to help you fix problems and optimise your application.

To disable debug mode, add this line at the start of your app:
  Ractive.DEBUG = false;

To disable debug mode when your app is minified, add this snippet:
  Ractive.DEBUG = /unminified/.test(function(){/*unminified*/});

Get help and support:
  http://docs.ractivejs.org
  http://stackoverflow.com/questions/tagged/ractivejs
  http://groups.google.com/forum/#!forum/ractive-js
  http://twitter.com/ractivejs

Found a bug? Raise an issue:
  https://github.com/ractivejs/ractive/issues


  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬․․․․․․․․․․․․․․․․․․]%cRactive.js: %cIf supplied, options.data should be a plain JavaScript object - using a non-POJO as the root object may work, but is discouraged color: rgb(114, 157, 52); color: rgb(85, 85, 85);
%cRactive.js: %cIf supplied, options.data should be a plain JavaScript object - using a non-POJO as the root object may work, but is discouraged color: rgb(114, 157, 52); color: rgb(85, 85, 85);
  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]

  302 passing (618ms)

phantomjs is ready. running tests...
Error: Attempted to call a non-existent method ("activate")

  file:///Users/oskar/github/ractive/tmp/ractive-legacy.js:9283 in fire
  file:///Users/oskar/github/ractive/tmp/ractive-legacy.js:9403
Tests completed in 4186 milliseconds. 2300 assertions of 2300 passed, 0 failed.
> tests passed. minifying...
  minified ractive.runtime.js
  fixing ractive.runtime.min.js sourcemap (/Users/oskar/github/ractive/tmp)
  fixed ractive.runtime.min.js sourcemap
  minified ractive-legacy.runtime.js
  fixing ractive-legacy.runtime.min.js sourcemap (/Users/oskar/github/ractive/tmp)
  fixed ractive-legacy.runtime.min.js sourcemap
  minified ractive.js
  fixing ractive.min.js sourcemap (/Users/oskar/github/ractive/tmp)
  minified ractive-legacy.js
  fixing ractive-legacy.min.js sourcemap (/Users/oskar/github/ractive/tmp)
  fixed ractive.min.js sourcemap
  fixed ractive-legacy.min.js sourcemap
> emptying build folder...
> copying library to build folder...
> copying *.json files to build folder...
> build complete
```

Rich-Harris:
:metal: woot!

Can you run me through what's going on with the block marked `HACK`?! Makes me a little nervous :grinning:

Victorystick:
It isn't really a hack. It boils down to my changing `Scope.reference` from creating a logic-less placeholder if a name isn't defined, to it throwing an Error. I must manually create a placeholder. The advantage of this approach is that we can put logic in `mark` to make missing definitions easier to find.

Victorystick:
@Rich-Harris Did you manage to reproduce the build?

Rich-Harris:
Yes, it does work for me as well. I have to be honest though, it feels really kludgy having separate `Id` and `LateBoundIdPlaceholder` classes for what are basically the same thing, depending on whether the `export` statement is encountered before the definition (which is how Babel rewrites modules). Shouldn't we be generating a complete picture of the scope before trying to bind everything together? This seems like one of the advantages of having a more straightforward hierarchical structure without these sorts of bindings.

Victorystick:
I appreciate your honest opinion. Criticism is needed for something to improve.

I'd also very much like to avoid `LateBoundId` if possible. Perhaps we could have a `isDefined` flag that we check in `mark()` to see whether or not to throw. Not a particularly pretty solution, but maybe better than two arguably similar classes?

I do believe that something like it is required when two modules reference each other. Since one module will be analysed before the other it won't be able to bind it's imports to the exports of the other.

`even.js`
```js
import odd from './odd';

export default function ( num ) {
  if ( num == 0 ) return true;
  return !odd( num - 1 );
}
```
`odd.js`
```js
import even from './odd';

export default function ( num ) {
  if ( num == 1 ) return true;
  return !even( num - 1 );
}
```

I agree that it is far from perfect, but I _do_ believe that this (or something similar) is a step in the right direction.

The reason I like the `Scope` and `Id` model is that we have easy access to not just the name of an identifier, but also metadata and logic. If a statement depends on some identifiers, making sure they're included in the bundle is as easy as `locals.lookup( name ).mark()`. By making `Module` implement the `Id` interface, it doesn't event matter if it's a simple variable or a namespace (module) that's being included.

We have immediate access to other metadata, such as `modifierStatements`. If it is an empty array, we can be sure that the identifier is never modified. If we add a `isConstant` flag, we can be sure that a variable isn't redefined. Having this data available in different parts of the system allows us to be more clever about what we do. The namespace lookup optimisation (NLO), although maybe not all that common, is only one possibility. I wouldn't know how to do it with the current approach.

Below is an example of how the NLO could be extended. Since modules are constant, we can alias namespace imports using `const`.
**Disclaimer**: _I understand that this particular case isn't a priority._
```js
import * as foo from './foo';

// Since we can know if `foo.bar` is a namespace or not
// this `const` could bind directly to the namespace.
// Instead of having to mark all exports (as we currently would have to)
// we could remove the statement entirely, and even optimise the `bar.baz` below.
const bar = foo.bar;

bar.baz( 1337 );
```

Rich-Harris:
Cool – as I say, I do get the motivation, and it's important to try these things out. I've been thinking a bit about what the 'best of both worlds' scenario looks like, and I think I have an idea. Here's a high-level sketch:

1. Parse and generate scope info, as presently, but capture all references (e.g. `var a = b` generates a `Reference` around the `b` AST node)
2. Do another pass of the AST to discover *declarations* (e.g. `scope.addDeclaration(node)` generates a new `Declaration`). Ignore imports and exports, we're not binding anything yet
3. The bundle then links each `Reference` to its corresponding `Declaration`
4. We mark statements that need to be included (those with side-effects, including being exported from the bundle, and the statements they depend on), generating the module dependency graph as we go
5. The bundle deconflicts all included top-level declarations. Because each declaration is aware of its references, it's relatively trivial to rewrite those references. Deshadowing is the only complication, but that would also be made easier – in this system, we could have a `scope.deshadow('foo')` method

Declarations would be a little bit smarter – for example, a function declaration would know that nothing could strongly depend on it because of hoisting, unless we were reading/writing a property of the function. A reference could be to `foo.bar`, so if it turned out that `foo` was a namespace, optimisation would be easy (but the reference doesn't need to *know* whether or not it's to a namespace). Crucially, imports and exports are 'dumb' – the only binding that cuts across the `Bundle`–`Module`–`Statement` hierarchy is directly between references and declarations, mediated by the bundle.

How does that sound? Am tempted to have a crack at it this morning.

Victorystick:
(This is a quick explanation I wrote a while ago. Didn't want to throw it away. Don't know if it helps at all, but...)

### Id
Fundamentally, the `Id` interface just requires an object to have a `name` property. That's the only thing that `Scope` needs to handle deconfliction. Anything else is extra information that we can use to our advantage.

`Module.js` defines an implementation of `Id`, which includes the defining `module`, `statement`, a list of `modifierStatements` and a `mark` method. When `mark` is called, everything that is needed for the identifier to be properly defined and modified will be included.

The same goes for the `Module` itself, where `mark()` ensures that the modules `exports` can be dynamically accessed. It marks a boolean flag, and marks all exported identifiers. It can't get much easier than that.

### Scope
New `Scope`s are created by passing an optional parent. This creates an _actual_ scope, isolated from all others. This scope can then be shared while appearing to be isolated. This is a _virtual_ scope, created using the `virtual` method (taking an optional `inheritParent` boolean). Creating virtual scopes for the `exports` and the `locals` of modules, allows all of them to live in a single, shared scope. The `locals` inherit the parent (which is the global scope), and the `exports` don't.

Each virtual scope has a set of `names` which are used within the scope. These names can either be defined in the scope or referenced from it. The actual `Id`s and `Reference`s are stored in the `ids` array, which is shared by all virtual scopes sharing a scope. When we define or bind something that does not exist in the `names` of a scope, we define the name to index a unique slot in the `ids` array. That slot will later contain either a `Reference` or an `Id`. At deconflict time, we iterate through the `ids` array do two things.

1. The names of all `Reference`s to identifiers in higher scopes get to _in use_.
2. We give all `Id`s a unique name that isn't currently _in use_.

Victorystick:
@Rich-Harris The plan for best-of-both sounds good. I'm really excited to hear it! I also considered doing an early pass for declarations. It's a good thing they're top-level, since it should be fast and we can do something like this to get at them!
```js
this.statements.filter( s => /Declaration/.test(s.node.type) ).forEach( ...
```

Victorystick:
Closing since #163 got merged.

Issue #149, opened 2015-09-26
Bad variable rename on large project


lukeapage:
I'm trying rollup out on less.js and found a bad rename. I cannot reproduce simply for now, so at the moment my test case is the less.js repo

1. fetch https://github.com/less/less.js/tree/rollup
2. npm install
3. copy the latest master dist build into node_modules (the fix for the last bug I reported is needed)
4. grunt rollup

this creates a file `tmp/less.js`

You could also run rollup directly on it, I'm using `umd` and the entry file is `lib/less-browser/bootstrap.js`

on line 6242
```
    function _environment(externalEnvironment, fileManagers) {
```
on line 6259
```
    }environment.prototype.getFileManager = function (filename, currentDirectory, options, environment, isSync) {
```

from the original sources
```
export default function environment(externalEnvironment, fileManagers) {
...
...
environment.prototype.getFileManager = function (filename, currentDirectory, options, environment, isSync) {
```

so it has renamed the export default, but not the prototype function.

I've gone from commonjs to es6 modules, so there could be a mistake somewhere in the modules, but even if there is, it shouldn't cause this side effect..

lukeapage:
actually.. looking later in the file
on line 6281 from the same source file, its fine
```
    _environment.prototype.addFileManager = function (fileManager) {
```

it also seems a bit suspicious that its the only line thats screwed up whitespace. Is it possible that the newline character and _ character have been removed somewhere as part of the transformation?

lukeapage:
after manually adjusting and running the tests I've found a couple more programming errors which I fixed and pushed but they haven't had any effect on this (line numbers unchanged)

Victorystick:
I've reproduced the issue. 

> it also seems a bit suspicious that its the only line thats screwed up whitespace. Is it possible that the newline character and _ character have been removed somewhere as part of the transformation?

It's odd that only that line appears to be affected...

lukeapage:
adding
```
			if (id.name === "environment") {
				console.log("renaming", id.name, "to", name);
				for(var i = 0; i < id.modifierStatements.length; i++) {
					var modifierStatement = id.modifierStatements[i];
					console.log("at...")
					console.log(modifierStatement.module.source.substring(modifierStatement.node.start, modifierStatement.node.end));
				}
			}
```
in deconflict gives
```
renaming environment to _environment
at...
environment.prototype.getFileManager = function (filename, currentDirectory, options, environment, isSync) {

    if (!filename) {
[...]

PR #161, opened 2015-09-30
Fix for #149

lukeapage:
am looking for a fix, but might be some time, I'll keep this updated, let me know if you want to take over.

It seems that the parameter having the same name as the export is removing the rename in that scope.

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -398,8 +398,8 @@ export default class Statement {
 						return this.skip();
 					}
 
+					replacementStack.push( names );
 					names = newNames;
-					replacementStack.push( newNames );
 				}
 
 				if ( node.type === 'MemberExpression' ) {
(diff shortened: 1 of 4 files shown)


lukeapage:
some debugging info is on a `gh-149-debug` branch, I'm off for the day.
@Victorystick any pointers, or if you want to take over you're welcome. was going to next try and understand where scopes are created for function definitions and why the parameter is getting confused onto the global scope.

lukeapage:
so...
this: https://github.com/rollup/rollup/blob/6e839f2d38b8b0e058d990c283849af7ae237e27/src/Statement.js#L401

looks wrong to me - the reason being that the newscopes names are being popped into a parent scope on leaving the node. so, my first change was to change this to...

```
					replacementStack.push( names );
					names = newNames;
```

next up.. this line: 
https://github.com/rollup/rollup/blob/6e839f2d38b8b0e058d990c283849af7ae237e27/src/Statement.js#L435
stops parameters from being renamed.. but if a parameter is shadowing a global that is being introduced (as in this test case) the parameter needs to be renamed..

I had to also remove the check on the next line because otherwise the function declaration wasn't getting changed as a result of the first thing.

I guess this code is here to stop the function name in a function expression not getting renamed as per the global scope.

Nextup, the crux of the problem. params are added to the block scope underneath them here: https://github.com/rollup/rollup/blob/6e839f2d38b8b0e058d990c283849af7ae237e27/src/Statement.js#L82

this means that it cannot rename parameters because the scope is the wrong one when walking the function declaration parameters.

I tried a fix to put the scope on the function declaration and add the params there e.g.

```
					case 'FunctionDeclaration':
						scope.addDeclaration( node, false, false );
					case 'FunctionExpression':
						newScope = new Scope({
							parent: scope,
							block: false,
							params: node.params
						});

						// named function expressions - the name is considered
						// part of the function's scope
						if ( node.type === 'FunctionExpression' && node.id ) {
							newScope.addDeclaration( node, false, false );
						}
						break;

					case 'BlockStatement':
						newScope = new Scope({
							parent: scope,
							block: true
						});
						break;
```

Now that, combined with the other 2 fixes, results in my test passing... however...

now `renamed-arguments` fails... it seems to be because the param and function name are the same, it thinks the function name has been re-declared and therefore doesn't need renaming.

I worked around that by making the function name rename happen before applying the scope as a special case..

any comments welcome, I now have 3 tests failing

lukeapage:
am putting params back onto the body so that
```
function foo(foo) {
    var foo = foo;
}
```
works..

lukeapage:
okay, fixed it, not the simplest thing to work out, let me know what you think.

lukeapage:
@Rich-Harris I enhanced the test, do you want that? otherwise this can be closed

Rich-Harris:
Updated the test, thanks

Issue #144, opened 2015-09-25
Loose mode for import * as


lukeapage:
Hi,

When using `import * as` we get the following output

```
var a = {
    get foo() { return foo; }
}
```

which is "correct", however a shorter form would be
```
var a = { foo: foo };
```

Would you consider a option that didn't bother with writing getters.. since I very much doubt any code will depend on not being able to set things on that object?

Victorystick:
@Rich-Harris has [considered avoiding getters](https://github.com/rollup/rollup/blob/master/src/Bundle.js#L236-L243), at least for `exports` and ES3 environments. I don't think that it'd be possible in general, as it breaks with the ECMAScript spec for exports that get reassigned.

However, it would be possible to optimise `const` exports. For instance,
```js
export const foo = {};
````
when imported as `import * as a` _could_ generate
```js
var a = { foo: foo };
Object.defineProperty(a, 'foo', {
  enumerable: true,
  value: foo
});
```
since we know that `foo` isn't going to be reassigned. It isn't shorter exactly, but wouldn't include the overhead of a getter function.

lukeapage:
> I don't think that it'd be possible in general, as it breaks with the ECMAScript spec for exports that get reassigned.

Could you expand on that? I understand it would no longer follow the spec, thats why I am suggesting an option along the lines of babel's loose mode (and its strict mode = false) which make the code shorter/faster but don't follow the spec perfectly.

It would be more work for rollup, but I would imagine the code that generates this is in one or two specific places. I'd be interested in any downsides to that for the consumer or anything you think might break.

Victorystick:
With the loose more you're suggesting, Rollup could generate a bundle like the one below. The semantics state that the access of the `foo` variable should be the same wether it's through the namespace `a`, directly through the imported `foo`. But this is broken by the _loose mode_.

```js
//--- foo.js
export function nextFoo () { foo++; }
export var foo = 1;

//--- main.js
import * as a from './foo.js';
import { foo, nextFoo } from './foo.js';

nextFoo();

assert.equal( a.foo, foo );

//--- proposed.js
function nextFoo () { foo++; }
var foo = 1;

var a = { foo: foo };

nextFoo();

assert.equal( a.foo, foo );
```

Current Rollup will generate the below bundle, and completely optimise the namespace away.
```js
//--- current.js
function nextFoo () { foo++; }
var foo = 1;

nextFoo();

assert.equal( foo, foo );
```

lukeapage:
[...]

PR #162, opened 2015-10-04
Rewrite as per discussion on #156

Rich-Harris:
This PR is a large-scale rewrite of the codebase, following the discussion on #156. Essentially, it introduces a more explicit relationship between *declarations* and *references*, such that each reference is directly tied to the thing that it's referencing. This makes it more straightforward to rename references to avoid conflicts, and simplifies the handling of tricky cases with default/namespace exports and external modules.

As well as reducing the LOC count by a fair chunk, it enables two new features:

* Namespace blocks only use getters where necessary (#144)
* We can easily track mutations of variables that are aliases of other variables (#96)

Still a bit of tidying up and testing to do, but thought I'd leave it here in the meantime anyway. @Victorystick thanks for pushing us in this direction!


Rich-Harris:
Closing in favour of #163 (same rewrite, different target)

Issue #23, opened 2015-06-06
Command line tool error re: source-map-support


curran:
Hello,

I am using the rollup command line tool, and I noticed that it breaks with the following error:

```
./node_modules/.bin/rollup --output dist/reactive-model.js src/reactiveModel.js 
module.js:338
    throw err;
          ^
Error: Cannot find module 'source-map-support'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/curran/repos/reactive-model/node_modules/rollup/bin/runRollup.js:1:63)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
```

After `npm install source-map-support` everything works well, but I was just wondering if it might make sense to promote `source-map-support` from `devDependencies` to `dependencies` in your `package.json`?

Loving rollup so far, great work!

Rich-Harris:
Good catch, thanks! Fixed and released as 0.7.7.

curran:
Great! Thank you.

zygous:
I've just encountered exactly this issue in rollup version 0.46.0. I notice that `source-map-support` was removed from `dependencies` in commit c378f20e08aaf87deda0506c1ac97826465b1266; was that intentional?

PR #167, opened 2015-10-07
chalk and source-map-support are always needed

m0ppers:
Couldn't run rollup after a clean installation. I needed to install these two modules separately. I found a reference to an older bug: https://github.com/rollup/rollup/issues/23

However this seems to be broken (again?)


Rich-Harris:
Thanks! I forgot about the CLI – I thought they were only used in testing. And of course the absence of them didn't show up locally or in CI because devDependencies are present in both cases...

PR #173, opened 2015-10-09
split up export var a, b, c

### LINKED ISSUE #171: export var a, b, c doesn't get split up
This results in three 'synthetic nodes', allowing a single binding to be included in the bundle:

```js
var a = 1, b = 2, c = 3;
export { a, b, c };
```

But the same doesn't work if you do it like this:

```js
export var a = 1, b = 2, c = 3;
```

Rich-Harris:
fixes #171

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -208,6 +208,7 @@ export default class Module {
 
 		// export { foo, bar, baz }
 		// export var foo = 42;
+		// export var a = 1, b = 2, c = 3;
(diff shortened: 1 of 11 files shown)


Victorystick:
Looks good! I have to admit that I haven't checked the spec about this; just assumed that this is how it should be. Let me do that...

Victorystick:
Yup! (Would have been really surprised otherwise.)

PR #175, opened 2015-10-10
sourcemap aware transforms

Rich-Harris:
This needs some more battle-testing before it gets merged in, but this is a first step towards #107 – if all transformers return a `{ code, map }` object, the `map.mappings` will be traced so that the resulting sourcemap points to the correct locations in the original source.

Some things we'll have to worry about eventually:

* names, the fifth member of each segment. At present these are disregarded
* what to do if one of the transformers doesn't generate a sourcemap (e.g. it's just doing some simple string replacement to replace some `<@template@>` tags or something)?
* what to do if a sourcemap is imprecise (e.g. it just maps lines, rather than characters) – do we still try and trace segments? probably not

No reason they should delay this from being released though. *Should* make it possible to get sourcemaps working with rollup-babel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "estree-walker": "^0.1.3",
     "gobble": "^0.10.1",
     "gobble-babel": "^5.5.8",
-    "gobble-cli": "^0.4.2",
+    "gobble-cli": "^0.5.0",
     "gobble-rollup": "^0.10.0",
     "gobble-rollup-babel": "^0.4.0",
     "istanbul": "^0.3.20",
--- a/package.json
+++ b/package.json
@@ -59,9 +59,11 @@
     "source-map": "^0.5.1"
   },
   "dependencies": {
-    "minimist": "^1.1.1",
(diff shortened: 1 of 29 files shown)


Rich-Harris:
Merged #174 into this in the belief that rollup-babel would need resolver/loader fallbacks. It turns out it doesn't, but so be it.

One thing rollup-babel *does* need, which I've occasionally thought would be a good idea anyway, is the ability to insert content into the bundle in a sourcemap-aware way *inside* the format-specific wrapper (i.e. inside the IIFE or `define` call or UMD block). In this case, it allows us to inject `babelHelpers`. So this PR adds support for `intro` and `outro` options alongside `banner` and `footer` (which prepend/append content *outside* the wrapper, e.g. for license/version comments).

Have done a spot of testing with other libraries that use rollup-babel and sourcemaps appear to work beautifully – I'm going to push this out as 0.19.0 to make wider testing more practical (and because there have been a couple of other improvements since 0.18.5)

PR #205, opened 2015-10-21
Use all export declarations

### LINKED ISSUE #204: Exported namespaces must have `use()` called
See #203

This snippet
```js
// export * as ns from './mod';
import * as ns from './mod';
export { ns };
```

yields nothing but

```js
'use strict';

var ns = Object.freeze({
	a: a,
	b: b
});

exports.ns = ns;
```
resulting in a `ReferenceError: a is not defined`.

Victorystick:
Includes #203 
Fixes #204


Rich-Harris:
:+1:

Issue #189, opened 2015-10-14
Promise is not defined


devongovett:
Running on node v0.10.38, I get the following error:

```
ReferenceError: Promise is not defined
    at Bundle.resolveId (/usr/local/lib/node_modules/rollup/src/utils/first.js:9:6)
    at Bundle.build (/usr/local/lib/node_modules/rollup/src/Bundle.js:47:32)
    at Object.rollup (/usr/local/lib/node_modules/rollup/src/rollup.js:16:16)
    at bundle (/usr/local/lib/node_modules/rollup/bin/runRollup.js:51:16)
    at module.exports (/usr/local/lib/node_modules/rollup/bin/runRollup.js:40:3)
    at Object.<anonymous> (/usr/local/lib/node_modules/rollup/bin/rollup:35:26)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
```

Perhaps include a polyfill for old engines?

Victorystick:
We already _do_ include a polyfill, but must have forgot to reference it. I'll create a patch for it tomorrow, unless @Rich-Harris gets at it first. :wink:

PR #206, opened 2015-10-21
Release 0.19.2

Victorystick:
We might as well release 0.19.2 to include these fixes (#203, #205) along with the Promise related `ReferenceError` for those running node v0.10 (#189).


Rich-Harris:
agree

PR #207, opened 2015-10-21
[RFC] support plugins that return transform and intro functions

Rich-Harris:
I've created an experimental Babel plugin for testing (not yet released to npm): https://github.com/rollup/rollup-plugin-babel

The basic idea:

```js
rollup.rollup({
  entry: 'main.js',
  plugins: [ require( 'rollup-babel-plugin' )() ]
}).then(...);

// this is basically equivalent to
rollup.rollup({
  entry: 'main.js',
  plugins: [{
    transform: function ( code, id ) {
      // babelify code
    },
    intro: function () {
      // generate helpers to insert at top of bundle
    }
  }]
}).then(...);
```

So as far as Rollup is concerned, a plugin is just an object with various properties. As of this PR, the `transform` and `intro` properties are supported, because that's the bare minimum we need to support the Babel plugin. My feeling is that if we go down this road, we should add new hooks only as and when they become obviously necessary.

Does this seem like a sensible approach? Probably worth trying to build a couple of other plugins to test out the idea before committing to this interface, in any case (e.g. I've been wondering about a CommonJS -> ES6 plugin).

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -19,6 +19,8 @@ export default class Bundle {
 		this.entry = options.entry;
 		this.entryModule = null;
 
+		this.plugins = ensureArray( options.plugins );
+
 		this.resolveId = first( ensureArray( options.resolveId ).concat( defaultResolver ) );
 		this.load = first( ensureArray( options.load ).concat( defaultLoader ) );
 
@@ -28,7 +30,9 @@ export default class Bundle {
 		};
 
 		this.loadOptions = {};
-		this.transformers = ensureArray( options.transform );
+		this.transformers = ensureArray( options.transform ).concat(
+			this.plugins.map( plugin => plugin.transform ).filter( Boolean )
+		);
 
 		this.pending = blank();
 		this.moduleById = blank();
@@ -171,7 +175,14 @@ export default class Bundle {
 			}
 		});
 
-		if ( options.intro ) magicString.prepend( options.intro + '\n' );
+		const intro = [ options.intro ]
+			.concat(
+				this.plugins.map( plugin => plugin.intro && plugin.intro() )
+			)
+			.filter( Boolean )
+			.join( '\n\n' );
+
+		if ( intro ) magicString.prepend( intro + '\n' );
 		if ( options.outro ) magicString.append( '\n' + options.outro );
 
 		const indentString = getIndentString( magicString, options );


Rich-Harris:
Have created https://github.com/rollup/rollup-plugin-commonjs. Very basic and poorly tested, but a proof-of-concept – it works like so:

```js
// this...
var foo = require( './foo' );
module.exports = foo * 2;

// ...becomes this:
import foo from './foo';
let exports = {}, module = { exports: exports };
module.exports = foo * 2;
export default module.exports;
```

It wouldn't work with external packages in `node_modules`, because the default resolver would fail (upon not finding a `jsnext:main` field) before the conversion could take place. So it's a bit useless right now. I'm thinking we should probably remove that functionality from the core and move it into a plugin – something like this:

```js
rollup.rollup({
  entry: 'main.js',
  plugins: [
    require( 'rollup-plugin-npm' )({ jsnext: true, legacy: true }),
    require( 'rollup-plugin-commonjs' )()
  ]
}).then(...)
```

*Maybe* we could include some of the more common plugins as builtins:

```js
rollup.rollup({
  entry: 'main.js',
  plugins: [
    rollup.npm({ jsnext: true, legacy: true }),
    rollup.commonjs()
  ]
}).then(...)
```

Victorystick:
The changes are smaller than I thought they'd need to be to add support for plugins.

Do you think we could remove the `intro`/`outro` options in favor of a built-in plugin? If it reduces the complexity, and we choose to add logic for plugins, every little bit helps.

The CommonJS plugin in particular is a neat hack. Great for compatibility with all the existing packages. I look forward to trying it out.

Part of me wants Rollup (the library) to ship without any external resolvers or other plugins. That way Rollup itself status simple, pure and less bound to Node. The same goes for the default `load` function. A default external resolver doesn't make sense in a browser setting, we might as well load the plugin by-need. The (presumably Node based) CLI on the other hand would do well to include the NPM resolver (and perhaps CommonJS) by default.

Rich-Harris:
> Do you think we could remove the `intro`/`outro` options in favor of a built-in plugin?

I hadn't thought of that. I suppose that probably makes sense, since they were only added for the sake of rollup-babel.

> Part of me wants Rollup (the library) to ship without any external resolvers or other plugins.

Me too. The one thing I think we *do* need to be wary (or at least aware) of is creating a situation where package `a` imports package `b`, and `b` has its own plugin requirements that `a` somehow needs to take into account. Since this is all about promoting standards, it would be awkward if we ended up with `.rolluprc` files and `"rollup": {...}` in package.json files. That's a huge burden on the ecosystem that a change-weary community would be unlikely to accept (we've already seen that with `jsnext:main` to some degree).

Browserify draws a [distinction](https://github.com/substack/node-browserify/issues/501) between local and global transforms (local transforms being those specified in `pkg.browserify`). For Rollup, some transforms are naturally local (e.g. transpiling your library's TypeScript, or applying [environment variables](https://github.com/rollup/rollup/issues/208)), some are naturally global (the npm plugin for resolving external modules, for example). Some straddle the divide awkwardly, such as the CommonJS or Babel plugins.

Victorystick:
An issue I've encountered as I try to transition more towards ES modules is with testing. It seems to be related to global vs. local transforms. It is a problem without any obvious solution.

If I try to run mocha with Babel as a transpiler, it fails to process any of my (ES6) dependencies. (Perhaps its just a misconfiguration.)

So I started looking into if `mocha` could be integrated as a Gobble plugin using `.observe`. It required some hacks, and Mocha only executes the tests when Gobble is first launched, not after any subsequent changes.

> this is all about promoting standards

Indeed. I think we should look into making it easier to integrate Rollup into people's testing workflows, without resorting to any non-standard, Rollup specific config files.

It would be convenient to use Gobble for running Rollup's test suite after each code change.

Rich-Harris:
Yes, that's definitely something I want to look into. I haven't given Gobble anything like the attention it deserves recently (mostly because of time spent on Rollup!) but I intend to...

> Indeed. I think we should look into making it easier to integrate Rollup into people's testing workflows, without resorting to any non-standard, Rollup specific config files.

I think I might have been overthinking this earlier. It's totally reasonable for a developer to expect their dependencies to be ready-to-use, without having to think about transpiling someone else's TypeScript (or, indeed, ES6/7) – so there's really no need for an equivalent to `pkg.browserify`, which, fundamentally, exists because Browserify is bundling other people's source code rather than other people's distributable code.

Having said that, I, as the author of `foo` which depends on `bar`, might want to convert `bar` from CommonJS, and it's reasonable for me to be able to state that somewhere in `foo`'s build process. So a convention like `commonjs({ include: [ 'node_modules/foo/**' ] })` makes sense. In other words, put power in developers' hands to hack around whatever problems they might have, and don't be opaque and magical about where and when plugins get applied.

I think for me the confusion arose because of `jsnext:main`, which I thought of as being intended to point to external dependencies' source code, when in reality a package shouldn't ever have to think about its dependencies' source code. **For my part, I'm going to start using `jsnext:main` differently in my own packages:**

```js
{
  "main": "dist/foo.umd.js",

  // instead of e.g. "src/foo.js" - has already been through Babel etc
  "jsnext:main": "dist/foo.es6.js"
}
```

I do wonder if we need a less ambiguous name than `jsnext:main`, which is a little unclear about whether it includes non-module language features.

Issue #186, opened 2015-10-13
Warn about potential minification issues?


Rich-Harris:
An interesting case with UglifyJS – if you use `eval` or `with`, the containing scope's variable names can't be mangled. Since Rollup puts every module in a single scope, it's much more likely that mangling will fail for the top level of an entire bundle, meaning that a minified Rollup bundle ends up *larger* than the equivalent Browserify bundle, despite there being less code. (Gzipping goes some way to alleviating this, but not all the way.)

Do we try and catch and deal with those cases somehow? E.g. warn, or wrap everything (except declarations) in an IIFE if a module contains a top-level `eval` or `with`? My thinking is that most people wouldn't investigate too deeply before concluding that Rollup was just less efficient than Browserify or Webpack.

Via https://github.com/mishoo/UglifyJS2/issues/830#issuecomment-147870241

kzc:
It would be pretty straightforward to add such a warning in `uglifyjs` when `eval()` or `with` is encountered when mangling is enabled.

I think the Rollup project should be made aware of this issue as it could be fairly common. Perhaps they can perform some code transforms on their end to avoid this situation in some cases. i.e., don't hoist a function from a module in certain cases.

kzc:
Sorry - I mistakenly thought this was a post to the UglifyJS2 project.

Rich-Harris:
@kzc not at all! thank you for clarifying the issue.

I actually misunderstood the problem before – wrapping the `eval`-containing module in an IIFE wouldn't help, because the `eval` 'infects' all the parent scopes – instead, we'd need to wrap all the *other* modules to make them minifiable. Which makes that approach less appealing.

So, warning the user is probably a better solution.

PR #209, opened 2015-10-22
0.20.0

Rich-Harris:
This is incomplete and I can't do any more on it right now, I'm just pushing up what I've got in case anyone has feedback.

Changes (some fairly major breaking ones):

* The `transform`, `load`, `resolveId` and `resolveExternal` options are no longer supported – they should be dealt with by plugins. That means that to include third party modules with `jsnext:main` in `node_modules` (which the default resolver in 0.19 does for you) you should use [rollup-plugin-npm](https://github.com/rollup/rollup-plugin-npm)
* `options.external` is still supported, but if an external module *isn't* in that list it's a warning, not a fatal error
* Warnings are printed with `console.error` by default – you can override that with `options.onwarn` if you want. (This would also be useful for #186)
* Probably some other things I've forgotten about

* [ ] update documentation!


Rich-Harris:
Right, I think everything is ready to go. Given the magnitude of the breaking changes, I'll give everyone (cc @Victorystick!) a chance to weigh in before merging. This PR includes #207.

The one thing this doesn't yet do is allow you to use plugins via the CLI. CLI users who need Rollup to find their third party `jsnext:main` dependencies should stick with 0.19 for the time being.

Victorystick:
My initial reaction is; great! I'll have to wait until tomorrow to get a proper look though. :smile:

Victorystick:
This looks good! I'm a bit surprised how easy the transition to Plugins appears to be, both in terms of amount of code changed as well as changes required by users. :smile:

Regarding loading plugins from the CLI, I'd love to see something similar to what Gobble does. I can't help but to love what Gobble does; load Plugins relative the project and automatically install missing plugins. You have experience of both obviously. It'd be really neat to include something like that.

Another point regarding Rollup (the library) is whether it should include a default `load` function (that uses "fs") since it only makes sense in a Node context. Perhaps it should be included as a "default Plugin" in the non-browser builds.

And lastly a question; do you think it makes sense to create a `rollup-cli` package that is separate from the main `rollup` one? The CLI needs the Plugin-loading code, while the library doesn't. Opinion?

Rich-Harris:
> Regarding loading plugins from the CLI, I'd love to see something similar to what Gobble does

I think that's a good idea, yes. The CLI is a bit of a second-class citizen at the moment because of its lack of plugin support. Have been wondering about the best way to pass options to plugins – is it okay to do this sort of thing?

```bash
rollup --input src/main.js\
       --output dist/bundle.js\
       --format umd\
       --name myBundle\
       --sourcemap\
       --plugin babel {"sourceMap": true }\
       --plugin commonjs {"include": "node_modules/**" }
```

> Another point regarding Rollup (the library) is whether it should include a default load function (that uses "fs") since it only makes sense in a Node context. Perhaps it should be included as a "default Plugin" in the non-browser builds.

At the moment it sort of *is* a default plugin. We could completely remove all trace of it from the browser build, but given that 99% of the time it's used in a Node.js context I think we should prioritise making that as easy as possible – I definitely *don't* think users should have to explicitly choose to use the fs loader.

> And lastly a question; do you think it makes sense to create a `rollup-cli` package that is separate from the main `rollup` one?

Possibly – if we add features like incremental rebuilds (`rollup --watch` or something) I think it would make sense. There *is* a cost to doing so though – a package has to include both `rollup` and `rollup-cli` in its devDependencies if it uses Rollup in its build script, which adds a slight cognitive burden and increases the chances of e.g. mismatched versions. So I don't think we should try and squeeze it into this release, we should do it at leisure.

I'm just going to add one last thing – an error if you try to use the deprecated options, to ease upgrading – then I reckon we should hit the green button.

Victorystick:
I completely agree about that the Node build should be prioritized and that it should always include the `fs` loader by default. Having to manually import "rollup-plugin-fs-loader" isn't something anyone wants.

The browser build obviously _isn't as important_ as Node, certainly not at this time, but I think we could take advantage of the Rollup project sources and build processes to show how to present the "only include what you need" philosophy.

After all, that's one of its key selling points, at least for me.

TrySound:
@Rich-Harris Why sourcemaps for node version? It's just half mb of bad gzipping and useless stuff. Also browser version could be in separate project IMHO, caz the only usefull thing with it is playground.

Rich-Harris:
@Victorystick yep, that makes sense.

@TrySound The sourcemaps are very helpful for debugging – if you install `source-map-support` (which we [do in the tests](https://github.com/rollup/rollup/blob/master/test/test.js#L1)), any time there's an error, it gets traced back to the source file. It also enables more accurate [code coverage reports](https://codecov.io/github/rollup/rollup) (though the reporting is currently broken, which reminds me that we need to fix that). If you're not using them, the only cost is a few extra kb in the tar file that npm downloads.

Having a separate project for the browser build would introduce a *lot* of complexity, and for what gain? Again, if you're not using it, it doesn't really matter that it's there.

TrySound:
@Rich-Harris Okay sourcemaps are useful. But why sourcemaps in npm rollup module? I don't need to debug rollup afte installing. And only node sourcemaps removing saved 100kb.

ben-eb:
Agreed. Seems wasteful to publish sourcemaps to npm when they are used only for development purposes.

Rich-Harris:
> I don't need to debug rollup after installing

You might! No software is bug-free. If you encountered an error that was difficult to reproduce, I could ask you to install `source-map-support` and send us a stack trace, and it would make it a good deal more likely that we'd be able to identify the problem and resolve it.

I still don't understand what the value is of *not* having it? It makes literally no difference, unless when you `npm install rollup` you're downloading that extra 100kb over rubbish hotel wi-fi, in which case good luck using npm at all. What am I missing here?

PR #228, opened 2015-10-26
Disallow 'default' as an identifier name in `addReference`

### LINKED ISSUE #215: Rollup output sometimes illegally uses "default" as a var name
thanks to your nifty new site I can just link you straight to the bug, LIVE:

- [INVALID ROLLUP OUTPUT](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22es6%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22%2F%2F%20export%5Cnexport%20%7Bdefault%20as%20Foo%7D%20from%20'.%2Ffoo'%3B%5Cn%5Cnexport%20default%20function%20()%20%7B%5Cn%20%20%2F%2F%20whatever%5Cn%7D%3B%22%7D%2C%7B%22name%22%3A%22foo.js%22%2C%22code%22%3A%22export%20default%20class%20Foo%20%7B%5Cn%5Cn%7D%22%7D%5D%7D)
- [CONFIRMATION THAT IT'S INVALID](https://babeljs.io/repl/#?experimental=false&evaluate=false&loose=false&spec=false&code=class%20Foo%20%7B%0A%0A%7D%0A%0Afunction%20default%20()%20%7B%0A%20%20%2F%2F%20whatever%0A%7D%3B%0A%0Aexport%20%7B%20Foo%20%7D%3Bexport%20default%20default%3B)

Summary: if you `export {default as something} from './somewhere'` **and** try to `export default anotherThing` in the same file, the rollup output will contain the word `default` as a variable or function name, which is a syntax error.

Lovely site by the way

Victorystick:
Fixes #215

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -25,7 +25,7 @@ class SyntheticDefaultDeclaration {
 
 	addReference ( reference ) {
 		// Don't change the name to `default`; it's not a valid identifier name.
-		if ( reference.name !== 'default' ) return;
+		if ( reference.name === 'default' ) return;
 
 		reference.declaration = this;
 		this.name = reference.name;
--- a/src/Module.js
+++ b/src/Module.js
@@ -24,6 +24,9 @@ class SyntheticDefaultDeclaration {
 	}
 
 	addReference ( reference ) {
+		// Don't change the name to `default`; it's not a valid identifier name.
+		if ( reference.name !== 'default' ) return;
+
 		reference.declaration = this;
 		this.name = reference.name;
 	}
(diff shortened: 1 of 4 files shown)


src/Module.js
```diff
@@ -24,6 +24,9 @@ class SyntheticDefaultDeclaration {
 	}
 
 	addReference ( reference ) {
+		// Don't change the name to `default`; it's not a valid identifier name.
+		if ( reference.name === 'default' ) return;
```
Rich-Harris: shouldn't this come after `reference.declaration = this`? i.e.

```js
reference.declaration = this;
if ( reference.name !== 'default' ) this.name = reference.name;
```

Or do these types of reference never need to know about their declaration?

Victorystick: I'm not entirely sure to be honest. :confused:

PR #229, opened 2015-10-26
[RFC] add --config option to CLI (#226)

Rich-Harris:
This adds support for a `--config` option in the CLI, which, by extension, enables support for plugins (#226).

`rollup -c` or `rollup --config` will load options from a `rollup.config.js` config file in the current directory. You can override this file with `rollup -c some.other.file.js`.

Any additional arguments will override those specified in the config file (i.e. `rollup -c -f umd` will create a UMD build regardless of the config file.)

The config file is processed with Rollup before being required, so it can look like this...

```js
import babel from 'rollup-plugin-babel';

export default {
  entry: 'src/main.js',
  plugins: [ babel() ],
  format: 'umd'
};
```

...or this:

```js
var babel = require( 'rollup-plugin-babel' );

module.exports = {
  entry: 'src/main.js',
  plugins: [ babel() ],
  format: 'umd'
};
```

Thoughts welcome.

### DIFF
--- a/bin/help.md
+++ b/bin/help.md
@@ -7,6 +7,8 @@ Basic options:
 
 -v, --version            Show version number
 -h, --help               Show this help message
+-c, --config             Use this config file (if argument is used but value
+                           is unspecified, defaults to rollup.config.js)
 -i, --input              Input (alternative to <entry file>)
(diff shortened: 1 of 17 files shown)


Victorystick:
It's a bit late, but... :+1:

PR #231, opened 2015-10-27
0.20.2

Rich-Harris:
Couple of small fixes, plus we're now able to use Rollup CLI to build Rollup


rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+	.replace( '${version}', pkg.version )
+	.replace( '${time}', new Date() )
+	.replace( '${commitHash}', commitHash );
+
+export default {
+	entry: 'src/rollup.js',
+	format: 'cjs',
+	plugins: [
+		babel({
+			//include: [ 'src/**', 'node_modules/acorn/**' ]
```
Victorystick: Is this comment really supposed to be here?

Rich-Harris: good catch - fixed

PR #236, opened 2015-10-28
Remove unneeded `.js` extension in import.

eventualbuddha:


TrySound:
@eventualbuddha Squash?

eventualbuddha:
@TrySound sure. That's what happens when you try to do tweaks from the web UI. :stuck_out_tongue_winking_eye:

eventualbuddha:
@TrySound squashed.

TrySound:
@eventualbuddha Nice!)

Victorystick:
I'm not sure whether we want to _remove_ the `.js` extensions where they are used, or if we want to _add them_ where they're not.

TrySound:
@Victorystick WIthout them code looks much cleaner.

Rich-Harris:
There was a conversation a while back (can't remember where, sorry) where the conclusion was that `import` statements *should* include file extensions since it's less ambiguous and less likely to cause any issues with loader implementations (whenever they happen). I quite like the fact that `./foo.js` looks like a file but `foo` looks like a *package*, as well.

So my vote would be to add extensions everywhere – but in the meantime I vote in favour of increased consistency, so <merge>

eventualbuddha:
It's definitely a matter of preference, but to be honest I haven't come across any projects I can remember that intentionally include the optional extension. Perhaps I'm just too used to the node convention of omitting it.

Rich-Harris:
One thing that's highly problematic about the node convention: sometimes `./foo` means `foo.js`, sometimes it means `./foo/index.js`. That's okay when you're bundling (or loading synchronously, via `require`, even if it means wasting time trying out different paths), but obviously no good in browsers – another reason to avoid the ambiguity. (I seem to remember reading one of the OG node contributors lamenting extension-less requires as a bad mistake that had created a huge amount of unnecessary complexity that was simply too costly to fix.)

eventualbuddha:
Well, my primary interest is in consistency, so if adding extensions is the way to go then :+1:.

Rich-Harris:
Cool – have opened #237

PR #244, opened 2015-10-30
Export all bugs

Rich-Harris:
Found a couple of bugs while trying to use Rollup on Babel 6 (because I'm having some startup performance issues with it, which I'm fairly sure are related to the level of modularisation combined with the fact that [require is dog slow](https://kev.inburke.com/kevin/node-require-is-dog-slow/) – other people are [experiencing similar issues](https://news.ycombinator.com/item?id=10474175), though it looks as though npm 3 may help a lot).

1. you can't have multiple `export *` declarations in a module
2. missing namespace exports are treated as an error, when they should really be treated as a warning, I think

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -81,10 +81,9 @@ class SyntheticNamespaceDeclaration {
 
 			// throw with an informative error message if the reference doesn't exist.
 			if ( !original ) {
-				const err = new Error( `Export '${reference.name}' is not defined by '${this.module.id}'` );
-				err.code = 'MISSING_EXPORT';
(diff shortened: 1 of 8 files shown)


Victorystick:
Interesting. What was the cause of the export all bug?

Rich-Harris:
If you had `export * from './foo.js; export * from './bar.js`, it would try to trace exports from `bar` in `foo`, and throw an error when it failed. This change moves the job of throwing on missing imports from the exporter to the importer, which is more logical anyway

PR #249, opened 2015-10-31
exclude dead branches

Rich-Harris:
Follow-up to https://github.com/rollup/rollup/issues/208#issuecomment-150096281. This PR excludes dead branches, wherever it's possible to determine that through static analysis.

This goes slightly further than what e.g. UglifyJS does:

```js
var obj = {};
obj.foo = function () {
  console.log( 'this function should be excluded' );
};

if ( false ) obj.foo();
```

Rollup will exclude `obj.foo` but UglifyJS won't, as far as I can tell (there may be an option that I don't know about).

Alongside https://github.com/rollup/rollup-plugin-replace, this should make it easier to make different bundles for different targets (e.g. browser vs node, dev vs prod).


src/ast/conditions.js
```diff
@@ -0,0 +1,65 @@
...
+	if ( a.type === 'Literal' ) return a.value === b.value;
+	if ( a.type === 'Identifier' ) return a.name === b.name;
+
+	return false;
+}
+
+function nodesAreNotEqual ( a, b ) {
+	if ( a.type !== b.type ) return false;
+	if ( a.type === 'Literal' ) return a.value != b.value;
+	if ( a.type === 'Identifier' ) return a.name != b.name;
```
Victorystick: Any reason to use `!=` here while you use `===` above?

src/ast/conditions.js
```diff
@@ -0,0 +1,65 @@
...
+	if ( node.type === 'LogicalExpression' ) {
+		if ( node.operator === '&&' ) return isTruthy( node.left ) && isTruthy( node.right );
+		if ( node.operator === '||' ) return isTruthy( node.left ) || isTruthy( node.right );
+		return false;
+	}
+
+	return false;
+}
+
+export function isFalsy ( node ) {
```
Victorystick: It seems odd to duplicate all this code.

How about something like:

```js
function _isTruthy ( node ) {
	...

	// `null` signifies undeterminable
	return null;
}

export function isTruthy ( node ) {
	// replace `null` with `false`
	return _isTruthy( node ) || false;
}

export function isFalsy ( node ) {
	const truthy = _isTruthy( node );

	// replace `null` with `false`, and invert the truthy value.
	return truthy === null ? false : !truthy;
}
```

src/ast/conditions.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function isNotEqualTest ( node ) {
+	return node.type === 'BinaryExpression' && ( node.operator === '!==' || node.operator === '!=' );
+}
+
+function nodesAreEqual ( a, b ) {
+	if ( a.type !== b.type ) return false;
+	if ( a.type === 'Literal' ) return a.value === b.value;
+	if ( a.type === 'Identifier' ) return a.name === b.name;
```
Victorystick: This will fail for
```js
var a = NaN;

if ( a === a ) ...
```

Rich-Harris: good catch. fixed locally

src/ast/conditions.js
Rich-Harris: yeah, it's to do with strict vs non-strict equality. though now that you mention it, i think it'd be better to have separate `nodesAreStrictEqual` and `nodesAreNotStrictEqual` functions

src/ast/conditions.js
Rich-Harris: :+1: that makes a lot more sense

src/ast/conditions.js
Victorystick: No, I mean; what if an _identifier_ has the _value_ `NaN`?

Victorystick:
What I mean is, what if an _identifier_ has the _value_ `NaN`?

The assumption `a == a -> true` is invalid.

Rich-Harris:
> What I mean is, what if an identifier has the value NaN?

Ah. Good point. Bloody NaN! Is there any more useless feature in any programming language?

So do we skip identifier equality checks altogether? I guess the only alternative is to have an option controlling whether or not to care about NaN, which is territory I'd rather not get into

Victorystick:
I fear the best thing to do is to _skip-identifiers_. It's unlikely to end up with `NaN`s like that, but it could happen. Unless we can be sure of the identifiers' values.

Say our `Declaration`s had a `value` property, then we could compare actual values. `const` variables would be easy. It'd be a best-effort thing regardless.

Rich-Harris:
agreed

Victorystick:
@Rich-Harris Mind if I go ahead and merge this?

Rich-Harris:
Yep let's do it. We should probably merge #253 too and release it as 0.21 (so that people can pin 0.20 if it turns out the algorithm fails in certain scenarios)

Issue #45, opened 2015-07-16
Bundling pre-ES6 modules


callumlocke:
Hi Rich. This project looks very exciting. I'm sold.

I have a question about pre-ES6 modules... I tried importing lodash as an experiment (today's lodash), and got this error:

```
Package lodash does not have a jsnext:main field, and so cannot be included in your rollup. Try adding it as an external module instead (e.g. options.external = ['lodash']
```

Is this a permanent design decision or is it just early days? Will you ever add functionality that can read an old-style `main` script and traverse `require()` calls...or do you recommend another strategy for people who want to mix ES6 and old school modules in one bundle (e.g. using Browserify as another build step after rollup)?

Rich-Harris:
I've gone back and forth on this. For the time being I think it's better to focus solely on ES6 modules, at least until we iron out all the kinks, but long term we definitely need a sane way to handle legacy module formats.

In theory, it can be done by overriding the `resolveId` function to accept CommonJS modules, and adding a transform step that wraps or rewrites modules so that Rollup can analyse them. To what extent you can convert CommonJS etc, I don't know - in a lot of cases it'd be straightforward, but I've no idea how you'd handle things like this (it's not quite as contrived as it looks)...

```js
var myLib = exports;
myLib.foo = 'foo';

augment(myLib);
function augment(lib) { lib.bar = 'bar'; }
```

...so a strategy of wrapping rather than rewriting is probably better.

Luckily @guybedford has spent a lot of time thinking about these problems, and has done some initial research on incorporating Rollup's strategy inside JSPM: https://github.com/systemjs/builder/issues/199. This seems like the best route forward, since JSPM is a soup-to-nuts solution (repository, bundler, module loader polyfill etc all talking the same language) that already handles legacy formats.

For right now though, my workaround is to have a two-phase bundling process, exactly as you suggest. First, I bundle my app code (plus any npm depedencies that expose `jsnext:main`, i.e. anything I wrote myself :laughing:) as CommonJS, and then I pass that bundle over to Browserify (but could be Webpack or whatever), which takes it from there. It works perfectly well, it's just a PITA to have to juggle all these different tools which is why I'm excited to see where JSPM goes.

Victorystick:
[...]

PR #253, opened 2015-11-01
More accurate side-effect detection

Rich-Harris:
This is quite a big PR; bear with me. Among other things it needs some battle testing and a tidy up before it gets merged (some of the names are a bit weird, etc).

Currently, any top-level CallExpression (i.e. not inside a function declaration/expression) is assumed to contain side-effects, and is therefore included in the bundle. This is often [wasteful](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22amd%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20%7B%20random%20%7D%20from%20'.%2Fmaths.js'%5Cn%5Cnexport%20const%20r%20%3D%20random(%200%2C%20100%20)%3B%5Cn%22%7D%2C%7B%22name%22%3A%22maths.js%22%2C%22code%22%3A%22export%20function%20power%20(%20n%20)%20%7B%5Cn%20%20return%20function%20(%20base%20)%20%7B%5Cn%20%20%20%20return%20Math.pow(%20base%2C%20n%20)%3B%5Cn%20%20%7D%3B%5Cn%7D%5Cn%5Cnexport%20const%20square%20%3D%20power(%202%20)%3B%5Cnexport%20const%20cube%20%3D%20power(%203%20)%3B%5Cn%5Cnexport%20function%20random%20(%20min%2C%20max%20)%20%7B%5Cn%20%20return%20min%20%2B%20(%20max%20-%20min%20)%20*%20Math.random()%3B%5Cn%7D%22%7D%5D%7D):

```js
// main.js
import { random } from './maths.js'

export const r = random( 0, 100 );

// maths.js
export function power ( n ) {
  return function ( base ) {
    return Math.pow( base, n );
  };
}

export const square = power( 2 );
export const cube = power( 3 );

export function random ( min, max ) {
  return min + ( max - min ) * Math.random();
}
```

In this example there's no reason for `square`, `cube` and `power` to be included. As of this PR, Rollup is smart enough to recognise that `power` refers to a pure function (i.e. calling it has no side-effects).

A few other changes that fall out of this way of doing things:

* Classes transpiled with Babel, which [currently get included](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22amd%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20%7B%20Foo%20%7D%20from%20'.%2Fclasses.js'%5Cn%5Cnexport%20default%20new%20Foo()%3B%5Cn%22%7D%2C%7B%22name%22%3A%22classes.js%22%2C%22code%22%3A%22%2F*%20the%20code%20below%20is%20what%20Babel%20produces%20from%20these%5Cn%20%20%20class%20declarations%5Cnexport%20class%20Foo%20%7B%5Cn%20%20method%20()%20%7B%7D%5Cn%7D%5Cn%5Cnexport%20class%20Bar%20%7B%5Cn%20%20method%20()%20%7B%7D%5Cn%7D%5Cn*%2F%5Cn%5Cnfunction%20_classCallCheck(instance%2C%20Constructor)%20%7B%20if%20(!(instance%20instanceof%20Constructor))%20%7B%20throw%20new%20TypeError(%5C%22Cannot%20call%20a%20class%20as%20a%20function%5C%22)%3B%20%7D%20%7D%5Cn%5Cnvar%20Foo%20%3D%20(function%20()%20%7B%5Cn%20%20function%20Foo()%20%7B%5Cn%20%20%20%20_classCallCheck(this%2C%20Foo)%3B%5Cn%20%20%7D%5Cn%5Cn%20%20Foo.prototype.method%20%3D%20function%20method()%20%7B%7D%3B%5Cn%5Cn%20%20return%20Foo%3B%5Cn%7D)()%3B%5Cn%5Cnvar%20Bar%20%3D%20(function%20()%20%7B%5Cn%20%20function%20Bar()%20%7B%5Cn%20%20%20%20_classCallCheck(this%2C%20Bar)%3B%5Cn%20%20%7D%5Cn%5Cn%20%20Bar.prototype.method%20%3D%20function%20method()%20%7B%7D%3B%5Cn%5Cn%20%20return%20Bar%3B%5Cn%7D)()%3B%5Cn%5Cnexport%20%7B%20Foo%2C%20Bar%20%7D%3B%22%7D%5D%7D) whether or not they're used because they get wrapped in an IIFE, are now only included if appropriate
* We no longer need to treat IIFEs differently – they're just the same as any other CallExpression.
* Detecting strong dependencies is slightly more robust – if a module calls `foo`, and `foo` refers to `bar`, the module depends strongly on the module where `bar` is defined. Previously, the module had to have a `bar` reference at the top level (or inside an IIFE, thanks to some special case logic introduced to accommodate a pattern used a lot inside Three.js)
* Global functions that are known to be pure (e.g. `new Array(foo)` or `Object.keys(bar)`) can be treated the same as local functions that are known to be pure

Unfortunately this only goes some way towards solving the problem identified [here](https://github.com/rollup/rollup/issues/45#issuecomment-151160765) – that apps using individual Lodash functions end up bundling a lot of unnecessary stuff (though with this PR it's nowhere near as bad as @callumlocke found). A lot of it could be eliminated if the lodash-es build were to forgo a default export altogether, but that's certainly not a perfect solution.

The reason the extra stuff gets included is because Rollup digs up the entire dependency graph and scans it for side-effects – and discovers statements that *could* have side-effects, but in modules we're not really interested in.

To be completely safe, we do need to include those side-effects because of [things like this](https://github.com/rollup/rollup/blob/d4eeb54dae518c60c90b07524e685d130e85ba2a/test/function/top-level-side-effects-are-preserved/rsvp.js#L5), where module `a` happens to muck about with the contents of module `b`, which we're using. But in a lot of cases – Lodash being a fine example – that's overly cautious. So this PR also introduces `aggressive` mode, which only looks for side-effects in the entry module:

```js
rollup.rollup({
  entry: 'main.js',
  aggressive: true
}).then(...)
```

With `aggressive: true`, these result in almost exactly the same sized bundle:

```js
// this...
import {chunk, zipObject} from 'lodash-es';

// ...or this
import chunk from 'lodash-es/array/chunk';
import zipObject from 'lodash-es/array/zipObject';

console.log(zipObject(chunk(['a', 'b', 'c', 'd'], 2)));
```

Oh, and I moved a couple of files around to make things a bit easier.

Thoughts welcome!


Victorystick:
Woo! Comments! I haven't had a chance to read through the PR on a computer, but I like the look of it.

eventualbuddha:
I don't really have the time to look at this properly, but I looked over the tests quickly and they seem sane. This is exciting!

Rich-Harris:
I think I have some changes on this branch that aren't checked in (on another machine) so I won't merge this just yet

PR #263, opened 2015-11-03
support async transformers

### LINKED ISSUE #260: Async plugins
Is there a way to use asynchronous in plugins? rollup could detect promise returned from plugin like
```js
export default function string (opts) {
	return {
		transform (code, id) {
			if (filter(id)) {
				if (opts.someAsyncOpt) {
					return new Promise(function (code) {
						resolve(code);
					});
				}
				return code;
			}
		}
	};
}
```

Rich-Harris:
closes #260


TrySound:
Cool, thanks!

Victorystick:
Looks good to me!

PR #268, opened 2015-11-04
Fix assignment to re-exported bindings.

### LINKED ISSUE #267: Exported bindings are unexpectedly renamed.
Related #11, where you wrote:

> I've changed some of the rewriting behaviour so that exported vars (or lets) are always rewritten as `export.foo = ...`. Involves a little bit of dancing around, but I think it's the best solution. Hopefully I didn't introduce any bugs in the process.

Here’s my entry file I’m trying to rollup:

```js
export {
  event,
  select,
  selectAll
} from "d3-selection";
```

The resulting code looks like this:

```js
// initialization of event is incorrectly renamed “event$1”
exports.event$1 = null;

// likewise with assignment
function example(foo) {
  exports.event$1 = foo;
}

// then the initial value is copied into the exports object under the correct name
exports.event = exports.event$1;
```

So, two problems:

* There’s an undesired export “event$1” which has the correct value.
* The desired export “event” is always null.

mbostock:
Fixes #267.

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -344,7 +344,7 @@ export default class Module {
 
 				if ( declaration ) {
 					declaration.addReference( reference );
-				} else {
+				} else if ( statement.node.type !== 'ExportNamedDeclaration' || !this.reexports[ reference.name ] ) {
 					// TODO handle globals
 					this.bundle.assumedGlobals[ reference.name ] = true;
 				}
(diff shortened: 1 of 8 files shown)


mbostock:
This didn’t quite fix my bug, because it appears that there’s another bug related to missing declarations. Specifically, changing from this:

```js
var window = defaultView(node),
    event = window.CustomEvent;
```

To this:

```js
var window = defaultView(node);
var event = window.CustomEvent;
```

Fixed another assumed `event` global.

mbostock:
Okay, fixed again. :grin:

Rich-Harris:
Thank you!

PR #270, opened 2015-11-04
Add a test that should fail, but doesn’t.

mbostock:
I think I may have broken something in my previous fix?

### DIFF
--- a/test/function/export-from-no-local-binding-var/_config.js
+++ b/test/function/export-from-no-local-binding-var/_config.js
@@ -1,5 +1,3 @@
-var assert = require( 'assert' );
-
 module.exports = {
 	description: 'export from does not create a local binding'
 };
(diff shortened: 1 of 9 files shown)


test/function/export-from-no-local-binding-var/_config.js
```diff
@@ -0,0 +1,5 @@
+var assert = require( 'assert' );
```
eventualbuddha: Unnecessary `assert` import?

mbostock:
Okay, I figured out the fix, and yes, I did introduce a bug. :frowning:

mbostock:
Okay, I removed the extraneous requires.

Rich-Harris:
> Okay, I figured out the fix, and yes, I did introduce a bug.

I think it's more a case of exposing a pre-existing bug! I was taking a look at the same time and came upon an alternative fix, which is to disregard re-exports altogether in the initial analysis – will add a PR in a sec

mbostock:
Nice, thanks for investigating.

Rich-Harris:
Thanks for the test! Released 0.20.5 with the fix

Issue #279, opened 2015-11-08
SourceMap test failure on Windows


nicolo-ribaudo:
I just cloned this repository and `npm install`ed (which automatically runs mocha)

These tests fail:
```
      sourcemaps
        1) generates amd
        2) generates cjs
        3) generates es6
        4) generates iife
        5) generates umd
      sourcemaps-inline
        6) generates amd
        7) generates cjs
        8) generates es6
        9) generates iife
        10) generates umd
```

Here is an example:
```
  1) rollup form sourcemaps generates amd:

      AssertionError: { version: 3,
  file: 'amd.js',
  sources: [ '../foo.js', '../bar.js', '../main.js' ],
  sourcesContent:
   [ 'export default f deepEqual { version: 3,
  file: 'amd.js',
  sources: [ '../foo.js', '../bar.js', '../main.js' ],
  sourcesContent:
   [ 'export default f
      + expected - actual

       {
         "file": "amd.js"
      -  "mappings": ";;CAAe,SAAS,GAAG,IAAI;EAC9B,OAAO,CAAC,GAAG,EAAE,mBAAmB,EAAE,CAAC;;;CCDrB,SAAS,GAAG,IAAI;EAC9B,OAAO,CAAC,GAAG,EAAE,mBAAmB,EAAE,CAAC;;;CCEpC,OAAO,CAAC,GAAG,EAAE,oBAAoB,EAAE,CAAC;;CAEpC,GAAG,EAAE,CAAC;CACN,GAAG,EAAE,;;"
      +  "mappings": ";;CAAe,SAAS,GAAG,IAAI;EAC9B,OAAO,CAAC,GAAG,EAAE,mBAAmB,EAAE;;;CCDpB,SAAS,GAAG,IAAI;EAC9B,OAAO,CAAC,GAAG,EAAE,mBAAmB,EAAE;;;CCEnC,OAAO,CAAC,GAAG,EAAE,oBAAoB,EAAE;;CAEnC,GAAG,EAAE;CACL,GAAG,EAAE,;;"
         "names": []
         "sources": [
           "../foo.js"
           "../bar.js"

```

Victorystick:
@nicolo-ribaudo I just installed Node 5 but haven't been able to reproduce the test failures.

Rich-Harris:
I would guess (but haven't checked yet) that it's an npm 3 thing rather than a node 5 thing – trashing `node_modules` and reinstalling would probably reproduce the issue

[...]

Issue #258, opened 2015-11-02
minification


TrySound:
Is there a way to minify ast before generating? Maybe possible to use plugin for this?

Victorystick:
Rollup doesn't generate its bundles from an AST. It manipulates an underlying MagicString instance based on the original source that preserves information required to generate source maps. That's also why all comments are preserved exactly as the we're written. Rollup isn't really designed to enable minification during bundling. We recommend using UglifyJS or a similar tool as a postprocessing step.

TrySound:
@Victorystick I just would like to use `bundle.write` method without own fs.writeFile calling

Victorystick:
Can't you just use `{ code } = bundle.generate( ... )` and pass the resulting code to UglifyJS?

TrySound:
@Victorystick Yes I can, but also need to write this code on disk. In this case code becomes less elegant as when we use `bundle.write`.

TrySound:
[Here's my example](https://github.com/TrySound/case/blob/3.0/gulpfile.js#L33-L57)

Victorystick:
I see. Unfortunately, minification is out of scope for Rollup itself, and I doubt we'll add support for a minification plugin. /cc @Rich-Harris

TrySound:
@Victorystick Thanks

Rich-Harris:
Yeah, we'd have to add something like a `postprocess` hook for plugins. But I'm not sure that it makes that much sense, because you'd essentially need to create the bundle twice – once with the uglify plugin and once without (unless you only wanted the minified version). Probably more efficient to keep the two steps separate.

PR #282, opened 2015-11-10
Add bundle write transform option

TrySound:
It's realy necessary thing. Use case is minifying.

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -29,38 +29,42 @@ export function rollup ( options ) {
 			generate: options => bundle.render( options ),
 			write: options => {
 				if ( !options || !options.dest ) {
-					throw new Error( 'You must supply options.dest to bundle.write' );
+					return Promise.reject( new Error( 'You must supply options.dest to bundle.write' ) );
 				}
 
 				const dest = options.dest;
 				let result = bundle.render( options );
 
-				if (typeof options.transform === 'function') {
-					result = options.transform(result);
-					if (typeof result !== 'object' || result === null) {
+				return Promise.resolve( result ).then( result => {
+					if ( typeof options.transform === 'function' ) {
+						return options.transform( result );
+					}
+					return result;
+				}).then( result => {
+					if ( typeof result !== 'object' || result === null ) {
 						throw new Error( 'options.transform should return { code, map }' );
 					}
-				}
 
-				let { code, map } = result;
+					let { code, map } = result;
 
-				let promises = [];
+					let promises = [];
 
-				if ( options.sourceMap ) {
-					let url;
+					if ( options.sourceMap ) {
+						let url;
 
-					if ( options.sourceMap === 'inline' ) {
-						url = map.toUrl();
-					} else {
-						url = `${basename( dest )}.map`;
-						promises.push( writeFile( dest + '.map', map.toString() ) );
+						if ( options.sourceMap === 'inline' ) {
+							url = map.toUrl();
+						} else {
+							url = `${basename( dest )}.map`;
+							promises.push( writeFile( dest + '.map', map.toString() ) );
+						}
+
+						code += `\n//# ${SOURCEMAPPING_URL}=${url}`;
 					}
 
-					code += `\n//# ${SOURCEMAPPING_URL}=${url}`;
-				}
-
-				promises.push( writeFile( dest, code ) );
-				return Promise.all( promises );
+					promises.push( writeFile( dest, code ) );
+					return Promise.all( promises );
+				});
 			}
 		};
 	});
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -33,7 +33,16 @@ export function rollup ( options ) {
 				}
 
 				const dest = options.dest;
-				let { code, map } = bundle.render( options );
+				let result = bundle.render( options );
+
+				if (typeof options.transform === 'function') {
+					result = options.transform(result);
+					if (typeof result !== 'object' || result === null) {
+						throw new Error( 'options.transform should return { code, map }' );
+					}
+				}
+
+				let { code, map } = result;
 
 				let promises = [];
 
--- a/test/test.js
+++ b/test/test.js
@@ -68,13 +68,13 @@ describe( 'rollup', function () {
 					}
 				}]
 			}).then( function ( bundle ) {
-				assert.throws( function () {
-					bundle.write();
-				}, /must supply options\.dest/ );
+				bundle.write().catch( function ( err ) {
+					assert.equal( err.message, 'You must supply options.dest to bundle.write' );
+				});
 
-				assert.throws( function () {
-					bundle.write({});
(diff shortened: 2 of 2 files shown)


TrySound:
Also this tests fails on windows
```
form
  sourcemaps
    1) generates amd
    2) generates cjs
    3) generates es6
    4) generates iife
    5) generates umd
  sourcemaps-inline
    6) generates amd
    7) generates cjs
    8) generates es6
    9) generates iife
    10) generates umd
```

Victorystick:
@TrySound Aah! That's probably related to #279! I couldn't reproduce it on Mac.

TrySound:
@Victorystick Exactly. Okay, as windows user will try to fix that in free time.

Victorystick:
I'd recommend using a tool like gulp or [`gobble`](https://github.com/gobblejs/gobble) if you're looking for a tool/build system to rollup and then minify code.

```js
var gobble = require( 'gobble' );

module.exports =
  gobble('src')
  .transform('rollup', {
    ...
  })
  .transform( 'uglifyjs' );
```

TrySound:
@Victorystick I use gulp, but I don't want create wrapper until rollup has to much `sourceMap` options (rollup and plugins). And I don't want to create a plugin caz rollup use Promise API, which is also used by gulp. But I want to use rollup write method. Why too much duplicate code? Hook is a good stuff for postprocessing.

TrySound:
@Rich-Harris https://github.com/rollup/rollup/issues/258#issuecomment-153210709
Not for plugins, but still hook :)

TrySound:
@Rich-Harris Let's discuss. This is the only thing which don't let me use rollup.

Victorystick:
@TrySound Why don't you just use [`gulp-rollup`](https://github.com/mcasimir/gulp-rollup)?

TrySound:
@Victorystick Why use unnecessary plugin if we can use Rollup and Promises API? `bundle.write` is small build utility which should cover some features like post processing. Why you think it's bad? This case is the same as gulp-browserify. We don't need to use wrapper, but we need some stuff for it.

Victorystick:
If what you want is to do in-memory minification, the why not use the in-memory, streaming gulp build system? We should keep Rollup's core focused. Why are you so determined to use `bundle.write`?

TrySound:
@Victorystick I like promises more. I can just return promise in gulp task. It's really cool. But gulp-rollup is excess.

TrySound:
If you want to have only process api, you should remove bundle.write method, else need to make to more friendly.

TrySound:
I don't like gulp-rollup. I want to use this hook.

TrySound:
Example
```js
rollup().then(function (bundle) {
  return bundle.write({
    dest: 'path.js',
    sourceMap: true,
    hook: function (result) {
      return uglify.minify(result.code, {
        fromString: true,
        inSourceMap: result.map
        outSourceMap: 'file.js.map'
      });
    }
  });
});
```

Rich-Harris:
Thanks for the example @TrySound, I see what you're getting at now. Putting the hook here makes sense, since it means it's possible to generate a minified and non-minified version from the same `bundle` object without creating a bunch of unnecessary work for Rollup.

That said, it should be done in such a way that it could work as a plugin (e.g. if you didn't need to create a non-minified version, you could use a rollup-plugin-uglify or whatever). In that case the plugin would look something like

```js
export default myPlugin ( options = {} ) {
  return {
    hook ({ code, map }) {
      // some code happens
    }
};
```

`hook` is obviously too abstract in this context, and it'd be nice to use the same terminology everywhere. Maybe `transformBundle`?

TrySound:
@Rich-Harris Maybe in bundle.write option it should have only `transform` option. But for plugins I'm not sure this feature is safe. There could be many unexpected behaviours.

Rich-Harris:
I'm not sure why this would be any more prone to unexpected behaviour than any other hook. As long as the API is clear, then it's down to plugin authors to not write broken plugins, surely?

TrySound:
@Rich-Harris Okay, but plugins bundle transformation is another pr I think

TrySound:
@Rich-Harris Should I do something more for merge?

TrySound:
@Rich-Harris Maybe we should let to return promise?

TrySound:
@eventualbuddha Reproduced!

TrySound:
I will add tests tomorrow. Maybe we can add uglifyjs [here](https://github.com/webpack/docs/wiki/comparison)?

Rich-Harris:
:+1:, though I think this needs to be a plugin hook as well – that way it would also work with the CLI and with `bundle.generate` for no extra work, and it would make it possible to have a rollup-plugin-uglifyjs plugin. We can also do the delicate work of combining sourcemaps, if necessary, rather than making that the dev's responsibility. Hook would have to receive `dest`/`sourceMapFile` options at minimum, of course.

I'll open a separate issue for that.

Victorystick:
How about renaming `hook` to `postprocess` or similar? It makes the intent clearer.

Rich-Harris:
[At the moment](https://github.com/rollup/rollup/pull/282/files#diff-1d7a04be4b67a51bd7ca47ce6ccb680fR39) it's `transform` – do we prefer that or `postprocess`? `transform` is consistent with module transforms, though there is a difference insofar as module transforms happen *before* Rollup gets a look at them, whereas this happens afterwards. So I'm indifferent.

Victorystick:
You suggested making it a plugin hook, so we've got to name it something other than `transform` since that's already used. That's why I suggested `postprocess`; an uglify plugin would just define a `postprocess` method. Unless you'd prefer `transformModule`/`transformBundle`?

TrySound:
@Victorystick I more like `transformBundle`

eventualbuddha:
:+1: `transformBundle`

Rich-Harris:
Cool. `transformBundle` it is.

@TrySound as the resident Windows guy, any idea what's behind [this](https://ci.appveyor.com/project/Rich-Harris/rollup/build/435/job/3sxp1l7pygflik5v)?

screen shot 2015-12-29 at 1 34 41 pm

Have seen it before but can't imagine why it's happening...

TrySound:
@Rich-Harris You mean this pr fail? It was fixed with my pluginutils pr. There was problem with path resolving in `rollup-plugin-babel`. Nothing was transpilled. Now everything should be okay.

Rich-Harris:
Ah ok. I suspect we need to update rollup-plugin-babel to pick up the newer version of rollup-pluginutils in that case. Which means adopting Babel 6. Which I think means ditching `class` in favour of good old-fashioned prototypes, since using loose mode is a bit of a PITA in Babel 6, and non-loose classes are fat and slow. Will open a separate issue for that

Rich-Harris:
@TrySound did you mean to close this?

TrySound:
@Rich-Harris Will make new pr.

PR #288, opened 2015-11-11
Log errors when using `rollup -c`

Victorystick:
When using `rollup -c` with an invalid rollup.config.js file, no warnings are logged. It should.


Rich-Harris:
:+1:

PR #295, opened 2015-11-12
Add error message when reexporting missing identifier

### LINKED ISSUE #274: export { nonexistent } from './other.js' in entry module fails
It *should* fail, but with a more informative error message than 'Cannot set property 'isExported' of undefined'

Victorystick:
Fix #274


Rich-Harris:
:+1:

Issue #305, opened 2015-11-14
Side-effect detection is imprecise.


mbostock:
[Original title: Imports of re-exports may pull in unused code.]

Given the [index.js](https://github.com/d3/d3-scale/blob/master/index.js) in [d3-scale](https://github.com/d3/d3-scale), I expect the following two entry modules to be equivalent. Here’s [indirect.js](https://gist.github.com/mbostock/61e37a526c99b0aa3135#file-indirect-js):

``` js
import {category10} from "d3-scale";

export default {
  scale: {
    category10: category10
  }
};
```

And [direct.js](https://gist.github.com/mbostock/61e37a526c99b0aa3135#file-direct-js):

``` js
import category10 from "./node_modules/d3-scale/src/category10";

export default {
  scale: {
    category10: category10
  }
};
```

However, the resulting bundles are very different: direct.bundle.js is 7,406 bytes and indirect.bundle.js is 70,126 bytes (9.5× as big)!

The full code is at https://gist.github.com/mbostock/61e37a526c99b0aa3135.

I think this issue might be a duplicate of #179, so feel free to close. My guess is even though those re-exports are unused, their initialization could have side-effects and so Rollup is including them to be safe. For example [src/time.js](https://github.com/d3/d3-scale/blob/master/src/time.js) in d3-scale initializes some time formats that will be used by time scales. Those formats are never used because time scales are not included in the bundle, but their initialization could have side-effects.

Perhaps this suggests that I should avoid greedy initialization of static variables, and instead somehow defer them until they are needed (such as upon the first instantiation of a time scale). Alternatively, some flag lets me tell Rollup to do the right (if unsafe) thing automatically would be cool. :)

Rich-Harris:
I just published 0.21.0, which includes #253. It does a *slightly* better job of ignoring irrelevant side-effects – 52,885 bytes versus 70,126 – but still nowhere near as good as the direct approach.

0.21 does have an `aggressive: true` option that ignores side-effects that don't contribute to the bundle's exports, but it turns out it's a little *too* aggressive – the result is 4,968 bytes because it's missing some essential code! Going to look into it and figure out if that's expected behaviour.

I don't yet know whether the answer will be more sophisticated analysis or more granular control. In the meantime importing directly from a file probably isn't the worst thing in the world in a situation like this. With the npm plugin you could omit the `./node_modules/` prefix.

mbostock:
Thanks for your hard work. Glad we’re making progress.

[...]

PR #307, opened 2015-11-14
Add failing aggressive test case.

mbostock:
Related #305. Pretty simple, maybe? Seems to be missing any property assignments to functions.


Rich-Harris:
Thanks. I spent a bit of time poking at this today and it turns out you basically lose all the benefits of aggressive mode when you make it behave in situations like this (the d3-scale example goes from 52K to 51K with 'less aggressive' mode). I've raised #309 which proposes doing away with it altogether – I don't think it justifies the added complexity.

PR #309, opened 2015-11-14
Remove aggressive mode

Rich-Harris:
This was an interesting experiment, but I don't think it's viable. It's easy enough to change the behaviour of aggressive mode to pass #307, but not without pulling in a lot of code that we probably want to exclude.

The difference in bundle size between normal mode and aggressive-but-not-downright-negligent mode is too small to justify the increased complexity (in terms of both surface area for bugs, and the API). So I think we should just remove it altogether, and continue to investigate ways to identify chunks of code that are safe to remove.


eventualbuddha:
:+1:

Issue #310, opened 2015-11-15
createFilter incorrect resolves transformations


TrySound:
```
const absolutePath = /^(?:\/|(?:[A-Za-z]:)?[\\|\/])/;
^^^^^
SyntaxError: Use of const in strict mode.
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (D:\_host\github\rollup\test\test.js:10:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at D:\_host\github\rollup\node_modules\mocha\lib\mocha.js:216:27
    at Array.forEach (native)
    at Mocha.loadFiles (D:\_host\github\rollup\node_modules\mocha\lib\mocha.js:213:14)
    at Mocha.run (D:\_host\github\rollup\node_modules\mocha\lib\mocha.js:453:10)
    at Object.<anonymous> (D:\_host\github\rollup\node_modules\mocha\bin\_mocha:393:18)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

eventualbuddha:
Are you using `rollup-plugin-babel`? If not, rollup won't do anything to transform `const` to `var`.

TrySound:
@eventualbuddha It's rollup itself. I can't test it

eventualbuddha:
What version of node, rollup, etc?

TrySound:
node0.12, frash pull from original master

TrySound:
Maybe something wrong with rollup-plugin-babel1?

eventualbuddha:
I'm having trouble reproducing this. Can you provide a specific command that fails?

TrySound:
No fail, just npm run build gives me code with constants

TrySound:
Of course tests fails after this

[...]

PR #311, opened 2015-11-16
Fix test for rejection when options or options.entry isn't passed

Permutatrix:
The test as it's currently written won't fail under any circumstances. This pull request fixes that and splits it up into two tests, one for `options` and one for `options.entry`.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -46,13 +46,19 @@ describe( 'rollup', function () {
 			assert.equal( typeof rollup.rollup, 'function' );
 		});
 
-		it( 'fails without options or options.entry', function () {
-			rollup.rollup().catch( function (err) {
-				assert( /must supply options\.entry/.test( err.toString() ) );
+		it( 'fails without options', function () {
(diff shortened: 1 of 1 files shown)


Permutatrix:
The AppVeyor build seems to have failed because of #310. :/

Victorystick:
@Permutatrix Good catch! This test wasn't updated in #303. Nice of you to fix it.

I'm used to Appveyor failing by now, but #310 is strange. Could it possibly be Windows specific?

TrySound:
@Victorystick Seems like babel doesn't work here at all.

Permutatrix:
@Victorystick You can rule out it being Windows-only; I'm using Arch Linux.

Permutatrix:
No, wait. It's AppVeyor failing, it works fine on my PC, so maybe it is Windows-only. Sorry, I just got up.

Rich-Harris:
Thanks! #310 is bewildering, I agree. I don't have a very good way to look into it at the moment though :/

PR #320, opened 2015-11-20
Move declaration assignment before the if statement

### LINKED ISSUE #314: namespace optimisation fails for default
```js
import * as foo from 'foo';

foo.default();
```

The static namespace optimisation fails for identifiers named `default`, see [this example](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22%2F%2F%20ES6%20modules%20let%20you%20import%20all%20of%20another%20module's%5Cn%2F%2F%20exports%20as%20a%20namespace...%5Cnimport%20*%20as%20assert%20from%20'.%2Fassert'%3B%5Cn%5Cn%2F%2F%20...but%20we%20can%20statically%20resolve%20this%20to%20the%5Cn%2F%2F%20original%20function%20definition%5Cnassert.default(%201%20%2B%201%2C%202%20)%3B%22%7D%2C%7B%22name%22%3A%22assert.js%22%2C%22code%22%3A%22export%20default%20function%20equal%20(%20a%2C%20b%2C%20msg%20)%20%7B%5Cn%5Ctif%20(%20!msg%20)%20msg%20%3D%20format(%20'%25s%20does%20not%20equal%20%25s'%2C%20a%2C%20b%20)%3B%5Cn%5Ctif%20(%20a%20!%3D%20b%20)%20throw%20new%20Error(%20msg%20)%3B%5Cn%7D%22%7D%5D%7D). The namespace code doesn't get included either.

Victorystick:
Fixes #314

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -76,10 +76,12 @@ export class SyntheticDefaultDeclaration {
 	}
 
 	addReference ( reference ) {
+		// Bind the reference to `this` declaration.
+		reference.declaration = this;
+
 		// Don't change the name to `default`; it's not a valid identifier name.
 		if ( reference.name === 'default' ) return;
 
-		reference.declaration = this;
 		this.name = reference.name;
 	}
 
--- a/test/function/resolves-namespace-default/_config.js
+++ b/test/function/resolves-namespace-default/_config.js
@@ -1,0 +1,9 @@
+var assert = require('assert');
+
+module.exports = {
+  description: "namespace's 'default' properties should be available",
+
+  exports: function ( exports ) {
+    assert.equal( exports, 42 );
+  }
+};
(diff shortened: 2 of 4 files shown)


eventualbuddha:
I verified that the added test fails on master. I don't fully understand the change, but… :+1:

Victorystick:
When we try to bind the reference `mod.default` to the declaration of `mod`'s default export, there was an if statement that aborted the binding if the name was `"default"`, since it also manages some renaming logic. Since we were binding to default, the whole thing failed. Moving the if makes the binding succeed, and the renaming fail. Just the way we want it.

Victorystick:
@Rich-Harris Merge-able?

Rich-Harris:
:+1:

PR #328, opened 2015-11-28
treat missing relative imports as error not warning - fixes #321

### LINKED ISSUE #321: Never treat modules starting with dot as external dependency
The problem I've encountered:
```js
import foo from './baar'; // I ment 'bar' but mistyped it
```
Rollup was ok with that, and just treated it as external:
```
Treating './baar' as external dependency
```
But this is clearly an error. How about if any module name starts with dot `.` and can't be found, raise error instead of treating it as external dependency?

Rich-Harris:
#321

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -161,6 +161,7 @@ export default class Bundle {
 			return Promise.resolve( this.resolveId( source, module.id ) )
 				.then( resolvedId => {
 					if ( !resolvedId ) {
+						if ( source[0] === '.' ) throw new Error( `Could not resolve ${source} from ${module.id}` );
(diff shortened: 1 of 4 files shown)


src/Bundle.js
```diff
@@ -161,6 +161,7 @@ export default class Bundle {
 			return Promise.resolve( this.resolveId( source, module.id ) )
 				.then( resolvedId => {
 					if ( !resolvedId ) {
+						if ( source[0] === '.' ) throw new Error( `Could not resolve ${source} from ${module.id}` );
```
eventualbuddha: Technically this should be starting with `./` or `../` since you could have a file or package starting with a period.

Rich-Harris: good catch – patch inbound

eventualbuddha:
:+1:

PR #348, opened 2015-12-06
Prevent invalid keys from being passed in options.

eventualbuddha:
These are all the keys needed to make the test pass, but I may have missed something.

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -38,7 +38,11 @@ export function rollup ( options ) {
 		return Promise.reject( new Error( 'The `transform`, `load`, `resolveId` and `resolveExternal` options are deprecated in favour of a unified plugin API. See https://github.com/rollup/rollup/wiki/Plugins for details' ) );
 	}
 
-	validateKeys( options, ALLOWED_KEYS );
+	const error = validateKeys( options, ALLOWED_KEYS );
+
+	if ( error ) {
+		return Promise.reject( error );
+	}
 
 	const bundle = new Bundle( options );
 
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -2,10 +2,32 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
 import { basename } from './utils/path.js';
 import { writeFile } from './utils/fs.js';
 import { keys } from './utils/object.js';
+import validateKeys from './utils/validateKeys.js';
 import SOURCEMAPPING_URL from './utils/sourceMappingURL.js';
 import Bundle from './Bundle.js';
 
 export const VERSION = '<@VERSION@>';
+
+const ALLOWED_KEYS = [
+	'entry',
+	'dest',
+	'plugins',
+	'external',
+	'onwarn',
+	'indent',
+	'format',
+	'moduleName',
+	'sourceMap',
+	'intro',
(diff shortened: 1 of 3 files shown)


eventualbuddha:
@Rich-Harris ping.

Victorystick:
@eventualbuddha Good idea. I see you're throwing an exception instead of returning a rejected Promise.

Since you're doing argument validation, throwing makes more sense to me. The reason I point this out is that the _two other argument validation_ conditions return a rejected Promise instead. I think those should probably throw as well.

eventualbuddha:
@Victorystick ah, good point. Those should be consistent, I agree. It might make more sense for the function to always return a promise, though. That way callers simply need to have a `.catch(` call chained as the sole way to handle errors. Making `validateKeys` return a promise would be a little strange, though, so perhaps it's best to wrap it somehow in the call to `rollup`?

Victorystick:
Isn't there a semantic difference between a rejected Promise and a thrown Error? I agree that it's better to have a single way to catch errors, but these are two completely different kinds of errors, no?

eventualbuddha:
Imagine a caller of the `rollup` function wants to handle errors. With the
way it's done now there are two kinds of error: immediately thrown errors
and rejected-promise errors. That means you need something like this:

```js
try {
rollup(…)
.then(…)
.catch(/* rejected promise handler */);
} catch (err) {
/* immediately thrown handler */
}
```

Whereas if we always return a promise you can drop the surrounding
`try/catch`.

TrySound:
@eventualbuddha If we use promises in rollup then we should always reject and error with it. As an alternative you can wrap all in Promise and throw and error there
```js
function rollup() {
  return Promise.resolve().then(function () {
    throw Error('Some error');
  });
}
```
This will be the same as returning Promise.reject();

arv:
FWIW, this is the same motivation why ES'15 Promise functions do not throw but return rejected promises.

```js
Promise.all(42); // rejected Promise
```

(V8 gets this wrong [#4603](https://code.google.com/p/v8/issues/detail?id=4603))

TrySound:
@arv Doesn't matter in this case

eventualbuddha:
I'll redo it with the semantics discussed.

eventualbuddha:
Done. Added a test for it as well.

Rich-Harris:
I took the liberty of replacing the `for...of` loop with a `while` – with our current build setup, `for...of` isn't actually transpiled. We could add that setting to the `.babelrc`, but it [doesn't transpile well](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&playground=false&code=for%20(%20let%20key%20of%20actualKeys%20)%20%7B%0A%09%09if%20(%20allowedKeys.indexOf(%20key%20)%20%3C%200%20)%20%7B%0A%09%09%09return%20new%20Error(%0A%09%09%09%09%60Unexpected%20key%20'%24%7B%20key%20%7D'%20found%2C%20expected%20one%20of%3A%20%24%7B%20allowedKeys.join(%20'%2C%20'%20)%20%7D%60%0A%09%09%09)%3B%0A%09%09%7D%0A%09%7D), so I tend to include in the 'ES6 features to avoid' category.

PR #353, opened 2015-12-11
add write hook for plugins

tylerbeck:
I would like to be able to create an import plugin that returns information about an object during the generate process and then outputs data accumulated through the generate process to a file outside of the destination JS file. Concatenated CSS is one example;  I would like use imports to keep componentized css, but then output a single CSS file as part of the build.  

The following pull request adds the ability for a plugin to expose a 'write' method that is called in the bundle.write method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ node_modules
 dist
 _actual
 coverage
+test-out
 .commithash
 .idea
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "pretest": "npm run build",
-    "test": "mocha",
+    "test": "mocha && rm -rf test-out/",
     "pretest-coverage": "npm run build",
     "test-coverage": "rm -rf coverage/* && istanbul cover --report json node_modules/.bin/_mocha -- -u exports -R spec test/test.js",
     "posttest-coverage": "remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped.json -b dist && remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped.lcov -t lcovonly -b dist && remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped -t html -b dist",
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -16,6 +16,7 @@ import collapseSourcemaps from './utils/collapseSourcemaps.js';
 import SOURCEMAPPING_URL from './utils/sourceMappingURL.js';
 import callIfFunction from './utils/callIfFunction.js';
 import { isRelative, resolve } from './utils/path.js';
+import { getMethod } from './utils/plugin.js';
 
 export default class Bundle {
 	constructor ( options ) {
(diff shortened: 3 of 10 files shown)


TrySound:
@tylerbeck Such things as css or another additional data should be exported with callbacks, Rollup shouldn't response for any other files. It's not build tool for them.

eventualbuddha:
How does this fit into your build system? I'm curious.

tylerbeck:
The plugin I'm working on allows you to reference and or process external assets using inline, nestable commands in import statements. Take the following:
```
// UIComponent.js
import { path as icon } from "img( ../assets/images/icon_normal.png,   )";
import { path as iconActive } from "img( ../assets/images/icon_active.png  )";
import styles from "css( ../assets/styles/UIComponentStyle.css  )";
```
In this scenario, the two `img` processors and`css` processor (if configured to do so) would register copy callbacks on the plugin, which would ideally get called in bundle.write to ensure the files that are referenced by the imports are in the proper places; not all files can or should be wrapped and injected by JS.  Rollup would remain agnostic to the files... the responsibility for managing them would be on the plugin.

Victorystick:
What you want could be achieved using the `footer` plugin method. It gets called one for each call to either `bundle.write` or `bundle.generate`.

```js
plugins: [
  {
    footer() {
      // copying code...

      // don't return anything (or it will be appended to the bundle's output
    }
  }
]
```

Rich-Harris:
Interesting. It seems like semantically `write` isn't the proper hook here, since it's not actually *influencing* what gets written, whereas other plugin hooks modify the `intro` or `outro` or `footer` or transform individual modules.

So maybe we need a separate class of hooks, like `oncomplete`, that get called at various points but don't otherwise affect the process. In the case of `onwrite`, perhaps we'd want to pass in the destination (e.g. if you wanted to write out your CSS or whatever in a location relative to the JS bundle).

Any thoughts?

Victorystick:
Something like `oncomplete` makes more sense for these purposes.

tylerbeck:
I think that `oncomplete` sounds reasonable.  I'll work on a new PR that includes this feature.

tylerbeck:
Now that I've had some time to look at the code, I think multiple event-type hooks may make sense... the obvious points of interaction would be pre and post generate, and post write.  The `onwritecomplete` hook is relatively easy to implement as the `rollup.write` method returns a promise.  The generate hooks are more complicated;  `rollup.generate` returns an object (from `Bundle.render`) ... What do you all think about returning a promise from generate?  The promise object could potentially be appended with `code` and `map` in order to NOT break the existing interface to `rollup`

Rich-Harris:
I'm not sure I understand why `bundle.generate` should return a promise? It's not doing anything asynchronous.

tylerbeck:
Rich,  I think that allowing plugins to implement event hooks asynchronously is important.  Take a look at the last commit I made... rollup.generate now returns a promise that is resolved once plugin `on generate` methods are resolved, but `code` and `map` are still available as properties on the returned value.

TrySound:
I don't think promise modifying is a good and safe way.

tylerbeck:
The other option that I thought could work would be to just add the hook promise to the generate return value e.g. `{ code, map, promise }`

tylerbeck:
Hey guys, what else do you think needs to be done before this is merged?
As it currently stands, this PR adds the following functionality:

1) Expose `onwrite` lifecycle hook to plugins.  The `options` parameter passed to `bundle.write` is passed on to each plugin. The Promise returned by `bundle.write` method will now resolve once all `onwrite` hooks and the bundle write and bundle.map write resolve.

2) Expose `ongenerate` lifecycle hook to plugins.  The `options` parameter passed to `bundle.generate` is passed on to each plugin. The object returned by `bundle.generate` has an `ongenerate` attribute which is a promise that resolves once all hooks resolve; the existing `code` and `map` parameters are unchanged.

3) Added `utils/plugin.js` file with a few convenience methods for executing hooks on plugins.

Swatinem:
Thanks for pointing me to this PR.

This is more or less what I had in mind. I would prefer the hooks to be serial rather than parallel though. So for example the (on)write hook should have the guarantee that the `dest` file is already written.

Rich-Harris:
Sorry for the long, long delay on this. I've opened #742 which is similar to this, but with some important changes I think are necessary for this to be viable (serial rather than parallel as @Swatinem mentioned, and synchronous `ongenerate` hooks – though `onwrite` can be async)

Rich-Harris:
I've merged #742 so will close this – thanks all

PR #356, opened 2015-12-14
Ensure namespace imports are not combined with other import types.

### LINKED ISSUE #355: Illegal imports
hi,

im trying to import an other module of mine (which was compiled by rollup to es2015) in my current module. But it seems there is an syntax error in the generated *import* in the other module.

```
SyntaxError:  /myproject/node_modules/js-utils/index.es2015.js: Unexpected token (1:14)
> 1 | import * as fs, { FSWatcher, ReadStream, Stats, WriteStream, createReadStream, createWriteStream, unwatchFile, watch, watchFile } from 'fs';
    |               ^
  2 | 
  3 | /* eslint-disable no-console */
  4 | function asyncMain(func) {
```

changing the generated code to:
```javascript
import * as fs from 'fs';
import { FSWatcher, ReadStream, Stats, WriteStream, createReadStream, createWriteStream, unwatchFile, watch, watchFile } from 'fs';
```
~~fixed~~ workarounded the error.

ps: 
[@see my build script (other module)](https://github.com/sudeti/js-utils/blob/f7397e8c00dea4fcd1c5ee192d702bda69bfa3d2/gulpfile.babel.js#L31-L47)
[@see source code (other module)](https://github.com/sudeti/js-utils/blob/f7397e8c00dea4fcd1c5ee192d702bda69bfa3d2/src/fs.js)
[@see my build script (current module)](https://github.com/rollup/rollup-plugin-commonjs/issues/21)
pps: maybe it's a bug in babel

eventualbuddha:
It is a syntax error to combine namespace import specifiers with named or default specifiers.

Fixes #355.

### DIFF
--- a/src/finalisers/es6.js
+++ b/src/finalisers/es6.js
@@ -8,6 +8,7 @@ export default function es6 ( bundle, magicString ) {
 	const importBlock = bundle.externalModules
 		.map( module => {
 			const specifiers = [];
+			const specifiersList = [specifiers];
 			const importedNames = keys( module.declarations )
 				.filter( name => name !== '*' && name !== 'default' );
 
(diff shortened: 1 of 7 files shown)


Victorystick:
Good work! One question though; must namespace imports always stand on their own? I think it can be combined with default imports?

```js
import foo, * as bar from './baz';
```

eventualbuddha:
@Victorystick it seems [you're right](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=import%20foo%2C%20*%20as%20bar%20from%20'.%2Fbaz'%3B). Fixed to only split them if both are present.

Rich-Harris:
:+1:

Issue #345, opened 2015-12-04
[RFC] Multiple entry points


eventualbuddha:
I created [rollup-plugin-multi-entry](https://github.com/eventualbuddha/rollup-plugin-multi-entry) as an approach to solving the multiple entry point problem. It's pretty naive, but works for my use cases. I'm curious to hear feedback on it.

One thing that would make it better is if there was a plugin hook to manipulate options. That way you could simply omit `entry` and let the plugin add the unique token for you:

```js
// rollup.config.js
import multiEntry from 'rollup-plugin-multi-entry';

export default {
  plugins: [multiEntry(['test/**/*.js'])]
};
```

I realize there's potential for weirdness with the plugins coming in as part of the initial options, but I think it could be made sane. Thoughts?

Rich-Harris:
I think this is great, really elegant. Solves the problem in a way that allows people to be highly efficient (e.g. glob syntax) without adding any opinions to the core library.

I like the idea of plugins manipulating options. Opened #371. I think it's reasonable to just make it clear in the documentation that plugins don't get to muck around with `options.plugins`. Will close this in favour of #371.

eventualbuddha:
I finally got around to updating rollup-plugin-multi-entry to use the new `options` hook. Here's what it looks like [in rollup-starter-project](https://github.com/rollup/rollup-starter-project/commit/84737b90c5cdc8de0b1d7ac1ac7c4af0ceb741b3):

```js
import babel from 'rollup-plugin-babel';
import multiEntry from 'rollup-plugin-multi-entry';

export default {
  entry: 'test/**/*_test.js',
  plugins: [babel(), multiEntry()],
  format: 'cjs',
  dest: 'build/test-bundle.js'
};
```

Luchillo:
That uses globbing to allow multiEntry right? what about if one needs to select specific files in different paths to be used as entries? also is there a way to execute each entry and output 1 file per entry? maybe i'm thinking a little to much with Webpack configuration in mind.

snuggs:
Then perhaps specify the specific paths as the entry points @Luchillo?

snuggs:
@Luchillo try rollem. Does enough to do the job.

PR #371, opened 2015-12-22
allow plugins to manipulate options

Rich-Harris:
Follow-up to @eventualbuddha's idea on #345 – allows plugins to register an `options` hook that either manipulates the `options` object, or return a replacement one.


eventualbuddha:
:+1: I like it

Rich-Harris:
Released as 0.22 – docs updated

PR #382, opened 2015-12-29
Handle stringified sourcemaps

### LINKED ISSUE #377: Handle stringified sourcemaps
Via https://github.com/egoist/rollup-plugin-postcss/issues/1#issuecomment-166938098. It's not uncommon for compilers to return sourcemaps as JSON rather than as objects, so perhaps it would make sense for Rollup to detect when that's the case and handle it accordingly. Would be a one-liner somewhere around [here](https://github.com/rollup/rollup/blob/cabc4ee097c6b0ee1b03c300709233ef9495d293/src/utils/transform.js#L29)

Victorystick:
Fixes #377

### DIFF
--- a/src/utils/transform.js
+++ b/src/utils/transform.js
@@ -25,6 +25,10 @@ export default function transform ( source, id, transformers ) {
 						map: null
 					};
 				}
+				// `result.map` can only be a string if `result` isn't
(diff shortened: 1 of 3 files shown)


Rich-Harris:
:+1:

PR #383, opened 2015-12-29
support namespaced exports (global.foo.bar.baz)

### LINKED ISSUE #378: Support name-spaced UMD module names
I'd like to set the `--name` option to be something like: `Foo.bar`. Rollup will accept this as input and output `Global.Foo.bar = ...` in the UMD module, but this will break if `Foo` isn't already defined.

Browserify supports name-spaced module names by recognizing this pattern and generating code to make sure the object is defined before the member expression assignment. Something like:

```js
(global.Foo || (global.Foo = {})).bar = ...
```

It would be great if Rollup also supported this for name-spaced UMD module names.

Rich-Harris:
Fixes #378


Victorystick:
:+1:

mbostock:
:+1: :tada:

Issue #381, opened 2015-12-29
Bundle transform hook


Rich-Harris:
Like the module transform hook, except for the bundle – takes care of sourcemap combining etc. Would enable e.g. uglifyjs plugin

Rich-Harris:
Added in 0.23

PR #387, opened 2015-12-30
Add transformBundle plugin hook

TrySound:
Need to decide what to do with promises. This does not seem like a good idea.
```js
.then(bundle => {
  return bundle.generate({...});
}).then(({ code, map }) => {})
```

### DIFF
--- a/src/utils/transformBundle.js
+++ b/src/utils/transformBundle.js
@@ -1,0 +1,39 @@
+import MagicString from 'magic-string';
+
+export default function transformBundle ( source, transformers ) {
+	if ( typeof source === 'string' ) {
+		source = {
+			code: source,
+			map: null
+		};
+	}
+
+	return transformers.reduce( ( previous, transformer ) => {
+		let result = transformer( previous.code, previous.map );
+
+		if ( result == null ) return previous;
+
+		if ( typeof result === 'string' ) {
+			result = {
+				code: result,
+				map: null
+			};
+		}
+		// `result.map` can only be a string if `result` isn't
+		else if ( typeof result.map === 'string' ) {
+			result.map = JSON.parse( result.map );
+		}
+
+		if (result.map != null) {
+			let map = new MagicString.Bundle().generateMap({});
+			map.file = result.map.file;
+			map.sources = result.map.sources;
+			map.sourcesContent = result.map.sourcesContent;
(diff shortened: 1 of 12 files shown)


Rich-Harris:
Ah, good point. I think it's okay to expect bundle transformers to be synchronous, rather than changing the API to be async – after all, if someone *really* needs to do something asynchronously, this is the one place where they can do that, since it's the last stage in the process.

TrySound:
@Rich-Harris Do you have any test suggestions?

src/utils/transformBundle.js
```diff
@@ -0,0 +1,28 @@
+export default function transformBundle ( source, transformers ) {
+	if ( typeof source === 'string' ) {
+		source = {
+			code: source,
+			map: null
+		};
+	}
+
+	return transformers.reduce( ( previous, transformer ) => {
+		let result = transformer( previous );
```
Rich-Harris: Module transformers are called with the code from the previous step, rather than a `{code, map}` object – I think we should stick to that, because it makes transformers a lot more straightforward. Somehow we need to figure out how to compose all the sourcemaps together :grin:

Rich-Harris:
@TrySound we should probably have some sourcemap tests. Maybe we should go the whole hog and try uglifying code to check the intended use case?

src/utils/transformBundle.js
TrySound: I think composing sourcemaps should to transformer. Like with uglify we just pass previous map.

Rich-Harris: Uglify is very rare in that it accepts an input sourcemap. Sourcemaps are bewildering; expecting people to manage that complexity creates a huge burden (worrying about the sourcemaps from your *own* transformation step is hard enough, let alone someone else's), and in practice would mean that this hook never got used for anything except Uglify.

TrySound:
Fix https://github.com/rollup/rollup/issues/381

TrySound:
@Rich-Harris Let's merge this and you can rebase your branch from master.

Rich-Harris:
@TrySound what's the advantage to doing it that way? Both branches have the green light (#391 is based on this branch)

TrySound:
@Rich-Harris Clean commits history is good advantage for me.

Issue #385, opened 2015-12-29
painful to get started


boneskull:
This project needs some better tutorials or examples.  Here's my use case:
1.  I have code written in ES2015; I am using Babel to transpile
2.  I am also using npm dependencies
3.  These dependencies (mainly) use CommonJS instead of ES2015 modules
4.  This is a test runner with a programmatic API
5.  The resulting build(s) must be runnable in Node.js, a browser, and a CLI context

I've discovered the Babel plugin, the NPM plugin, and the CommonJS plugin.
It's unclear:
1.  In what order the plugins are supposed to run
2.  If order actually matters (though I'm pretty sure it does)
3.  Why the CommonJS plugin and NPM plugin are separated at all
4.  How to configure the Babel plugin to work with the above two plugins
5.  Why the Babel plugin is necessary
6.  Where Babel's responsibilities end and Rollup's begin--don't you need Babel to use `import` statements anyway?
7.  If/when Browserify or Webpack is necessary
8.  How to configure Karma to make my tests run in a browser
9.  How to explain to my users, in a README, how to include my library given a matrix of potential configurations (Node.js, browser, Babel, no-Babel, Rollup, etc)
10. When to use a config file or when and how to use a build script (`import rollup from 'rollup'`) 

I'm here asking these questions, there's probably 5 other people who had the same ones and didn't post an issue.  Surely I'd like my questions addressed, but the real need is more documentation.

Thanks!

boneskull:
As an aside, Babel "presets" are very helpful in that they provide an abstraction which doesn't require the user to manually include *n* modules and configure them all.  Perhaps packages for common use cases would take some responsibility off of the user's shoulders?

eventualbuddha:
@boneskull Thanks for the thorough issue notes! I assume you've looked at the [wiki](https://github.com/rollup/rollup/wiki)? Were you looking for a Getting Started page and didn't find it? Did you find [rollup-starter-project](https://github.com/rollup/rollup-starter-project)? Here are a few answers to your questions, though as you said the better thing to do is to create better documentation.

> In what order the plugins are supposed to run
> If order actually matters (though I'm pretty sure it does)

The order does matter since they run in the order listed in the `plugins` array for each of their [hooks](https://github.com/rollup/rollup/wiki/Plugins#creating-plugins). You can think of plugins as a pipeline, where each plugin can modify the results of the one before it.

> How to configure the Babel plugin to work with the above two plugins?

[...]

PR #388, opened 2015-12-30
support --environment in CLI

### LINKED ISSUE #376: --environment argument for CLI
Discovered that you can't do `export FOO=bar && rollup -c` on Windows. I'm not sure if there's a way to provide environment variables to an `npm run` script that works in both Windows and *nix. So it might be helpful if you could do this sort of thing:

```bash
rollup -c --environment FOO=bar
```

Any thoughts on whether this is the best way? Examples from other tools?

Rich-Harris:
Closes #376. Adds an `--environment` option to the CLI with the following syntax, modelled after the syntax for `-g`:

```bash
rollup -c --environment PRODUCTION,FOO:bar,BAZ:qux
```

This results in the following conditions inside the `rollup.config.js` file:

```js
process.env.PRODUCTION === 'true'; // all process.env properties must be strings
process.env.FOO === 'bar';
process.env.BAZ === 'qux';
```

### DIFF
--- a/bin/help.md
+++ b/bin/help.md
@@ -20,14 +20,23 @@ Basic options:
 -m, --sourcemap          Generate sourcemap (`-m inline` for inline map)
 --no-strict              Don't emit a `"use strict";` in the generated modules.
 --no-indent              Don't indent result
+--environment <values>   Settings passed to config file (see example)
 
 Examples:
 
+# use settings in config file
 rollup -c
 
+# in config file, process.env.INCLUDE_DEPS === 'true'
+# and process.env.BUILD === 'production'
+rollup -c --environment INCLUDE_DEPS,BUILD:production
+
+# create CommonJS bundle.js from src/main.js
 rollup --format=cjs --output=bundle.js -- src/main.js
 
-rollup -f iife --globals jquery:jQuery,angular:ng \
+# create self-executing IIFE using `window.jQuery`
+# and `window._` as external globals
+rollup -f iife --globals jquery:jQuery,lodash:_ \
   -i src/app.js -o build/app.js -m build/app.js.map
 
 Notes:
--- a/bin/runRollup.js
+++ b/bin/runRollup.js
@@ -20,6 +20,17 @@ module.exports = function ( command ) {
 		command.input = command._[0];
 	}
 
+	if ( command.environment ) {
+		command.environment.split( ',' ).forEach( function ( pair ) {
+			var index = pair.indexOf( ':' );
+			if ( ~index ) {
+				process.env[ pair.slice( 0, index ) ] = pair.slice( index + 1 );
+			} else {
+				process.env[ pair ] = true;
+			}
+		});
+	}
+
 	var config = command.config === true ? 'rollup.config.js' : command.config;
 
 	if ( config ) {
--- a/test/cli/config-env/_config.js
+++ b/test/cli/config-env/_config.js
@@ -1,0 +1,5 @@
+module.exports = {
+	description: 'passes environment variables to config file',
+	command: 'rollup --config --environment PRODUCTION,FOO:bar',
+	execute: true
+};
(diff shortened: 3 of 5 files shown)


eventualbuddha:
👍🏻

himalayafan:
@Rich-Harris Is this only for CLI? If you want to do the same inside rollup.rollup{}, how you do it?

I was re-thinking Rollup, and now with the new commits, I start to like it! Maybe it's should be a step-by-step intro page for nobs for me so people can get into this quickly?

Rich-Harris:
> Is this only for CLI?

Yep, the assumption is that if you're using the JavaScript API you already have full control over the environment. So you can vary any of the options based on whatever conditions are relevant. Is there something specific you're unsure how to do?

> I was re-thinking Rollup, and now with the new commits, I start to like it! Maybe it's should be a step-by-step intro page for nobs for me so people can get into this quickly?

Welcome back! #385 has some anecdotes from other people who struggled initially; it seems like some intro tutorials or screencasts would be a good idea. Hopefully we'll get round to that soon

himalayafan:
Issue for me is Mac vs Windows. If I use Rollup config file and want it to react on a variable environment, I have to use two build commands in the script section in the package.json. One for mac and one for windows.

`"NODE_ENV=development babel-node scripts/build-dist"`
`"set NODE_ENV=development babel-node scripts/build-dist"`

I didn't found a workaround for this, or can I use Rollup CLI here, and call the config file with it? 

Yep! Would be awsome if you get some tutorials and screencasts. Would helped me and others a lot!

thisconnect:
@himalayafan I'd love if NPM would provide a first class way to set NODE_ENV and others cross-platform. FYI there is https://www.npmjs.com/package/cross-env

Rich-Harris:
You could certainly use the Rollup CLI here...

```js
{
  "scripts": {
    "build": "rollup -c --environment ENV:development"
  }
}
```

...unless your build script has a lot of custom stuff (since `rollup.config.js` is basically just returning a straightforward options object – it gives you a lot of flexibility, but not *total* flexibility). If you wanted to set environment variables inside `scripts/build-dist` you *should* be able to do it like so:

```js
{
  "scripts": {
    "build": "babel-node scripts/build-dist dev"
  }
}
```

...and then inside `build-dist.js` you would have something along these lines:

```js
var target = process.argv[2] === 'dev' ? 'development' : 'production';

/* rest of the build script */
```

thisconnect:
`"babel-node scripts/build-dist dev"` was not enough for me 
```
(function (exports, require, module, __filename, __dirname) { import { rollup } from 'rollup';
                                                              ^^^^^^
SyntaxError: Unexpected token import
```

`"babel-node --presets es2015 scripts/build-dist dev"` did work for me.
Note: it has to be `es2015`, (not `es2015-rollup`). Since there are many references in the wiki, I would like to have a short "run babel" section within the JavaScript API  is that ok?

Rich-Harris:
That makes sense, running Rollup with Babel seems to be a common point of confusion. Until we have such a guide, I recommend having two `.babelrc` files – one in the root of your project that includes a modules transformer...

```js
// my-project/.babelrc
{
  "presets": ["es2015"]
}
```

...and one that goes in the `src` folder (or wherever you keep the files that Rollup is bundling) that doesn't:

```js
// my-project/src/.babelrc
{
  "presets": ["es2015-rollup"]
}
```

This way, the appropriate configuration will be applied at the appropriate time.

thisconnect:
Makes sence. I guess I always disabled the `.babelrc` to have all the configs in one place. Personally I prefer to use the cli arg and rollup configs over having multiple `.babelrc`'s

Rich-Harris:
@thisconnect you have to be careful though – options passed in directly will be merged with the nearest `.babelrc`, with unexpected results. I don't know if that behaviour has changed in more recent versions, but I got some pretty wild results with duplicated plugins at one point

thisconnect:
@Rich-Harris good to know, thanks. `'babelrc': false,` will protect me from the next `.babelrc` that wants to merge with my options.

Issue #344, opened 2015-12-03
CLI equivalent of sourceMapFile?


developit:
Hi @Rich-Harris & co:

I can't seem to find a way to specify `sourceMapFile` from the rollup CLI.  This would be useful as it triggers relative paths in the generated external sourcemap, which are better handled by uglify-js.

Am I just missing something, or is this a mapping that could be added?

Victorystick:
Rollup's `-m` flag is overloaded. `-m inline` results in an inline source map. Any other name creates a new file with that name. Try `rollup entry.js -d out.js -m another-name.js.map`.

Rich-Harris:
Actually it doesn't look like there's a way to specify a different sourcemap file name – in the example above the `another-name.js.map` part gets discarded.

Done some digging and it turns out that the `dest` option is being passed from the CLI to Rollup, but it's relative to `process.cwd()` (i.e. `-o out.js` is passed straight through). Because of that, the sources are absolute (because magic-string, which is actually generating the sourcemaps, can't know that `/path/to/my-project/src/foo.js` is in fact a short hop away from `/path/to/my-project/dist/bundle.js` when all it sees is `bundle.js`.

I think what we need to do is ensure that the `dest` or `sourceMapFile` option is always resolved. PR inbound

Rich-Harris:
Closing, as fixed in latest releases

PR #389, opened 2015-12-30
ensure dest is resolved against cwd, to make sourcemap paths relative when using CLI

Rich-Harris:
Fixes the issue described in #344


developit:
Perfect! Your test matches the issue I ran into, thanks Rich.

PR #392, opened 2015-12-30
Fix UMD global export

ericf:
This wraps the comma expression generated when using a namespaced global module name; e.g.: `Foo.bar`. Previously the generated code would fail in a CommonJS or AMD module environment.

See: https://github.com/rollup/rollup/issues/378#issuecomment-168063404


src/finalisers/umd.js
```diff
@@ -50,7 +50,7 @@ export default function umd ( bundle, magicString, { exportMode, indentString },
 		`(function (global, factory) {
 			typeof exports === 'object' && typeof module !== 'undefined' ? ${cjsExport}factory(${cjsDeps.join( ', ' )}) :
 			typeof define === 'function' && define.amd ? define(${amdParams}factory) :
-			${defaultExport}factory(${globalDeps});
+			(${defaultExport}factory(${globalDeps}));
```
ericf: The actual code change. The rest of the changes are for the test fixtures.

Rich-Harris:
Thanks @ericf!

Issue #293, opened 2015-11-11
Underscore-separated, not camel-cased, global names?


mbostock:
When using the UMD finalizer, dependencies on external modules in globals are referenced using `module.name` which is generated using [`makeLegalIdentifier`](https://github.com/rollup/rollup/blob/master/src/utils/makeLegalIdentifier.js). This replaces letters preceded by a hyphen with capitalized letters, such replacing “d3-arrays” with “d3Arrays”.

This is personal preference, but it’d be nice to substitute different logic for generating the global names, such as “d3_arrays”. I can do that using Rollup’s `-g` argument, but it’s a pain since I have to rename each external module. For example, in [d3-request](https://github.com/d3/d3-request):

```
rollup \
    -f umd \
    -g d3-arrays:d3_arrays,d3-dispatch:d3_dispatch,d3-dsv:d3_dsv \
    -u d3-request \
    -n d3_request \
    -o build/d3-request.js \
    -- build/bundle.js
```

It doesn’t look too hard to add an option that the UMD finalizer can use to substitute a different strategy for makeLegalIdentifier, but perhaps you have an opinion on whether this is the right approach?

Rich-Harris:
In the context of a command line option I think it's probably a bit tricky, because everyone who doesn't want the default behaviour is likely to have slightly different needs. Two options spring to mind, but both rely on either using the Node API or a config file:

* Throw an error if `globals` isn't supplied (since if you're relying on the default makeLegalIdentifiers behaviour and it works, it's basically blind luck...)
* Allow a function to be specified instead of a map

In the meantime, you could do it with a config file like so:

```bash
rollup -c
```

```js
// rollup.config.js
var pkg = require( './package.json' );

function snakeCase ( id ) {
  return id.replace( /-/g, '_' );
}

export default {
  entry: 'src/index.js',
  dest: 'build/' + pkg.name + '.js',
  format: 'umd',
  moduleId: pkg.name,
  moduleName: snakeCase( pkg.name ),
  globals: Object.keys( pkg.dependencies ).reduce( function ( obj, id ) {
    return obj[ id ] = snakeCase( id ), obj;
  }, {})
};
```

That way it could be shared between different modules with relative ease.

Victorystick:
I say enable both options.

* I don't think we should make Rollup's users rely on "blind luck" because of how `makeLegalIdentifier` works. We should make the globals option required and throw if we can't determine the name of a required global.
[...]

PR #396, opened 2015-12-30
allow options.globals to be a function, warn on miss

Rich-Harris:
Follow-up to #293. Allows `options.globals` to be a function as opposed to an object. In addition, if a global name isn't provided for an external module (in IIFE or UMD mode), Rollup will print a warning along these lines:

```
No name was provided for external module 'my-lib' in options.globals – guessing 'MyLib'
```

The 'guess' is based on how the external library is referenced in the bundle – if the code has something like...

```js
import MyLib from 'my-lib';
```

...it's a pretty safe bet that `MyLib` is the correct name. But in cases like this...

```js
import { foo } from 'named-only';
```

...we're forced to guess on the basis of the module ID. Either way, the recommendation is to be explicit.


Victorystick:
Looking good!

PR #400, opened 2015-12-31
Side-effect detection in cyclical function calls

### LINKED ISSUE #397: Side effect detection failure in cycle
I noticed this while [fiddling in the repl][1]

even.js:
```js
import { odd } from './odd'

export var counter = 0;

// This should be in the output
export const foo = odd(12);

export function even(n) {
  counter++;
  alert(counter)
  return n == 0 || odd(n - 1);
}
```

odd.js:
```js
import { even } from './even';

export function odd(n) {
  return n != 0 && even(n - 1);
}

export const bar = even(11);
```

With main.js like so:
```js
import { counter, even } from './even'

console.log(even(5))

console.log(counter)
```
The output is:
```js
function odd(n) {
  return n != 0 && even(n - 1);
}

const bar = even(11);

var counter = 0;

function even(n) {
  counter++;
  alert(counter)
  return n == 0 || odd(n - 1);
}

console.log(even(5))

console.log(counter)
```

with `foo` missing. 
If `console.log(even(5))` is commented out, `foo` magically appears!

```js
function odd(n) {
  return n != 0 && even(n - 1);
}

const bar = even(11);

var counter = 0;

// This should be in the output
const foo = odd(12);

function even(n) {
  counter++;
  alert(counter)
  return n == 0 || odd(n - 1);
}

//console.log(even(5))

console.log(counter)
```

[1]: http://rollupjs.org/#{%22options%22%3A{%22format%22%3A%22es6%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A{}%2C%22moduleId%22%3A%22%22}%2C%22modules%22%3A[{%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20{%20counter%2C%20even%20}%20from%20%27.%2Feven%27\n\n%2F%2F%20console.log%28even%285%29%29\n\nconsole.log%28counter%29%22}%2C{%22name%22%3A%22even.js%22%2C%22code%22%3A%22import%20{%20odd%20}%20from%20%27.%2Fodd%27\n\nexport%20var%20counter%20%3D%200%3B\n\n%2F%2F%20This%20should%20be%20in%20the%20output\nexport%20const%20foo%20%3D%20odd%2812%29%3B\n\nexport%20function%20even%28n%29%20{\n%20%20counter%2B%2B%3B\n%20%20alert%28counter%29\n%20%20return%20n%20%3D%3D%200%20||%20odd%28n%20-%201%29%3B\n}%22}%2C{%22name%22%3A%22odd.js%22%2C%22code%22%3A%22import%20{%20even%20}%20from%20%27.%2Feven%27%3B\n\nexport%20function%20odd%28n%29%20{\n%20%20return%20n%20!%3D%200%20%26%26%20even%28n%20-%201%29%3B\n}\n\nexport%20const%20bar%20%3D%20even%2811%29%3B%22}]}

Rich-Harris:
Fixes #397. It's not *totally* perfect, because if it encounters an infinite loop it's forced to assume that all functions in the cycle have side-effects, but it's a rare case and we're erring on the side of caution, so it's probably good enough.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -43,14 +43,20 @@ export default class Declaration {
 
 	run ( strongDependencies ) {
 		if ( this.tested ) return this.hasSideEffects;
-		this.tested = true;
+
 
 		if ( !this.functionNode ) {
 			this.hasSideEffects = true; // err on the side of caution. TODO handle unambiguous `var x; x = y => z` cases
 		} else {
+			if ( this.running ) return true; // short-circuit infinite loop
+			this.running = true;
+
 			this.hasSideEffects = run( this.functionNode.body, this.functionNode._scope, this.statement, strongDependencies, false );
+
+			this.running = false;
 		}
 
+		this.tested = true;
 		return this.hasSideEffects;
 	}
 
--- a/src/Module.js
+++ b/src/Module.js
@@ -570,7 +570,7 @@ export default class Module {
 		let marked = false;
 
 		this.statements.forEach( statement => {
-			marked = marked || statement.run( this.strongDependencies, safe );
+			marked = statement.run( this.strongDependencies, safe ) || marked;
 		});
 
 		return marked;
--- a/test/form/side-effect-m/_config.js
+++ b/test/form/side-effect-m/_config.js
@@ -1,0 +1,4 @@
+module.exports = {
+	solo: true,
+	description: 'detects side-effects with cyclical dependencies'
+};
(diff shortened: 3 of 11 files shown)


Victorystick:
I see... If a set of functions call each other recursively, if either of them has any side-effects then they all do. This is a good first step, but we should try to implement such logic, shouldn't we?

Rich-Harris:
Ideally, yes, definitely. Might be a bit awkward though so I reckon best move is to merge this to fix the bug, and add that as a separate enhancement

PR #406, opened 2016-01-01
handle a => a at top level

### LINKED ISSUE #403: Exception parsing arrow functions
I get an exception when using Rollup on arrow functions that return a single identifier:
```js
// main.js
const f = a => a;
console.log(f(1));
```
```
$ rollup -i main.js
Cannot read property 'type' of null
[...]
```
This also causes an exception:
```js
const a = 1;
const f = () => a;
console.log(f());
```
This works fine though:
```js
const f = (a => a);
console.log(f(1));
```

Rich-Harris:
Fixes #403

### DIFF
--- a/src/ast/isReference.js
+++ b/src/ast/isReference.js
@@ -4,6 +4,8 @@ export default function isReference ( node, parent ) {
 	}
 
 	if ( node.type === 'Identifier' ) {
+		if ( !parent ) return true;
+
 		// TODO is this right?
 		if ( parent.type === 'MemberExpression' ) return parent.computed || node === parent.object;
 
--- a/src/ast/isReference.js
+++ b/src/ast/isReference.js
@@ -4,6 +4,9 @@ export default function isReference ( node, parent ) {
 	}
 
 	if ( node.type === 'Identifier' ) {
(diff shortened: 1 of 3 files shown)


src/ast/isReference.js
```diff
@@ -4,6 +4,8 @@ export default function isReference ( node, parent ) {
 	}
 
 	if ( node.type === 'Identifier' ) {
+		if ( !parent ) return true;
```
Victorystick: Perhaps add a comment explaining why this exists? It's a special case.

eventualbuddha: It's not clear from the commit code alone why there'd be no parent. Is `parent` not the node that most immediately contains `node`? If it is, then only `Program` nodes don't have parents, right?

Rich-Harris: It's true that only `Program` doesn't have a parent, but as far as the enter handler goes it depends where you start the walk – in this situation we're starting at the function body. Ordinarily you couldn't reach an identifier node without first going through its parent, but there's an exception in this case, because an identifier can form the entire body of an arrow function expression. Adding a comment to the code

Issue #410, opened 2016-01-01
Allow relative requires to be treated as externals


guybedford:
If I have:

```javascript
import 'dep.js';
```

I can return `false` in `resolveId` to have dep excluded from the rollup optimization.

But if instead I have:

```javascript
import './dep.js';
```

Then returning `false` in `resolveId` will give the error _Could not resolve ./dep.js from test.js`_.

Because we need externals to occur across module format boundaries in SystemJS, and a relative module can be say CommonJS, this case breaks down here.

Victorystick:
This may be useful, but it could cause conflicts since `./dep.js` could point out any number of modules.

```
/project
  /src
    file.js
    dep.js
  index.js
  dep.js
```

Which `./dep.js` should be excluded? Any external modules must be an absolute path or an external name to avoid this ambiguity.

guybedford:
This is a great point, and I think this could be considered a conflict due to a conflation of unresolved and resolved namespaces.

This can be avoided by letting the `externals` excludes act against the **resolved** module list. That is still run the `resolveId` function for possible externals, then check exclusions within the resolved namespace.

This way, we always know exactly what we are excluding as it happens in a concrete resolved namespace which is known to be unique.

Rich-Harris:
Ha, we just passed each other @guybedford! Turns out great minds think alike – @Victorystick implemented it exactly as you suggest.

guybedford:
Awesome! Thanks for your great work here :)

joeldenning:
I've just been poring through the code and I'm not sure if I am missing something or if even after @Victorystick's pull request, rollup errors out because it's trying to resolve a relative import that I have specifically declared to be external. Below is a self contained example (I'm running rollup 0.25.4):

```
// rollup.config.js
export default {
    entry: 'src.js',
    dest: 'bundle.js',
    external: ['./dep.js'],
}
```

```
// src.js
import dep from './dep.js';

console.log(dep)
```

```
// dep.js
export default 1;
```

My expected result is that bundle.js would have an `import` statement for `./dep.js` still. But my actual result is the following error:
```
Error: Could not resolve ./dep.js from /Users/joel/playground/test/src.js
    at /Users/joel/playground/test/node_modules/rollup/www/ROLLUP/rollup/src/Bundle.js:186:42
    at tryCatch (/Users/joel/playground/test/node_modules/rollup/www/ROLLUP/rollup/node_modules/es6-promise/lib/es6-promise/-internal.js:185:12)
[...]

PR #411, opened 2016-01-02
prevent resolution of external IDs

Rich-Harris:
See #410. This ensures that Rollup doesn't attempt to resolve IDs that are in `options.external`.


eventualbuddha:
:+1:

PR #416, opened 2016-01-04
Remove 3 unused Promise.resolve calls

Victorystick:
Since `resolveId` and `load` always return Promises we don't need these! :raised_hands:


Rich-Harris:
good spot!

PR #418, opened 2016-01-04
Force external

### LINKED ISSUE #410: Allow relative requires to be treated as externals

Victorystick:
Force ids specified in `external` to actually _be_ external.
    
Fixes #410
Builds on top of #417

### DIFF
--- a/test/cli/config-external/main.js
+++ b/test/cli/config-external/main.js
@@ -1,0 +1,4 @@
+import { execute } from './_config.js';
+import { ok } from 'assert';
+
+ok( execute );
--- a/bin/runRollup.js
+++ b/bin/runRollup.js
@@ -88,7 +88,8 @@ var equivalents = {
 };
 
 function execute ( options, command ) {
-	var external = command.external ? command.external.split( ',' ) : [];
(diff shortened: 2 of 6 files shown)


test/cli/config-external/main.js
```diff
@@ -0,0 +1,4 @@
+import { execute } from './_config.js';
```
Rich-Harris: This appears to be [failing on Windows](https://ci.appveyor.com/project/Rich-Harris/rollup/build/611/job/leyq1rf3jk64fp03#L617) – presumably we need to normalise path separators or something?

ericf:
This change doesn't appear to be logging "Treating './foo.js' as external dependency" when it is in fact correctly _not_ bundling it.

PR #420, opened 2016-01-04
Force external (windows)

Rich-Harris:
#418, but with (hopefully) Windows support. Separate PR because I didn't want to mess up the commit history if this turns out to be wrong.

Damn you, Windows. Damn you forever


Victorystick:
Windows, right. :+1:

PR #423, opened 2016-01-05
handle calls to default exports other than function expressions

### LINKED ISSUE #421: Default exports with parens fails
Via https://github.com/rollup/rollup-plugin-babel/issues/28.

[This fails](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20foo%20from%20'.%2Ffoo.js'%3B%5Cnfoo()%3B%22%7D%2C%7B%22name%22%3A%22foo.js%22%2C%22code%22%3A%22%2F%2F%20remove%20the%20parens%20and%20it%20works%5Cnexport%20default%20(function%20()%20%7B%5Cn%20%20console.log(%5C%22Hi!%5C%22)%3B%5Cn%7D)%22%7D%5D%7D):

```js
// main.js
import foo from './foo.js';
foo();

// foo.js
export default (function () {
  console.log("Hi!");
})
```

Remove the parens around the default export and it works fine. It appears to be specific to functions ([this works fine](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20foo%20from%20'.%2Ffoo.js'%3B%5Cnconsole.log(foo)%3B%22%7D%2C%7B%22name%22%3A%22foo.js%22%2C%22code%22%3A%22export%20default%20(42)%22%7D%5D%7D)).

This is particularly problematic because this is how Babel transpiles arrow function expressions.

Rich-Harris:
Fixes #421

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -110,9 +110,15 @@ export class SyntheticDefaultDeclaration {
 			return this.original.run( strongDependencies );
 		}
 
-		if ( /FunctionExpression/.test( this.node.declaration.type ) ) {
-			return run( this.node.declaration.body, this.statement.scope, this.statement, strongDependencies, false );
+		let declaration = this.node.declaration;
+		while ( declaration.type === 'ParenthesizedExpression' ) declaration = declaration.expression;
+
+		if ( /FunctionExpression/.test( declaration.type ) ) {
+			return run( declaration.body, this.statement.scope, this.statement, strongDependencies, false );
 		}
+
+		// otherwise assume the worst
+		return true;
 	}
 
 	use () {
--- a/test/function/call-non-function-default-exports/_config.js
+++ b/test/function/call-non-function-default-exports/_config.js
@@ -1,0 +1,3 @@
+module.exports = {
+	description: 'calls non-function default exports'
+};
--- a/test/function/call-non-function-default-exports/foo.js
+++ b/test/function/call-non-function-default-exports/foo.js
@@ -1,0 +1,9 @@
+function x () {
+  global.answer = 'x';
+}
+
+function y () {
+  global.answer = 'y';
+}
+
+export default Math.random() < 0.5 ? x : y;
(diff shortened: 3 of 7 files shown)


src/Declaration.js
```diff
@@ -110,9 +110,15 @@ export class SyntheticDefaultDeclaration {
...
-			return run( this.node.declaration.body, this.statement.scope, this.statement, strongDependencies, false );
+		let declaration = this.node.declaration;
+		while ( declaration.type === 'ParenthesizedExpression' ) declaration = declaration.expression;
+
+		if ( /FunctionExpression/.test( declaration.type ) ) {
+			return run( declaration.body, this.statement.scope, this.statement, strongDependencies, false );
 		}
+
+		// otherwise assume the worst
+		return true;
```
Victorystick: What does this mean exactly? Anything that isn't a `FunctionExpression` is considered to have side-effects?

Rich-Harris: yeah – previously we were assuming that if you called the `foo` in `import foo from './foo.js'`, the default export from `foo.js` was either a reference to another declaration, or a FunctionExpression or ArrowFunctionExpression. But that might not be true:

```js
export default Math.random()  < 0.5 ? x => x * 2 : x => x * 3;
```

We're not *currently* equipped to handle complex cases like that, so we should err on the side of caution. The ParenthesizedExpression stuff (which this was originally designed to address) is really just an optimisation.

Victorystick: Right. So `run` is only called if the `declaration` refers to something that's called as a function?

Rich-Harris: Yeah, exactly. Otherwise we just include all the statement's dependencies by default

Issue #430, opened 2016-01-06
Failure exporting imported sub-modules


brianmhunt:
### Minimal Working Example

_tl;dr_ [A minimal example exhibiting the problem is](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22umd%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%2C%22moduleId%22%3A%22%22%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20*%20as%20maths%20from%20'.%2Fmaths.js'%3B%5Cnexport%20%7Bmaths%7D%22%7D%2C%7B%22name%22%3A%22maths.js%22%2C%22code%22%3A%22import%20*%20as%20algebra%20from%20'.%2Falgebra.js'%5Cnexport%20%7Balgebra%7D%22%7D%2C%7B%22name%22%3A%22algebra.js%22%2C%22code%22%3A%22export%20var%20wink%20%3D%2012%22%7D%5D%7D):
### entry.js

``` javascript
import * as pkg from './package.js'
export {pkg}
```
### package.js

``` javascript
import * as sub from './sub.js'
export {sub}
```
### sub.js

``` javascript
export var wink = 10
```

Running `rollup entry.js` gives:

```
var wink = 10

var pkg = Object.freeze({
    sub: null
});

export { pkg };⏎
```

Where `pkg.sub` should be `Object.freeze({wink: wink})`

Running `rollup package.js` works as expected.

For some more context, see below.

---

In the package [tko.utils](https://github.com/knockout/tko.utils), I have the following code:
### spec/utilsDomBehaviors.js

``` javascript
import * as utils from '../index.js'
// ...
```
### index.js

``` javascript
import './src/bind-shim.js'
export * from './src/array.js'
// ...
import * as virtualElements from './src/dom/virtualElements.js'
import * as domData from './src/dom/data.js'

export {virtualElements, domData}
```

This compiles as-expected when running `rollup index.js` (or the equivalent gulp task, `gulp make`).

However, a problem exhibits when I run, from the project root (or from the `spec` directory, with the `entry` changed accordingly):

``` javascript
> rollup({ entry: 'spec/utilsDomBehaviors.js' }).then((b) => global.bundle = b)
> console.log(bundle.generate({ format: 'es6' }).code)
// ... Code!
```

[...]

PR #432, opened 2016-01-06
Add test re. #430 - include-as-chain

brianmhunt:
When running `mocha` this branch fails with:

```

  487 passing (4s)
  1 failing

  1) rollup function import-chain-as:

      AssertionError: { first: null } deepEqual { first: { value: 42 } }
      + expected - actual

       {
      -  "first": [null]
      +  "first": {
      +    "value": 42
      +  }
       }
```

This PR is purely diagnostic.


brianmhunt:
Note that this test is finnicky.

When I enable the `console.log` output, the tests pass, which would suggest a race condition.

Rich-Harris:
@brianmhunt Many thanks for the test. I've opened #433 which is based off this branch and includes a fix. The problem was that `second.first` wasn't directly referred to anyway in the code, so Rollup didn't realise it needed to create a namespace export block for `first` – solution is to make it realise that the reference to `first` in the `second` namespace block counts just as though it had been written in the original code.

brianmhunt:
Fantastic, thank you kindly for the fix and explanation, @Rich-Harris

I trust there is no reason to leave this PR open, so I will close it out.

I look forward to the PR being merged into a release.

Cheers.

PR #439, opened 2016-01-09
Ensure computed property keys are considered references.

### LINKED ISSUE #434: Fails on computed object literal keys
Given the code `var x = {[foo]: 'bar'};`, rollup seems to be unaware that the variable `foo` is used. It will remove `foo`'s declaration if it is not used elsewhere, or, if `foo` is renamed to avoid conflict with another file, will not rename the reference in the object literal (causing it to refer to the wrong value).

eventualbuddha:
Fixes #434


Rich-Harris:
:+1:

PR #441, opened 2016-01-09
Add a failing test illustrating a renaming bug in ES6 output.

eventualbuddha:
I haven't bothered looking into fixing this yet, but I can if it's not obvious to anyone else.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -127,6 +127,13 @@ export default class Bundle {
 
 		this.externalModules.forEach( module => {
 			module.name = getSafeName( module.name );
+
+			// ensure we don't shadow named external imports, if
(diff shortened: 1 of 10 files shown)


eventualbuddha:
I went back as far as e1d1ce1a1c2733148976bd88defae5c1c1c8b046 to see if this was a recent regression -- it seems it's not, as it fails on that SHA as well.

Rich-Harris:
Fixed, at the cost of needlessly renaming declarations when generating non-ES6 bundles. I don't think that's really a problem though.

eventualbuddha:
That's too bad, but it's better that it works. And the output isn't too terrible. :+1:

Issue #442, opened 2016-01-10
TypeError: declaration.run is not a function


bcherny:
what could cause this?

i'm using:

- gulp-rollup
- rollup-plugin-babel
- rollup-plugin-commonjs
- rollup-plugin-npm

```js
function buildJs (src, dest) {
  return () => gulp
    .src(src)
    .on('error', jsError)
    .pipe(rollup({
      plugins: [
        commonjs({}),
        babel({ runtimeHelpers: true }),
        npm({
          browser: true,
          jsnext: true,
          main: true,
          skip: ['react', 'react-dom', 'react-router']
        })
      ]
    }))
    .pipe(gulp.dest(path.dirname(dest)))
}
```

```
events.js:141
      throw er; // Unhandled 'error' event
      ^
TypeError: declaration.run is not a function
    at Object.walk.enter (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5336:23)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5185:9)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5209:4)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5206:5)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5206:5)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5206:5)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5209:4)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5206:5)
    at visit (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5209:4)
    at walk (/Users/boris/Sites/foo/client/node_modules/rollup/dist/rollup.js:5163:2)
```

Rich-Harris:
Hmm. Are you able to whittle it down to a reproduction of the bug that you can share (i.e. removing all imports that don't cause the bug to appear)? It sounds like the bug is in Rollup itself.

bcherny:
it seems that importing [history](https://github.com/rackt/history) is causing the issue. i use it like this:

```js
import * as createBrowserHistory from 'history/lib/createBrowserHistory'

createBrowserHistory()
// ...
```

Victorystick:
[...]

PR #451, opened 2016-01-18
error if namespace is called

### LINKED ISSUE #446: We should error if calling an imported namespace
[This](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22umd%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%22foo%22%3A%22foo%22%7D%2C%22moduleId%22%3A%22%22%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20*%20as%20foo%20from%20'foo'%3B%5Cnfoo()%3B%22%7D%5D%7D) isn't valid. We should reject it.

```js
import * as foo from 'foo';
foo();
```

Rich-Harris:
This PR fixes #446 (see also #442), whether the namespace is internal or external.

It also adds an `error` helper that makes it very slightly easier to deal with these sorts of errors, by adding properties like `file` and `loc`. We could perhaps make this helper specific to code-related errors and make it do more work, e.g. figuring out the location and including a snippet of the offending code in the error message, with a location marker. In an ideal world it would also handle sourcemap chains, so that it could show the location in the original code rather than transformed code. For now though I wanted to keep this PR focused on #446.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -134,6 +134,7 @@ export class SyntheticDefaultDeclaration {
 
 export class SyntheticNamespaceDeclaration {
 	constructor ( module ) {
+		this.isNamespace = true;
 		this.module = module;
 		this.name = null;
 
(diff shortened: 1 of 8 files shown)


Victorystick:
This is looking great! :+1:

Issue #426, opened 2016-01-05
var exports.foo


Rich-Harris:
[This](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22var%20foo%3B%5Cnfoo%20%3D%20%7B%7D%3B%5Cn%5Cnexport%20%7B%20foo%20%7D%3B%22%7D%5D%7D) clearly shouldn't happen...

Rich-Harris:
Easy way to fix this would be to remove the `var` declaration. But underlying issue is that we're marking `foo` as something that gets reassigned, and therefore needs to be rewritten as `exports.foo` (rather than just declaring it and later doing `export { foo }`), when in actual fact it isn't reassigned; it was never initialised (and even if it was reassigned, it doesn't happen inside a function, which means there's no danger of the value changing after it gets exported).

So there's actually two issues here.

mbostock:
A shot in the dark, but this looks related to #438.

Rich-Harris:
[Fixed](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22var%20foo%3B%5Cnfoo%20%3D%20%7B%7D%3B%5Cn%5Cnexport%20%7B%20foo%20%7D%3B%22%7D%5D%7D) in 0.25.1

PR #452, opened 2016-01-18
fix `var exports.foo`

Rich-Harris:
#426. Turns out there was a semi-related bug involving destructuring syntax, which is also fixed by this PR.


eventualbuddha:
:+1:

PR #450, opened 2016-01-17
follow symlinks

Rich-Harris:
PR for #447. Note that this only works for the default resolver – plugins would have to implement similar logic separately.

PR #456, opened 2016-01-19
Reexport fs functions directly

Victorystick:
#450 is looking good. :+1:

The only eye-sore I see are all the `export const xxx = fs.xxx`s at the bottom of `fs.js`. Let's just re-export them. :smile:

### DIFF
--- a/src/utils/fs.js
+++ b/src/utils/fs.js
@@ -2,12 +2,7 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
 import * as fs from 'fs';
 import { dirname } from './path.js';
 
-export {
-	lstatSync,
-	readdirSync,
-	readFileSync,
-	realpathSync
-} from 'fs';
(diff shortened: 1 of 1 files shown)


eventualbuddha:
Seems okay. Build fails though?

Rich-Harris:
Makes sense! In fact why not `export * from 'fs'`? More future-proof, and easier to keep the browser build up-to-date (don't need to create dummy methods just so they can be imported)

> Seems okay. Build fails though?

Is due to #447 – I don't know how to symlink on windows

Victorystick:
I thought you said `export * from ...` was an anti-pattern? :stuck_out_tongue: I agree that it's the most convenient option though.

Rich-Harris:
> I thought you said `export * from ...` was an anti-pattern?

'A foolish consistency is the hobgoblin of little minds' – Emerson

PR #458, opened 2016-01-20
use an interopDefault function for external CJS deps

### LINKED ISSUE #419: interopRequire
Via https://github.com/caolan/async/commit/85f5be9ccbd1055178d14162916d9757303d468c#commitcomment-15248909 – it would probably make sense to DRY these calls out with something similar to Babel's `interopRequire`

aearly:
Closes #419 .  Also makes async's modular builds a bit smaller. :)

### DIFF
--- a/test/form/external-imports-custom-names/_expected/cjs.js
+++ b/test/form/external-imports-custom-names/_expected/cjs.js
@@ -1,7 +1,8 @@
 'use strict';
 
-var $ = require('jquery');
-$ = 'default' in $ ? $['default'] : $;
+function _interopRequire (id) { var ex = require(id); return 'default' in ex ? ex['default'] : ex; }
+
+var $ = _interopRequire('jquery');
 
 $( function () {
 	$( 'body' ).html( '<h1>hello world!</h1>' );
--- a/test/form/external-imports-custom-names/_expected/cjs.js
+++ b/test/form/external-imports-custom-names/_expected/cjs.js
@@ -1,8 +1,8 @@
 'use strict';
 
-function _interopRequire (id) { var ex = require(id); return 'default' in ex ? ex['default'] : ex; }
+function _interopDefault (ex) { return 'default' in ex ? ex['default'] : ex; }
 
-var $ = _interopRequire('jquery');
+var $ = _interopDefault(require('jquery'));
 
 $( function () {
 	$( 'body' ).html( '<h1>hello world!</h1>' );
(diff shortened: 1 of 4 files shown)


Victorystick:
Good idea. To ensure that we don't get any name collisions, we should probably add `_interopRequire` to the list of _expected global names_ [here](https://github.com/rollup/rollup/blob/master/src/Bundle.js#L66).  That'll ensure that any declarations like `var _interopRequire` will be automatically renamed. We don't want to have our `_interopRequire` function inadvertently overridden.

test/form/external-imports-custom-names/_expected/cjs.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
-var $ = require('jquery');
-$ = 'default' in $ ? $['default'] : $;
+function _interopRequire (id) { var ex = require(id); return 'default' in ex ? ex['default'] : ex; }
+
+var $ = _interopRequire('jquery');
```
eventualbuddha: I think this should be done in two steps like it was before, otherwise using `browserify` or the like on the output won't really work. Unless I'm missing something?

eventualbuddha: That is, there should still be a `require('jquery')` for such tools to find.

eventualbuddha: i.e.

```js
function _interopRequire (mod) { return 'default' in mod ? mod['default'] : mod; }
var $ = _interopRequire(require('jquery'));
```

Victorystick:
I agree with @eventualbuddha.

Also just noticed that the `alphabet` test is broken after these changes. The current behaviour
```js
var alphabet = require('alphabet');
var alphabet__default = 'default' in alphabet ? alphabet['default'] : alphabet;
```
This PR
```js
// alphabet equals alphabet__default, although it shouldn't
var alphabet = _interopRequire('alphabet');
var alphabet__default = alphabet;
```
What we want is to extract the `'default' in alphabet ? alphabet['default'] : alphabet` into a common function, and execute that on the _whole import_.
```js
var alphabet = require('alphabet');
var alphabet__default = _interopRequire(alphabet);
```

test/form/external-imports-custom-names/_expected/cjs.js
aearly: Yeah, you're right -- it'll have to be a two-step process.

aearly:
One other thing -- we can also do something similar for the AMD/UMD/IIFE outputs too by modifying `finalisers/shared/getInteropBlock.js`.  I wanted to get the CJS output working first.

Rich-Harris:
awesome, thanks @aearly!

PR #465, opened 2016-01-22
Skip dead branches of a conditional expression.

eventualbuddha:
This allows replacement of inline conditions such as in `export const resolve = isWindows ? win32Resolve : posixResolve;`.

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -300,7 +300,7 @@ export default class Module {
 		}
 
 		walk( ast, {
-			enter: node => {
+			enter: (node, parent, prop) => {
 				// eliminate dead branches early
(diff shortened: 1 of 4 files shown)


Victorystick:
@eventualbuddha This looks great! Are you reimplementing Node's `path` module in ES2015? I've thinking about what Node's builtin modules would look like if rewritten in ES2015, and how well they could be bundled by Rollup. This is a great step in the right direction! :+1:

eventualbuddha:
@VictoryStick yes, exactly. I'll share once I get this merged and tested out. The only thing I'm not sure of how to handle is how to trigger the inlining of `isWindows`. I'd rather not hardcode that specific case.

PR #472, opened 2016-01-27
Support export { global }

Victorystick:
I encountered an error when attempting to export a global variable, like so:

```js
export { Promise };
```
```
TypeError: Cannot set property 'isExported' of null
  ...
```

Rollup requires each exported binding to have a corresponding Declaration. This PR introduces a SyntheticGlobalDeclaration to circumvent this problem.


Rich-Harris:
nice :+1:

PR #491, opened 2016-02-03
check that _interopDefault's input is a non-null object

### LINKED ISSUE #474: interopDefault fails if imported module is a non-object
If you have a situation like this...

```js
// node_modules/external-dependency/index.js
module.exports = 'a string';

// main.js
import x from 'external-dependency'
```

...the `interopDefault` function searches for `default` inside `'a string'` when the bundle runs. Need to check it's a non-null object.

borisirota:
fixes #474


dariocravero:
It would be great to see this merged :)

Rich-Harris:
Thanks @borisirota!

dariocravero:
@Rich-Harris would you mind releasing that to npm? :) Thanks!

Rich-Harris:
@dariocravero done!

PR #499, opened 2016-02-07
Don't overwrite moduleName object for IIFE and UMD

### LINKED ISSUE #493: Use existing moduleName object for IIFE and UMD if present
Hey!
Right now rollup seems to reset module namespace every time the file is executed as seen in [this example](http://rollupjs.org/#{%22options%22%3A{%22format%22%3A%22iife%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A{}%2C%22moduleId%22%3A%22%22}%2C%22modules%22%3A[{%22name%22%3A%22main.js%22%2C%22code%22%3A%22export%20function%20baz%28%29%20{}%22}]})(`(this.myBundle = {})`). This is problematic when you have multiple packages that should exist in common namespace inside window object, but some of them are lazy loaded when rarely needed. In this situation the variable will be cleared every time dynamic bundle is included.

The solution would be to slightly change the code so the generated piece would look like: `(this.myBundle = this.myBundle || {})`.

krzksz:
Fixes #493. This pull request allows to have multiple bundles that attach variables to common namespace. Without the fix, all of the properties set by first bundle would be erased by a second one.

Having multiple bundles enables better caching control for each file and async loading so I think this should be helpful :smile:


eventualbuddha:
This seems reasonable. Where did the trailing beeline changes come from?

krzksz:
Whoops, no idea, could be the editor linting config. Will investigate and fix asap.

krzksz:
Ok, there are some inconsistencies, but I've fixed pull request to match existing whitespacing.

Rich-Harris:
Thanks @krzksz!

PR #512, opened 2016-02-12
Detect side effect in member expression assignment when not top level

### LINKED ISSUE #476: Function calls with side effects are removed (when they shouldn't be)
Here is a reduced test case:

http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22amd%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%2C%22moduleId%22%3A%22%22%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22const%20foo%20%3D%20%7B%7D%3B%5Cn%5Cnconst%20set%20%3D%20(key%2C%20value)%20%3D%3E%20%7B%5Cn%20%20foo%5Bkey%5D%20%3D%20value%3B%5Cn%7D%3B%5Cn%5Cnset(%5C%22foo%5C%22%2C%201)%3B%5Cnset(%5C%22bar%5C%22%2C%202)%3B%5Cnset(%5C%22qux%5C%22%2C%203)%3B%5Cn%5Cnconsole.log(foo)%3B%22%7D%5D%7D

As you can see, it completely removes the calls to `set`

In my real application, I am using an object to store configuration, and then I am using function calls to populate the object.

But rollup removes all of the function calls, so the configuration ends up being empty.

kzc:
Fixes #476

### DIFF
--- a/src/utils/run.js
+++ b/src/utils/run.js
@@ -95,6 +95,8 @@ export default function run ( node, scope, statement, strongDependencies, force 
 
 				if ( declaration ) {
 					if ( declaration.isParam ) hasSideEffect = true;
+				} else if ( !scope.isTopLevel ) {
+					hasSideEffect = true;
 				} else {
 					declaration = statement.module.trace( subject.name );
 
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Great stuff, thank you @kzc

PR #521, opened 2016-02-17
Plugins can be passed as functions

TrySound:
Don't like plugins with empty parentheses. Think this looks much nicer.

```js
{
  plugins: [
    eslint,
    commonjs({ jsnext: true })
    babel,
    uglify
  ]
}
```

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -19,7 +19,8 @@ import { isRelative, resolve } from './utils/path.js';
 
 export default class Bundle {
 	constructor ( options ) {
-		this.plugins = ensureArray( options.plugins );
+		this.plugins = ensureArray( options.plugins )
+			.map( plugin => typeof plugin === 'function' ? plugin() : plugin );
 
 		this.plugins.forEach( plugin => {
 			if ( plugin.options ) {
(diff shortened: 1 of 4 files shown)


TrySound:
Postcss has plugin api
```
postcss.plugin('plugin-name', function (options) {
  return function (css, result) {  };
});
```

We can do some good stuff.
```
rollup.plugin('plugin-name', function (options) {
  return {
    transform: function () {}
  };
});
```
`rollup.plugin` can wrap function with try-catch and throw beautified error with plugin name. The same it can do with all props (try-catch + catch). This will make error handling less confusing and more similar.

eventualbuddha:
Personally I'd favor conceptual simplicity over slightly less punctuation.

Rich-Harris:
I agree with @eventualbuddha – I've been bitten in the past by trying to be too accommodating with this sort of thing, and while it saves a couple of keystrokes, ultimately it only creates additional surface area for bugs, and confusion and anxiety for people trying to understand how everything fits together.

`rollup.plugin(...)` means that plugins have a dependency on Rollup, which is a really bad place to be – from experience, it creates all kinds of nightmares with versioning etc. And it doesn't actually add anything that we couldn't do inside Rollup itself (i.e. the try-catch stuff, beautified error messages). The only missing piece is the plugin name, which we could do like so:

```js
export default function myPlugin ( options = {} ) {
  return {
    name: 'my-plugin', // if absent, defaults to e.g. 'anonymous plugin'
    transform () {...}
  };
}
```

Victorystick:
I agree. People will understand that plugins are _by convention_ distributed as functions that should be called to be activated.

PR #528, opened 2016-02-23
Make sure shorthand destructuring assignments don't break

### LINKED ISSUE #527: Hygienic rename breaks shorthand destructuring assignments
Consider following code:

index.js:
```javascript
import { useModuleStuff } from './module';
const Stuff = 'localStuff';

useModuleStuff();
```

module.js:
```javascript
const { Stuff } = GlobalNamespace;

export function useModuleStuff() {
  return new Stuff();
}
```

In order to avoid conflicts between local and imported `Stuff` variables, Rollup renames the latter one, which is a good thing, but it doesn't take into account the destructuring semantics and generates following code:

```javascript
const { Stuff$1 } = GlobalNamespace;

function useModuleStuff() {
  return new Stuff$1();
}

useModuleStuff();
```

As you can see, destructuring got the form `const { Stuff$1 } = GlobalNamespace` which is not semantically equivalent to `const Stuff$1 = GlobalNamespace.Stuff` anymore.

Victorystick:
Fixes #527

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -44,7 +44,7 @@ export default class Statement {
 		let readDepth = 0;
 
 		walk( this.node, {
-			enter ( node, parent ) {
+			enter ( node, parent, prop ) {
 				// warn about eval
 				if ( node.type === 'CallExpression' && node.callee.name === 'eval' && !scope.contains( 'eval' ) ) {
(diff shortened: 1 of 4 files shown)


Rich-Harris:
:+1:

Issue #534, opened 2016-02-28
rollup@0.25.2 can't build react-router


TrySound:
Since 0.25.2 I get this error
```
Cannot overwrite the same content twice: 'invariant'
Error: Cannot overwrite the same content twice: 'invariant'
    at Object.patch (D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\node_modules\magic-string\dist\magic-string.es6.js:424:12)
    at Object.overwrite (D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\node_modules\magic-string\dist\magic-string.es6.js:393:8)
    at D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\src\Module.js:516:19
    at Array.forEach (native)
    at D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\src\Module.js:489:25
    at Array.forEach (native)
    at Module.render (D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\src\Module.js:429:19)
    at D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\src\Bundle.js:222:26
    at Array.forEach (native)
    at Bundle.render (D:\_station\nvm\v0.12.7\node_modules\rollup\www\ROLLUP\rollup\src\Bundle.js:221:23)
```

TrySound:
Ref https://github.com/rollup/rollup/issues/502

STAH:
https://github.com/rollup/rollup-plugin-replace/issues/4

TrySound:
My config
```js
import nodeResolve from 'rollup-plugin-node-resolve';
import commonjs from 'rollup-plugin-commonjs';
import replace from 'rollup-plugin-replace';

export default {
  entry: 'src/main.js',
  external: [
    'react',
    'react-dom'
  ],
  plugins: [
    nodeResolve({ jsnext: true }),
    commonjs(),
    replace({
      'process.env.NODE_ENV': JSON.stringify('production')
    }),
  ],
  dest: 'public/main.js',
  format: 'iife',
  sourceMap: true
};
```

Code for reproducing
```js
import { Router, Route, IndexRoute, hashHistory } from 'react-router';

Router();
Route();
IndexRoute();
hashHistory();
```

TrySound:
[...]

PR #535, opened 2016-02-28
Skip nested ternaries from optimization

STAH:
Resolve https://github.com/rollup/rollup/issues/534


eventualbuddha:
This should have tests. Also, isn't this just a workaround and not a real fix?

STAH:
Your optimization in https://github.com/rollup/rollup/commit/1a0011485e912295df28bea534fd69abc823b701 is useful for conditional export you mentioned. I'm not sure that is the use case for complex ternary expressions: they should be optimized by uglify plugin (if necessary).

eventualbuddha:
Since #534 is resolved, I think we no longer need this.

PR #542, opened 2016-03-02
Allow 'exports' key

alalonde:
Apparently this isn't used much, but we need it whitelisted for our gulp-rollup plugin integration.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -76,7 +76,7 @@ describe( 'rollup', function () {
 			return rollup.rollup({ entry: 'x', plUgins: [] }).then( function () {
 				throw new Error( 'Missing expected error' );
 			}, function ( err ) {
-				assert.equal( 'Unexpected key \'plUgins\' found, expected one of: banner, dest, entry, external, footer, format, globals, indent, intro, moduleId, moduleName, onwarn, outro, plugins, sourceMap', err.message );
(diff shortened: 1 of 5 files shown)


eventualbuddha:
Is there a test that could illustrate why this change is needed?

alalonde:
Well, we are using the gulp plugin, which passes the same `options` literal to `rollup.rollup()` AND `bundle.generate()` (https://github.com/mcasimir/gulp-rollup/blob/master/index.js#L34).  On your API wiki page it lists all of the options for BOTH of these functions, but only a subset of these option keys are whitelisted.  I imagine it is pretty common for people to consume your API in this manner, so not adding that key seems unnecessarily restrictive.

Rich-Harris:
Seems reasonable, and more consistent – I think the main reason to have a whitelist is to prevent typos etc, so if it makes life easier then :+1:. Thanks!

test/test.js
```diff
@@ -149,6 +149,7 @@ describe( 'rollup', function () {
 
 						// try to generate output
 						try {
+							if(config.bundleOptions) { console.log(config.bundleOptions); }
```
ayosec: Is this `config.log` used to debug code?

alalonde: Yeah, you can take it out.  Forgot to before merging.

Issue #552, opened 2016-03-11
browserify interoperablity


calvinmetcalf:
one of the issues I've had with rollup is I can't use all the previous modules I've written with it due to interoperability issues, so after talking to @nolanlawson about it I decided to write some stuff to make browserify modules 'just work'.
- [x] need a plugin for [node globals](https://github.com/calvinmetcalf/rollup-plugin-node-globals) like process and Buffer
- [x] need a plugin for [node built in modules which browserify supplies](https://github.com/calvinmetcalf/rollup-plugin-node-builtins)
- [x] need a plugin to [automatically apply browserify transforms](https://github.com/calvinmetcalf/rollup-plugin-auto-transform) from the package.json
- [x] need to [clean up some incompatibility](https://github.com/rollup/rollup-plugin-commonjs/pull/49) due to the commonjs plugin also handling the `global` global.
- [ ] need to trouble shoot why [crypto won't build](https://github.com/calvinmetcalf/rollup-plugin-node-builtins/issues/3)
- [ ] ability to replicate commonjs execution order (for circular deps)
- [ ] make a plugin/preset that combines node-globals, node-builtins, auto-transform, commonjs, node-resolve, and json with sensible defaults

callumlocke:
I know the list isn't all done yet, but would you mind posting your current best browserify-compatible plugins config?

calvinmetcalf:
```
import commonjs from 'rollup-plugin-commonjs';
import nodeResolve from 'rollup-plugin-node-resolve';
import globals from 'rollup-plugin-node-globals';
import builtins from 'rollup-plugin-node-builtins';
import json from 'rollup-plugin-json';
import autoTransform from 'rollup-plugin-auto-transform';
rollup({
  entry: 'main.js',
  plugins: [
        autoTransform(),
        globals(),
        builtins(),
        nodeResolve({ jsnext: true, main: true, browser: true }),
        commonjs({
          ignoreGlobal: true
        }),
        json()
]
})
```

also you'll want to use some specific versions of rollup modules due to the pulls stalling specifically 

- my nowrap branch of the commonjs module (calvinmetcalf/rollup-plugin-commonjs#nowrap) which has a more nuanced check for commonjs modules avoiding false positives and avoids wrapping the entire module in a function so that parts of it can get tree-shaken (without that they will always be side effects)  
[...]

PR #555, opened 2016-03-15
omit side effects in unreachable modules (#551)

calvinmetcalf:
the idea is that if you don't actually import anything from a module (i.e. because everything that imported it has been shaken out of the tree) then don't include it's side effects.

This would be very helpful for situations where you have something like a lodash or other utility thing that people are going to only one one or 2 functions out of.  So if you want `_.pluck` it shouldn't matter to you that `_.somethignWeird` depends on `giantModuleWithSideEffects`.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -92,6 +92,9 @@ export default class Bundle {
 					declaration.use();
 				});
 
+				// mark modules that should appear in the bundle
+				entryModule.use();
+
 				// mark statements that should appear in the bundle
 				let settled = false;
 				while ( !settled ) {
--- a/src/ExternalModule.js
+++ b/src/ExternalModule.js
@@ -25,7 +25,9 @@ export default class ExternalModule {
 			this.name = name;
 		}
 	}
-
+	use () {
+			// noop
+	}
 	traceExport ( name ) {
 		if ( name !== 'default' && name !== '*' ) {
 			this.exportsNames = true;
--- a/src/Module.js
+++ b/src/Module.js
@@ -59,6 +59,7 @@ export default class Module {
 		this.analyse();
 
 		this.strongDependencies = [];
+		this.used = false;
 	}
 
 	addExport ( statement ) {
@@ -598,6 +599,8 @@ export default class Module {
 	run () {
 		let marked = false;
 
+		if (!this.used) return marked;
+
 		this.statements.forEach( statement => {
 			marked = statement.run( this.strongDependencies ) || marked;
 		});
@@ -610,7 +613,7 @@ export default class Module {
 		if ( name in this.imports ) {
 			const importDeclaration = this.imports[ name ];
 			const otherModule = importDeclaration.module;
-
+			otherModule.use();
 			if ( importDeclaration.name === '*' && !otherModule.isExternal ) {
 				return otherModule.namespace();
 			}
@@ -657,5 +660,9 @@ export default class Module {
 
 			if ( declaration ) return declaration;
 		}
+	}
+
+	use () {
+		this.used = true;
 	}
 }
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -125,7 +125,7 @@ export default class Statement {
 	mark () {
 		if ( this.isIncluded ) return; // prevent infinite loops
 		this.isIncluded = true;
-
+		this.module.use();
 		this.references.forEach( reference => {
 			if ( reference.declaration ) reference.declaration.use();
 		});
--- a/test/function/side-effects-in-unreached-modules/_config.js
+++ b/test/function/side-effects-in-unreached-modules/_config.js
@@ -1,0 +1,3 @@
+module.exports = {
+	description: 'side effects in unreached modules are not executed'
+};
--- a/test/function/side-effects-in-unreached-modules/circle.js
+++ b/test/function/side-effects-in-unreached-modules/circle.js
@@ -1,0 +1,3 @@
+export function circle(r) {
+  return Math.PI * r * r;
+}
--- a/test/function/side-effects-in-unreached-modules/cube.js
+++ b/test/function/side-effects-in-unreached-modules/cube.js
@@ -1,0 +1,7 @@
+import square2 from './square2';
+
+function cubic(x) {
+  return cubic.square(x) * x;
+}
+export var cube = cubic;
+cube.square = square2;
--- a/test/function/side-effects-in-unreached-modules/main.js
+++ b/test/function/side-effects-in-unreached-modules/main.js
@@ -1,0 +1,3 @@
+import {cube} from './maths.js';
+
+console.log(cube(3));
--- a/test/function/side-effects-in-unreached-modules/maths.js
+++ b/test/function/side-effects-in-unreached-modules/maths.js
@@ -1,0 +1,5 @@
+export {cube} from './cube';
+
+export {square} from './square';
+
+export {circle} from './circle';
--- a/test/function/side-effects-in-unreached-modules/square.js
+++ b/test/function/side-effects-in-unreached-modules/square.js
@@ -1,0 +1,9 @@
+export function square(x) {
+  return x * x;
+}
+
+Object.defineProperty(square, "test", {
+  get: function() { throw new Error('should not import square'); },
+  set: function( arg ) {}
+});
+square.test();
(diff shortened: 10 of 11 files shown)


eventualbuddha:
This seems reasonable to me. Thanks!

Rich-Harris:
Thanks @calvinmetcalf. I'm not sure if this complies with the spec though – wouldn't a native module loader execute that code? Totally understand the motivation for omitting it but it's important to adhere to the proper semantics

calvinmetcalf:
a bundler and a loader ~~are~~ should never going to have exactly the same semantics because they are trying to do slightly different things.  Take browserify which statically resolves all dependencies without execution which does deviate from how node module loader works but in a way that makes more packages work most of the time.

Having a pedantic mode or no tree shaking mode might be a helpful addition for the minority of cases where it needs to act exactly like the module spec.

The practical backround was trying to get `import {randomBytes}  from 'crypto'` to work which ended up bringing in almost all of the dependencies, I'd guess in large part due to the inherits pattern which likely is incorrectly flagged as a side effect.

wearhere:
Since @calvinmetcalf mentioned this in #552, what's the status of this?

I agree with "a bundler and a loader should never have exactly the same semantics because they are trying to do slightly different things" fwiw. It seems very much in line with the vision of tree shaking that you could shake side effects out along with the module that defined them i.e. if all imports had been eliminated.

Notionally wouldn't the side effects be relevant only to the module that did the side effects, anyway? So if the entry point didn't end up using anything from the module, then it wouldn't care about the side effects either.

Perhaps this wouldn't be true all the time but in that case there could be a "pedantic or no tree shaking mode" like @calvinmetcalf suggests.

calvinmetcalf:
here are some real life pieces of code this hit this issue, this code from the browserify buffer shim causes the buffer code and all of it's deps from never being treeshaken out problematic line is highlighted


```js
if (Buffer.TYPED_ARRAY_SUPPORT) {
  Buffer.prototype.__proto__ = Uint8Array.prototype
  Buffer.__proto__ = Uint8Array
  if (typeof Symbol !== 'undefined' && Symbol.species &&
      Buffer[Symbol.species] === Buffer) {
    // Fix subarray() in ES2016. See: https://github.com/feross/buffer/pull/97
//VVVVVVVVV
    Object.defineProperty(Buffer, Symbol.species, {
      value: null,
      configurable: true
    })
//^^^^^^^^^^^
  }
}
```

the above one is relatively straightforward why it's causing the issue rollup is being conservative about the object assign even though buffer otherwise would be tree shaken out

```js
Buffer.TYPED_ARRAY_SUPPORT = global.TYPED_ARRAY_SUPPORT !== undefined
  ? global.TYPED_ARRAY_SUPPORT
  : typedArraySupport()
function typedArraySupport () {
  try {
    var arr = new Uint8Array(1)
    arr.__proto__ = {__proto__: Uint8Array.prototype, foo: function () { return 42 }}
// VVVVVVVV
    return arr.foo() === 42 && // typed array instances can be augmented
        typeof arr.subarray === 'function' && // chrome 9-10 lack `subarray`
        arr.subarray(1, 1).byteLength === 0 // ie10 has broken `subarray`
// ^^^^^^^^^^^^^
  } catch (e) {
    return false
  }
}
```

why the above code causes a tree shaking issue is beyond me, then from a dep of buffer

```js
function init () {
  var code = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
  for (var i = 0, len = code.length; i < len; ++i) {
    lookup[i] = code[i]
    revLookup[code.charCodeAt(i)] = i
  }

  revLookup['-'.charCodeAt(0)] = 62
  revLookup['_'.charCodeAt(0)] = 63
}
// VVVVVVVV
init()
//^^^^^^
``` 
taking it out of the function and putting it at the top level doesn't fix it either only differing the init call

Rich-Harris:
Tree-shaking that code away would involve some pretty sophisticated static analysis. My hope with #902 is that we'll be able to make some progress in that direction. It's too early to make promises, but it'd be a useful exercise to think through what kind of smarts Rollup would need to handle each of these cases. So:

```js
Object.defineProperty(Buffer, Symbol.species, {
  value: null,
  configurable: true
})
```

Rollup knows that, for example, `Object.keys(Buffer)` could be removed, because it knows that `Object.keys` [is a pure function](https://github.com/rollup/rollup/blob/master/src/utils/pureFunctions.js#L17). Because `Object.defineProperty` isn't in that list, it assumes the worst. If it understood that O.dP` *only* affects the first argument, then it could remove that statement because it knows that `Buffer` isn't used. That seems possible.

```js
Buffer.TYPED_ARRAY_SUPPORT = global.TYPED_ARRAY_SUPPORT !== undefined
  ? global.TYPED_ARRAY_SUPPORT
  : typedArraySupport()
function typedArraySupport () {
  try {
    var arr = new Uint8Array(1)
    arr.__proto__ = {__proto__: Uint8Array.prototype, foo: function () { return 42 }}
// VVVVVVVV
    return arr.foo() === 42 && // typed array instances can be augmented
        typeof arr.subarray === 'function' && // chrome 9-10 lack `subarray`
        arr.subarray(1, 1).byteLength === 0 // ie10 has broken `subarray`
// ^^^^^^^^^^^^^
  } catch (e) {
    return false
  }
}
```

As of #902, Rollup will know that `arr` could only have one value – `new Uint8Array(1)`. In order to know that `arr.subarray(...)` didn't have side-effects, it would need to understand prototype chains *and* know that `Uint8Array.prototype.subarray` could either be a side-effect-free method that returns another Uint8Array *or* `undefined`. In order to know that `arr.foo()` didn't have side-effects, it'd need to be smarter still – it would need to understand that you can override the prototype chain. That feels a little bit hairy (what if the `arr.__proto__` assignment happened inside a function? Could it still trust its knowledge of `Uint8Array.prototype`? Tricky.)

```js
function init () {
  var code = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
  for (var i = 0, len = code.length; i < len; ++i) {
    lookup[i] = code[i]
    revLookup[code.charCodeAt(i)] = i
  }

  revLookup['-'.charCodeAt(0)] = 62
  revLookup['_'.charCodeAt(0)] = 63
}
// VVVVVVVV
init()
//^^^^^^
```

We know that `code` is a string, so we can know that `code.charCodeAt(...)` is a side-effect-free function that returns a number. With that dealt with, Rollup should already recognise that it can disregard assignments to `lookup` and `revLookup` *unless* they're used elsewhere in the bundle.

So in summary, some of these things can be massaged away, some might be a little bit harder. I think smarter static analysis is the right direction to go in (rather than brute-forcing Rollup into ignoring certain modules, etc – we tried that once and it was a disaster) but we'll need to try and see how far we get.

calvinmetcalf:
A shorter term solution might be the ability to annotate your code to tell
rollup it is side effect free similar to how Istanbul or eslint allow you

GGAlanSmithee:
Another option would be to add a `pureFunctions` config, which would be a list of functions that you, as a user, consider to be pure. This could then be concatenated with the one in [pureFunctions.js](https://github.com/rollup/rollup/blob/master/src/utils/pureFunctions.js#L17) before parsing?

calvinmetcalf:
@GGAlanSmithee that wouldn't necessarily be helpful as what we need is a way of saying that a function only modifies the certain arguments which would help with `Object.defineProperty` and other similar cases, but that would only fix the first example, the third one would be fixed by this pull as everything else was removed from the file, the second example is a bit more complex as it's, from the perspective of rollup, a user defined function which would be harder to name.

your solution would work for own code but it would also be tricky for dependencies in context this is coming up in my work trying to get the node standard library shims to work with rollup so it wouldn't be code directly used in your own app.

GGAlanSmithee:
@calvinmetcalf Ah, I see. Mabye that could help with `Object.assign` too then, which is also considered to be impure. (because it invokes getters/setters?)

calvinmetcalf:
@Rich-Harris your idea for handling functions that only modify the first thing could work with 

'Object.defineProperty',
	'Object.defineProperties',
	'Object.assign',
	'Object.freeze',
	'Object.preventExtensions',
	'Object.seal',
	'Object.setPrototypeOf'

and fix a whole lot of subtle side effect bugs

bfred-it:
Can this be rescued or should it be closed?

calvinmetcalf:
this has been bit rotted something fierce so I'll close it

PR #560, opened 2016-03-17
Cannot read property 'node' of undefined when declaration is SyntheticGlobalDeclaration

QiV:
Could not catch this in mocha tests. it seems like to crash only at compile time.

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -189,6 +189,7 @@ export default class Module {
 			const declaration = this.declarations[ name ];
 			const statement = declaration.statement;
 
+			if ( !statement ) return;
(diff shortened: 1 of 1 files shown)


TrySound:
@QiV Can you produce test case?

TrySound:
Do you mean compile rollup itself? Or when rollup compile something?

QiV:
Sorry for delay.
Rollup crashes when i try to compile my code, which exports global vars (window, document, location,  etc.)
I`ll try to produce this case at new repo a little bit later.

QiV:
git clone git@github.com:QiV/test-rollup.git && npm i && npm test

QiV:
Is it reproducable? Anything else I can do?

PR #561, opened 2016-03-19
use async versions of pulgin functions by default

calvinmetcalf:
the docs suggest plugins should use async methods, the internal code should probably follow it's own advice.

### DIFF
--- a/src/utils/defaults.js
+++ b/src/utils/defaults.js
@@ -1,18 +1,25 @@
-import { isFile, readFileSync } from './fs.js';
+import { isFile, readFile } from './fs.js';
 import { dirname, isAbsolute, resolve } from './path.js';
 import { blank } from './object.js';
 
 export function load ( id ) {
-	return readFileSync( id, 'utf-8' );
+	return readFile( id, 'utf-8' );
 }
 
 function addJsExtensionIfNecessary ( file ) {
-	if ( isFile( file ) ) return file;
-
-	file += '.js';
-	if ( isFile( file ) ) return file;
-
-	return null;
+	return isFile( file )
+		.then( answer => {
+			if ( answer ) {
+				return file;
+			}
+			file += '.js';
+			return isFile(file).then( answer => {
+				if ( answer ) {
+					return file;
+				}
+				return null;
+			});
+		});
 }
 
 export function resolveId ( importee, importer ) {
--- a/src/utils/fs.js
+++ b/src/utils/fs.js
@@ -2,30 +2,9 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
 import * as fs from 'fs';
 import { dirname } from './path.js';
 
-function mkdirpath ( path ) {
-	const dir = dirname( path );
-	try {
-		fs.readdirSync( dir );
-	} catch ( err ) {
-		mkdirpath( dir );
-		fs.mkdirSync( dir );
-	}
-}
-
-export function isFile ( file ) {
-	try {
-		const stats = fs.statSync( file );
-		return stats.isFile();
-	} catch ( err ) {
-		return false;
-	}
-}
-
-export function writeFile ( dest, data ) {
+function readdir ( dir ) {
 	return new Promise( ( fulfil, reject ) => {
-		mkdirpath( dest );
-
-		fs.writeFile( dest, data, err => {
+		fs.readdir( dir, ( err ) => {
 			if ( err ) {
 				reject( err );
 			} else {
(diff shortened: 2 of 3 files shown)


src/utils/fs.js
```diff
@@ -2,38 +2,76 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
+			}
+		});
+	});
 }
 
 export function writeFile ( dest, data ) {
-	return new Promise( ( fulfil, reject ) => {
-		mkdirpath( dest );
+	return mkdirpath(dest).then(()=>
+		new Promise( ( fulfil, reject ) => {
```
TrySound: You should return promise.

src/utils/defaults.js
```diff
@@ -1,18 +1,13 @@
-import { isFile, readFileSync } from './fs.js';
+import { isFile, readFile } from './fs.js';
 import { dirname, isAbsolute, resolve } from './path.js';
 import { blank } from './object.js';
 
-export function load ( id ) {
-	return readFileSync( id, 'utf-8' );
```
TrySound: You still should specify 'utf-8'. Just replace with promise version, not delete.

src/utils/fs.js
```diff
@@ -2,38 +2,76 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
+				} else {
+					fulfil();
+				}
+			});
+		})
+	);
+}
+export function readFile ( id ) {
+	return new Promise( ( fulfil, reject ) => {
+		fs.readFile( id, 'utf-8', ( err, data ) => {
```
TrySound: Do not apply encoding in this file.

src/utils/fs.js
calvinmetcalf: Woops

src/utils/defaults.js
calvinmetcalf: I specified it in the other definition but I can switch it to do so over here

src/utils/fs.js
calvinmetcalf: wait I am returning the promise, due to the lack of { braces }, I can switch it to use curly braces though

TrySound: I think better explicit return here.

calvinmetcalf:
this is failing on travis and I'm not sure why

src/utils/fs.js
```diff
@@ -2,38 +2,82 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
-		return stats.isFile();
-	} catch ( err ) {
-		return false;
-	}
+	return new Promise( ( fulfil, reject ) => {
+		fs.stat( file, ( err, stats ) => {
+			if ( err ) {
+				reject( err );
+			} else if ( stats.isFile() ) {
+				fulfil( file );
```
eventualbuddha: Are we going with the British or American spelling here?

src/utils/fs.js
```diff
@@ -2,38 +2,82 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
-	try {
-		const stats = fs.statSync( file );
-		return stats.isFile();
-	} catch ( err ) {
-		return false;
-	}
+	return new Promise( ( fulfil, reject ) => {
+		fs.stat( file, ( err, stats ) => {
+			if ( err ) {
+				reject( err );
```
eventualbuddha: This API doesn't feel correct to me. I typically write async functions using the same API as the synchronous ones, but with async semantics. If this were synchronous (i.e. what's on master now), it would return `true` or `false` to answer the question and throw an exception in some other weird case. It would not return the filename on success and throw otherwise. I see that this is just to make `addJsExtensionIfNecessary`'s job easier, but I don't think that's a valid reason to have this API.

src/utils/defaults.js
```diff
@@ -1,18 +1,15 @@
...
 
 function addJsExtensionIfNecessary ( file ) {
-	if ( isFile( file ) ) return file;
-
-	file += '.js';
-	if ( isFile( file ) ) return file;
-
-	return null;
+	return isFile( file )
+	  .catch( () => isFile( file + '.js' ) )
```
eventualbuddha: Weird API aside, this indentation is incorrect.

src/utils/fs.js
```diff
@@ -2,38 +2,82 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
 	const dir = dirname( path );
-	try {
-		fs.readdirSync( dir );
-	} catch ( err ) {
-		mkdirpath( dir );
-		fs.mkdirSync( dir );
-	}
+	return readdir( dir ).catch( ()=>{
+		return mkdirpath( dir ).then( ()=>{
+			return mkdir( dir ).catch( e=>{
```
eventualbuddha: Need spaces around `=>`.

src/utils/fs.js
calvinmetcalf: updated

src/utils/fs.js
calvinmetcalf: I am copying what was already there on the page

src/utils/fs.js
```diff
@@ -2,38 +2,80 @@ import Promise from 'es6-promise/lib/es6-promise/promise.js';
...
 				reject( err );
 			} else {
-				fulfil();
+				fulfil( data );
 			}
 		});
 	});
 }
-
 export const readdirSync = fs.readdirSync;
```
TrySound: We still need this?

TrySound:
@calvinmetcalf You should also update browser patch
https://github.com/rollup/rollup/blob/master/browser/fs.js

src/utils/fs.js
calvinmetcalf: I don't think you ever need it, will remove

TrySound:
Great! Now squash commits please.

calvinmetcalf:
and done

Rich-Harris:
Thanks for the PR @calvinmetcalf. I'm not 100% sure if the advice in favour of async methods is valid in this context – my understanding, from when I've looked into this before, is that async is only faster if you're doing a bunch of fs operations in parallel. Rollup processes modules serially, so I *believe* that sync methods are faster. (Can anyone confirm or deny, or at least clarify?) The async alternatives involve more code and complexity so the benefits would have to be tangible and unambiguous.

calvinmetcalf:
I'm pretty sure it's doing some stuff in parallel because there were some
race conditions that showed up.

But yes your right the async stuff tends towards higher latency.

Rich-Harris:
hey – am going to close this PR as it's not clear that there are benefits that justify the extra code, and I'm having a bit of a spring clean. Thanks anyway!

PR #562, opened 2016-03-20
Fix edge case for non-exported variable declaration aliases for exported globals.

eventualbuddha:
Closes #560.


eventualbuddha:
I'm not convinced this is the best way to do it. @Rich-Harris?

PR #572, opened 2016-03-24
Allow external files to be non-existent, relative files. Fixes #532.

### LINKED ISSUE #532: Feature for treating non-existent, relative files as external files
https://github.com/rollup/rollup/issues/410 and https://github.com/rollup/rollup/issues/418 (as explained in [the wiki](https://github.com/rollup/rollup/wiki/JavaScript-API#external)) make it possible to treat relative files as external by putting `path.resolve('./relative-dep.js')` into the `externals` array of your rollup configuration. However, as far as I can tell, that approach for skipping relative files will only work if the file exists (otherwise `path.resolve` will fail inside of the rollup configuration).

The use case for skipping a relative import without there being a file of the same name is for bundling projects that will later be `System.import`ed (using SystemJS). The reason why the import does not match the relative filename exactly is because the project specifies which SystemJS plugin to use when loading the relative import. For example, my file tree may look like this:
```
src
-->main.js
-->main.css
```

where main.js is
```
import './main.css!cssmodules';
...
```

Is there a way I am unware of that allows for you to skip `./main.css!cssmodules` when the filename is really just `./main.css`??

joeldenning:
See #532 for a detailed explanation of the problem this pull requests solves, and also the use case for non-existent, relative files to be treated as external dependencies.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -178,8 +178,15 @@ export default class Bundle {
 		return mapSequence( module.sources, source => {
 			return this.resolveId( source, module.id )
 				.then( resolvedId => {
-					// If the `resolvedId` is supposed to be external, make it so.
-					const forcedExternal = resolvedId && ~this.external.indexOf( resolvedId.replace( /[\/\\]/g, '/' ) );
+					let externalName;
+					if ( resolvedId ) {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
LGTM, thanks!

PR #580, opened 2016-03-29
Add an option to add noConflict method to umd global

evs-chris:
I don't know if this was even something on the radar for rollup, but it seemed like a better solution than regexp hacking build output. So feel free to ignore this if it's not something rollup should handle :smile:

This _should_ only affect UMD modules when they're used as a global. I'm not 100% sure this is the best way to handle it, or if I've even put all the right things in the right places, but I did do a test build with this with Ractive and it seems to work as intended. 

Linking - ractivejs/ractive#2066


Rich-Harris:
Great idea – `noConflict` is used in a bunch of libraries so I think it's totally valid to have it as an option here. Cheers @evs-chris 🍻

PR #583, opened 2016-03-31
Add useStrict option to ALLOWED_KEYS.

olegskl:
Because it is [documented](https://github.com/rollup/rollup/wiki/JavaScript-API#usestrict) but not allowed.


Rich-Harris:
thank you!

PR #591, opened 2016-04-04
Normalize relative external paths.

fatfisz:
I've encountered a problem in such a scenario:

* I had a path to a module declared as an external (so that the file would be imported later, not by `rollup`)
* a file that was not in the same directory as the entry file was importing that module (`../../module`)

It turned out that the imported module in the bundle was still `../../module`, while I expected it to be normalized with respect to, let's say, the entry file.

I investigated a bit and found out something scary: if a relative module is imported from many files in a nested structure, there will be more than one import in the resulting bundle.

For example, I get this:
```js
var foo = _interopDefault(require('./foo'));
var foo$1 = _interopDefault(require('../foo'));
var foo$2 = _interopDefault(require('../../foo'));

var first = foo$1 + foo$2;
var main = foo + first;
module.exports = main;
```

instead of this:
```js
var foo = _interopDefault(require('./foo.js'));

var first = foo + foo;
var main = foo + first;
module.exports = main;
```

This PR fixes that problem by normalizing the paths with respect to the resolved entry directory.

--
I probably did some terrible mistakes because I don't fully understand what's happening there. So please be gentle :wink:


src/Bundle.js
```diff
@@ -28,6 +28,7 @@ export default class Bundle {
 		});
 
 		this.entry = unixizePath( options.entry );
+		this.entryId = null;
```
Victorystick: You don't need to create this variable. Use `this.entryModule.id` instead.

test/function/relative-external-include-once-nested/_config.js
```diff
@@ -0,0 +1,18 @@
...
+	context: {
+		require: function ( required ) {
+			assert.equal( required, './first/foo.js' );
+			return 1;
+		}
+	},
+	exports: function ( exports ) {
+		assert.equal( exports, 3 );
+	}
+};
```
Victorystick: All files should have a trailing new line.

src/Bundle.js
fatfisz: I tried to, but `this.entryModule` is `null` at this point. This is because it is still being fetched (and it won't be completed until all its dependencies are fetched too).

test/function/relative-external-include-once-nested/_config.js
fatfisz: I noticed that some of the test files don't have the trailing new line (for example the `consistent-renaming` tests) and even went through some pains to disable auto line adding in my editor because of that :smile: Will fix this in a jiffy.

src/Bundle.js
Victorystick: That's too bad. I guess it's fine then.

fatfisz:
I added another test case, thought of it today:
```
|--first
|  |-b.js
|  \-config.js
|-a.js
\-config.js

Let's say that for some reason we want both `config.js` files to be external. Then if `a.js` references `config.js` and `first/b.js` references `first/config.js`, only one of the `config.js` files will be required in the bundle (the other one will be completely ignored).

Of course my patch fixes this too.

@victorystick Is there anything still holding us back from merging this? Can I help somehow?

Victorystick:
@fatfisz I think it's looking good. /cc @rollup/collaborators

Rich-Harris:
LGTM! Thanks @fatfisz, sorry for the wait – going to merge a few other PRs then cut a new release shortly 👍

PR #603, opened 2016-04-09
Update dependencies.

eventualbuddha:
I left all the major-revision changes out because I figured they may require more work.


TrySound:
@eventualbuddha Non-semver are major too. :)

eventualbuddha:
@TrySound heh, true. But people generally follow semver even in the 0.x range.

Issue #533, opened 2016-02-27
src/utils/path.js behavior on Windows


Conduitry:
Is there an architectural reason for src/utils/path.js to re-implement so many things that already exist in the node path module? At least for the non-browser build, it seems it would make sense to use node's path functions.

Node's path module also includes support for Windows paths. Today on Windows I noticed an issue where Rollup's resolved version of the path to a file contained forward slashes while one that I had generated myself in a plugin (using node's path module) contained backslashes - and this resulted in Rollup thinking these were two separate modules and it was included twice in the bundle.

It seems there are three options:

- Have users make sure paths to modules sent to Rollup always use forward slashes, regardless of the platform's real path separator (this is what I'm doing for now but it seems like a bit of a hack)
- Update the function `resolve()` in path.js. Right now the line `return resolvedParts.join( '/' ); // TODO windows...` is always assuming a `/` separator
- Do a larger-scale replacement of path.js implementations with the functions in node's path module

Options 1 is very easy because it's pretty much "do nothing". But option 2 doesn't seem very difficult at all. We could check whether the global `process` is defined and whether `process.platform === 'win32'` and if so, use the separator `\` and otherwise use `/`. This would keep the same behavior as exists now in browser environments, where `process` is not available.

Thoughts?

TrySound:
Windows supports forward slash as well. By the way we use minimatch for filters which supports only unix separator.

Conduitry:
I'm trying to figure out again what my initial issue was. I think it was that I had a module that was getting included twice, if one time it was referred to with forward slashes in the path and another time with backslashes. I had a module that I was both a) importing normally, using imports with relative paths in my code, and b) importing as the result of generated code from a transform. The former had forward slashes in its path, and the latter (generated using path.join) had backslashes. My solution at the time was to change the backslashes in the generated path to forward slashes. I haven't really revisited whether more recent rollup versions changed what I had to do here.

PR #627, opened 2016-04-26
Follow symlinks when outputting sourcemap's list of files

IvanSanchez:
This resolves the locations of files in the sourcemap. Only useful if symlinked files are being rolled up.

In other words: the sourcemaps of my gobble-rollup workflow now correctly display `src/foo/foo.js` instead of `../../.gobble/09-merge/foo/foo.js`.

Haven't tried this in Windows.

(On a related note: Y U NO [PATH](https://nodejs.org/api/path.html)?)

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -303,10 +303,9 @@ export default class Bundle {
 
 			map.sources = map.sources.map( unixizePath );
 
-			const entryDir = dirname( this.entry );
 			const destDir = dirname( file );
 			map.sources.forEach( ( sourceFile, i) => {
-				map.sources[i] = relative( entryDir, realpath( resolve( destDir, sourceFile) ) );
+				map.sources[i] = relative( destDir, realpath( resolve( destDir, sourceFile) ) );
 			});
 		}
 
--- a/test/sourcemaps/symlinks/_config.js
+++ b/test/sourcemaps/symlinks/_config.js
@@ -4,6 +4,8 @@ module.exports = {
 	description: 'sourcemap resolves symlinks to the real path name',
 
 	test: function (code, map) {
-		assert.deepEqual( map.sources, [ 'real.js', 'main.js' ]);
+		// Source is in symlinks/, rolledup code and sourcemaps will be in symlinks/_actual,
+		// hence the '..'.
+		assert.deepEqual( map.sources, [ '../real.js', '../main.js' ]);
 	}
 };


Rich-Harris:
Thanks @IvanSanchez – could you add a test to this PR please? It's the sort of thing that could very easily be inadvertently removed otherwise!

Re path, there is a related issue for this: https://github.com/rollup/rollup/issues/533. The motivation was in-browser usage, though there are other ways to achieve browser-friendliness that we should probably adopt instead.

IvanSanchez:
I guess I can try making a test or two :-)

IvanSanchez:
OK, so adding the test (and moving the logic out of `bundle.write()` and into `bundle.render()`) was easy enough.

However, I've just broken *every other* unit test which involves sourcemaps, like so:

```diff
   + expected - actual
   "sources": [
-    "foo.js"
-    "bar.js"
-    "main.js"
+    "../foo.js"
+    "../bar.js"
+    "../main.js"
   ]
```

In order for the symlinks to be resolved, they kinda need to be relative to the `entry` directory, not to the directory where the sourcepath is supposed to end up in (`dest`).

This has the (desireable?) side effect of removing any `../` or `src/` or `../src/` paths from sourcemap sources, which just look bad in some web browsers (firefox shows them, chrome collapses them).

I've had a look at the sourcemap v3 standard proposal, and looked into the sourcemap devs email archives, but haven't found anything specifically addressing this issue.

Thoughts?

Rich-Harris:
I can't remember off the top of my head whether paths are supposed to be relative to the `.js` file or the `.map` file (probably the `.js` file, since sourcemaps can be inline), but since they're almost always siblings it's the same result anyway: the paths *definitely* shouldn't be relative to the entry file, which to all intents and purposes is arbitrary.

They do look wonky in devtools, but that's because `../` relative to `/` is ignored – if you have e.g.

```html
<script src='node_modules/my-lib/dist/index.js'></script>
```

...then the sources pane will show the correct original sources in `node_modules/my-lib/src` (or whatever), in my experience. It's only when you move `index.js` to the root (breaking the link between generated code and source code) that it falls apart – but that's on the developer to fix, it's not something a tool can realistically plan for.

IvanSanchez:
I don't agree 100% with your point of view, but kinda agree.

> It's only when you move `index.js` to the root (breaking the link between generated code and source code) that it falls apart

Then I think we have a problem with the gobble magic bits of sorcery applied when serving things through localhost:4567 - the original location will point to `localhost:4567/../../..`  :-/

bfred-it:
Can this be rescued or should it be closed?

guybedford:
If this is no longer progressing, marking for closing due to inactivity.

Issue #590, opened 2016-04-04
Duplication caused from importing the same module with different casing


nhunzaker:
**TL:DR:** What do you think about warning when code imports the same module with different casing?

When a module is mistakenly imported with different casing, on file systems that don't care, duplication occurs when importing the module. I've made a test case here:

https://gist.github.com/nhunzaker/2a36fa377a41f6d6b054196af299d909

In this case, both `green` and `orange` require `yellow`, however one requires it as `Yellow` (with a capital "Y"). Under the hood, Rollup appends `$1` to the second yellow, like so:

``` javascript
var blue = "blue"

var yellow = "yellow"

var green = [ yellow, blue ]

var red = "red"

var yellow$1 = "yellow"

var orange = [ yellow$1, red ]

export { green, orange };
```

What about warning these cases?

Rich-Harris:
My feeling is that Rollup should fail altogether if incorrect casing is used. Implemented in #630

Rich-Harris:
Closing, as Rollup now insists on the correct casing

nhunzaker:
👍. Awesome. Even better!

PR #630, opened 2016-04-28
BREAKING: ensure import paths are case-sensitive

Rich-Harris:
Addresses #590. The suggestion there was that Rollup should warn if a module is imported twice with different casing, but to be honest I can't imagine a situation where you wouldn't want it to just fail altogether.


guybedford:
One thing to think about is that Es modules as implemented in browsers will be case sensitive as far as I am aware. I believe importing the same module with different cases will execute twice in browsers.

So strictly, this is implementing nonstandard behavior even though it makes the most sense. Perhaps @domenic can weigh in here as well.

domenic:
> I believe importing the same module with different cases will execute twice in browsers.

Not unless your web server serves the same body for `GET /foo.js` and `GET /FOO.js`. Behavior there generally varies across servers, but e.g. GitHub.com 404s.

If you want to emulate strict web servers here, using `fs.realPath` and making sure the imported case matches the real path might be the way to go.

tomByrer:
Hmmm, I've heard of this mixed case problem [before](https://github.com/ractivejs/ractive/issues/527). ;)

Rich-Harris:
Went ahead with the current implementation rather than `realpath` just because I expect that would cause some breaking behaviour with linked modules. Should be the same as regards case sensitivity. Can always revisit in future though. Thanks everyone

Issue #650, opened 2016-05-12
Export __esModule when bundling to CJS or UMD in named mode.


cameron-martin:
When the `format` option is set to `'umd'` or `'cjs'` and the `exports` option is set to `'named'`, we should automatically export the `__esModule` property (with value `true`).

This means that a library compiled using rollup can be imported using an app which uses ES6 syntax for imports, then compiles to CommonJS using babel.

Can anyone see any downsides to doing this? If not, I'd be happy to submit a pull request.

eventualbuddha:
We should do that. I've been meaning to for a long time.

Rich-Harris:
This is enabled in 0.26.7 – thanks

ericf:
I've started receiving some issues people ran into with because of this change. They were incorrectly importing the default export of React Intl which does not have a default export, then accessing a named export off of it.

- https://github.com/yahoo/react-intl/issues/499
- https://github.com/yahoo/react-intl/issues/503

Obviously the user's `import` statements are incorrect, but I just wanted to point this out since @cameron-martin asked about potential downsides.

PR #652, opened 2016-05-14
Exported __esModule when bundling to CJS or UMD in named mode.

cameron-martin:
This is a follow-up PR for issue https://github.com/rollup/rollup/issues/650.

One thing that I'm not sure about is if I have broken sourcemaps. Since I have just added to the end of the output, the mapping should still be valid. But can somebody who is more knowledgeable confirm that I have not?

Also, should I add new test cases to explicitly test this behaviour?

EDIT: I see that [magic-string] takes care of keeping source maps correct.

[magic-string]: https://github.com/Rich-Harris/magic-string

### DIFF
--- a/src/finalisers/cjs.js
+++ b/src/finalisers/cjs.js
@@ -1,4 +1,5 @@
 import getExportBlock from './shared/getExportBlock.js';
+import esModuleExport from './shared/esModuleExport.js';
 
 export default function cjs ( bundle, magicString, { exportMode }, options ) {
 	let intro = options.useStrict === false ? `` : `'use strict';\n\n`;
@@ -35,7 +36,7 @@ export default function cjs ( bundle, magicString, { exportMode }, options ) {
 	if ( exportBlock ) magicString.append( '\n\n' + exportBlock );
 
 	if (exportMode === 'named') {
-		magicString.append('\n\nexports.__esModule = true;\n\n');
+		magicString.append(esModuleExport);
 	}
 
 	return magicString;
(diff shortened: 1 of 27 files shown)


cameron-martin:
I noticed that babel [inserts this as the first thing in the file][babel], even before doing any requires. I assume this is to make circular dependencies work correctly.

In that case, this needs a minor rethink, potentially moving from [this][pr], to this:

```javascript

(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ?
	(Object.defineProperty(exports, "__esModule", { value: true }), factory(exports, require('external'))) :
	typeof define === 'function' && define.amd ? define(['exports', 'external'], factory) :
	(factory((global.myBundle = global.myBundle || {}),global.external));
}(this, function (exports,external) { 'use strict';

	class Foo extends external.Component {
		constructor () {
			super();
			this.isFoo = true;
		}
	}

	class Bar extends external.Component {
		constructor () {
			super();
			this.isBar = true;
		}
	}

	class Baz extends external.Component {
		constructor () {
			super();
			this.isBaz = true;
		}
	}

	const foo = new Foo();
	const bar = new Bar();
	const baz = new Baz();

	exports.foo = foo;
	exports.bar = bar;
	exports.baz = baz;

}));
```

[babel]: https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0%2Cstage-2&code=import%20c%20from%20'foo'%3B%0A%0Aexport%20const%20a%20%3D%20'foo'%3B%0Aexport%20const%20b%20%3D%20'foo'%3B%0A%0Aexport%20default%20'foo'%3B
[pr]: https://github.com/cameron-martin/rollup/blob/a660b896dcbca62bbc0ee950aa4a9ebeef9460b9/test/form/dedupes-external-imports/_expected/umd.js

guybedford:
Can we also add this in the AMD case? AMD also respects the __esModule flag in SystemJS, which would fix these builds in jspm too.

joeldenning:
+1 for adding this to the AMD case.

Rich-Harris:
Thanks all – have added support for AMD and the cyclical case with CJS. Had to do it in a separate branch, so there's a separate PR – #685

PR #658, opened 2016-05-21
WIP: Add incremental support build by passing previous bundle

TrySound:
Just a suggestion to make some movement in this very necessary feature.

ref #191


TrySound:
bundle object (in this case only `modules` property) should be stringifiable to let us write it on disk for faster cold start.

TrySound:
However I'm not sure about holding ast. Do we really need this? Sourcemaps holding will be very tricky too.

yaycmyk:
@TrySound I think the right question is what is the slowest thing to rebuild? That should be the thing to tackle to make incremental rebuilds do-able. If sourcemaps can be cheaply and quickly regenerated, then there's no sense trying to chase them (if that's the case, if not then nvm.)

TrySound:
@yaycmyk It's just realization questions. My proposing is only preventing extra transformations. Nothing smarter. It's useful with transpillers, but not with modules themselves. They are still will be parsed and processed,

Rich-Harris:
Finally getting round to digging into this. First thoughts: bravo, I was expecting this to be a lot more complex, but I think you've hit upon the correct approach. I was envisaging keeping the existing `bundle` object and perhaps having a `bundle.invalidate(id)` method (followed by another `bundle.build()`, perhaps), which would allow us to keep existing `Module` objects. That would *ultimately* yield better results (since creating modules isn't free), but would also be really, really hairy. (For example, in some cases plugins would have to know about invalidations etc.)

Regenerating everything from scratch but bypassing transformations is a much more practical approach, in my view. The possibility of saving stuff to disk for fast cold builds is a great benefit too.

We should definitely keep ASTs and sourcemap chains, I think. Acorn is blazing fast but it's still worth avoiding the overhead of re-parsing (we annotate ASTs with `_scope` information, but since those properties are non-enumerable we can still serialize them), and we definitely don't want to lose sourcemaps. (Perhaps we can collapse sourcemap chains into a single step where applicable, to make that process quicker on re-runs. I guess that would entail something like `bundle.generateCacheObject()` which we'd want for saving stuff to disk anyway.)

About to head out, but look forward to investigating more thoroughly – am excited about this 🍻

Rich-Harris:
Shit. Just realised there's a wrinkle in this plan which might turn out to be a dealbreaker. Some plugin functions have side-effects – e.g. rollup-plugin-babel's `transform` function keeps track of which helpers are used so that they can be injected once into the top of the bundle, and rollup-plugin-commonjs tracks whether it needs to inject the `__commonjs` function. And so on.

Without running the `transform` functions for every module (because the module is cached), these side-effects won't happen and the bundle intro will be broken.

It would be nice if we could require plugin functions to be pure, but I'm not sure how this could possibly work.

I see two possible routes:

* Require plugins that have side-effects to implement `save()` and `restore()` (or whatever) functions, which would capture the relevant state. Without this, bundles that use side-effectful plugins will become incorrect – in practical terms, it would be impossible to use `node_modules`.
* Go with the more ambitious `bundle.invalidate(id)` route. For completeness, plugins would need a corresponding `invalidate(id)` function, but if that function was missing the worst that would happen in most cases is that helpers would be injected unnecessarily. This would be less of an ecosystem-breaking change, but would be a lot harder initially.

Any thoughts?

tivac:
Dunno how helpful it is, but browserify (w/ watchify) invalidates specific files from the cache, but doesn't re-transform the files that weren't invalidated. In my experience injecting dependencies is handled by dynamically adding another module via an injected `require()` statement, so that just gets walked as part of the dependency tree and won't be duplicated.

It seems like if the `__commonjs` wrapper was handled as an injected dependency instead of a raw function it would help solve that class of problem. Wish I was more familar w/ rollup internals so I could offer more help :disappointed:

Rich-Harris:
Actually @tivac that might be the perfect answer. Initially I thought that'd be no good because you wouldn't know at the time the helper module was imported which helpers (in the Babel case, at least) were going to be used, but then I remembered...



...and we have tree-shaking, so we can just include the whole lot! So maybe we *can* require functions to be pure after all. Thanks!

TrySound:
@Rich-Harris Will you continue this PR?

tivac:
@Rich-Harris I think you should still provide a way for plugins to know when a cached file has been invalidated. For instance, in [my rollup plugin](https://github.com/tivac/modular-css/blob/master/src/rollup.js) I'd need to know so I could remove the version cached by the `processor` instance since I need to parse CSS files for includes and create a whole separate dependency tree.

[Here's](https://github.com/tivac/modular-css/blob/master/src/browserify.js#L123-L127) the relevant watchify usage, it's pretty simple but so far that seems to have been enough for my use case at least.

Rich-Harris:
@TrySound yep, I think we're on the right track here.

@tivac agree, that would be useful. If we're going down this stateless route, then we can't invalidate, per se, but a pair of `save()` and `restore()` functions that allow plugins to store information (presumably indexed against module IDs) between runs should enable the same thing, right? Or does that make things more complicated?

I've updated rollup-plugin-babel so that it's effectively stateless. Am in the process of doing the same to rollup-plugin-commonjs but I've realised that there's another issue here – if you have a magic ID like `__babelHelpers__` or `__commonjsHelpers__`, it's prone to get picked up by another plugin (e.g. node-resolve) at which point things break. Could get around it by allowing IDs to be non-strings (i.e. each plugin has `const MAGIC_ID = {}`, which only it can recognise), but plugins with `resolveId` functions would need to be aware of that possibility and bug out on `typeof id !== 'string'`. Unless anyone has better ideas?

tivac:
@Rich-Harris I'm not sure what state you expect a plugin to use `save()`/`restore()` to manage, but I don't know much about the guts of `rollup-plugin-babel` or `rollup-plugin-commonjs` so hopefully I'm just missing something obvious.

For my intended use-case w/ [`modular-css`](https://npmjs.com/modular-css) I just want to tell my backing object "pretend you never saw this file because it changed". Then the next time it gets seen it'll be re-parsed as though it was brand-new. I don't need to save anything off or restore anything in particular, the cost of re-parsing the changed files is pretty small in the grand scheme of things.

Slightly related to this issue is being able to inject files into the `watchify`-esque dependency graph to have them monitored for changes. Since CSS files in `modular-css` can include references to other CSS files it ends up making a shadow dependency graph. I'm able to inject those dependencies into `browserify` via https://github.com/tivac/modular-css/blob/master/src/browserify.js#L49-L54 but I don't think `rollup` has a similar idea atm.

Rich-Harris:
@tivac `save` and `restore` would be used for bundle-wide state that is generated during `transform`. It's not entirely thought through, and may in fact not be necessary. The big idea in this PR is that if the code that gets loaded for a particular module ID matches what our cache says the original code was previously, then rather than re-transforming that module, we can use the existing cached result.

So if you had a module that imported `foo.css`, `bar.css` and `baz.css`, modular-css would transform all three. If you then modified `foo.css`, the entire process would begin again (new plugin instance, new `Processor`, etc) *but* Rollup would recognise that `bar.css` and `baz.css` hadn't changed from what it loaded the previous time, so it would only run `foo.css` through `transform`. So as long as the transformation of `bar.css` and `baz.css` doesn't cause any side-effects (i.e. mutating the state of the `Processor` instance), then there'd be nothing to save or restore. (I couldn't work out from glancing at the code whether `Processor` is stateful in this way or not?)

Excellent point re shadow dependency graph, hadn't thought of that. I reckon the best solution would probably be to add a method to the context of `transform` – e.g. `this.addDependencies(file.deps)`. Those would get added to the relevant entry in `bundle.modules`, and it would be the responsibility of the watcher (e.g. `rollup --watch`) to watch those files in addition to the modules that Rollup is directly aware of.

tivac:
Ah, ok. I'm coming to this w/ a lot of `browserify` baggage so I apologize.

Unfortunately the way it is now adding files to the `Processor` instance is super-duper stateful, so I don't know that the current proposal would help `modular-css`. That's mostly due to the secondary/shadow dependency graph created by CSS files composing stuff from each other and the need to output every referenced CSS file as one combined CSS file at the end of the build ([code here](https://github.com/tivac/modular-css/blob/master/src/rollup.js#L63-L84)).

I could add support for a `save`/`restore` style dehydration/rehydration of a `Processor` instance but that seems a trickier than letting it live as long as the `rollup --watch` process and tracking when files change by removing them & re-processing them as they're discovered on the AST traversal. `Processor` instances don't re-process files they've already seen, so it's pretty fast to deal with already-known files.

Rich-Harris:
Ah, I see. In that case yeah, I don't see a great way to make this work with modular-css. In fact, I wonder if we'd need to have plugins explicitly state whether their transforms were cacheable or not.

I guess the next question is now many plugins are similarly stateful, and whether or not that affects the direction we go in.

tivac:
I'll keep tabs on this/other issues around it and can probably figure out something. Managing a separate dependency graph like that wouldn't be my first choice but it's the only easy way to have compatibility w/ all the different tools out there.

Victorystick:
@Rich-Harris The TypeScript plugin has always exploited Rollup's tree-shaking abilities to only include the helpers that are required. So it's also stateless.

Issue #633, opened 2016-04-28
Externals normalization


guybedford:
With the new normalization changes on 0.26, I'm getting a couple of bugs with externals normalization.

main.js
```javascript
import {fn} from './dep.js';
fn();
```

dep.js
```javascript
export function fn(){...}
```

The Rollup build is then done with:

```javascript
    return rollup.rollup({
      entry: 'main.js',
      external: ['dep.js'],
      plugins: [{    
        resolveId: function(id, importer, options) {
          if (id == './dep.js')
            return 'dep.js';
          return id;
        },
        load: function(id, options) {
          return sources[id];
        }
      }],
      onwarn: function(message) {}
    })
```

The output file then contains:

```javascript
import { fn } from '../dep.js';
fn();
```

Where the `../dep.js` has been added here incorrectly.

Rich-Harris:
Apologies for the long wait... this is fixed in 0.26.4

guybedford:
Thanks so much, SystemJS Builder finally works on 0.26.4!!

PR #661, opened 2016-05-22
ensure that external alias are not treated as relative path

### LINKED ISSUE #660: Alias & external dependencies
Hi,
Suppose this build configuration (I use gulp-rollup, but it does not matter):

```javascript
gulp.task('bundle', () => {
  return gulp.src(path.join(__dirname, 'src/index.js'))
    .pipe(rollup({
      external: ['underscore'],
      plugins: [
        alias({
          lodash: 'underscore'
        })
      ]
    }))
    .pipe(gulp.dest(path.join(__dirname, 'dist')));
});
```

What I want to do:
- Replace `lodash` imports with `underscore` imports.
- Treat underscore as an external dependency.

Here is the following output:

```javascript
import _ from '../../../../../underscore';
```

It seems that the alias is replace with a relative path (please, check out [this repository](https://github.com/mjeanroy/test-rollup-alias) to test by yourself).

I tried to debug what happens in rollup and it seems that [this line](https://github.com/rollup/rollup/blob/master/src/Bundle.js#L209) is responsible. Not sure if it could be a fix, but maybe rollup can check if the `resolvedId` value is a relative path ?

What do you think about this patch:

```diff
- } else if ( resolvedId ) {
+ } else if ( resolvedId && resolvedId.charAt(0) === '.' ) {
    normalizedExternal = this.getPathRelativeToEntryDirname( resolvedId );
}
```

If that sounds good to you, I can try to send you a PR.

Not sure, but it may be related to #633.

mjeanroy:
Should close #660 (and maybe #633).

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -15,7 +15,7 @@ import transformBundle from './utils/transformBundle.js';
 import collapseSourcemaps from './utils/collapseSourcemaps.js';
 import SOURCEMAPPING_URL from './utils/sourceMappingURL.js';
 import callIfFunction from './utils/callIfFunction.js';
-import { dirname, isRelative, relative, resolve } from './utils/path.js';
+import { dirname, isRelative, isAbsolute, relative, resolve } from './utils/path.js';
 
 export default class Bundle {
 	constructor ( options ) {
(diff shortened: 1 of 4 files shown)


Rich-Harris:
Many thanks – it does indeed fix #633 as well. Will push out a release with the fix shortly

PR #671, opened 2016-06-01
Load 'index.js' from the supplied module directory if it exists and the module doesn't

tomrw:
If you have a file structure similar to the following, and attempt to import `./first` from `main`, then a `"Could not resolve ./first"` error will be thrown by Rollup. In node (and Browserify!), this will resolve to `./first/index.js` if the file exists.

```
src
├──main.js
└──first
	└──index.js
```

### DIFF
--- a/src/utils/defaults.js
+++ b/src/utils/defaults.js
@@ -7,10 +7,17 @@ export function load ( id ) {
 }
 
 function addJsExtensionIfNecessary ( file ) {
+	let originalFileName;
+
 	if ( isFile( file ) ) return file;
+
+	originalFileName = file;
 
 	file += '.js';
 	if ( isFile( file ) ) return file;
+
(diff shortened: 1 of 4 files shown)


Victorystick:
Since the behavior of implicitly loading paths like `./first/index.js` when requesting `./first` is Node specific behavior we don't include it in Rollup's default `resolveId` function. Due to the [performance issues](https://kev.inburke.com/kevin/node-require-is-dog-slow/) of Node's module resolution algorithm (made worse by the use of implicit extensions and the `/index.js` behavior) we that you _always_ specify the extension of imported modules, and avoid the `/index.js` anti-pattern.

If a Rollup user wants compliance with Node's resolution algorithm, we recommend that they use [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve).

PR #676, opened 2016-06-04
feat: support tree-shaking of classes instantiated at the top level of a module

IgorMinar:
Only classes without constructors are currently supported because we can't
easily detect side-effect free constructors without a bigger refactoring.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -13,6 +13,12 @@ export default class Declaration {
 			} else if ( node.type === 'VariableDeclarator' && node.init && /FunctionExpression/.test( node.init.type ) ) {
 				this.isFunctionDeclaration = true;
 				this.functionNode = node.init;
+			} else if ( node.type === 'ClassDeclaration') {
+				this.isClassDeclaration = true;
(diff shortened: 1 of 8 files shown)


IgorMinar:
I'm not sure why node 0.12 is not happy. I can take a look tomorrow.

src/Declaration.js
```diff
@@ -48,8 +54,10 @@ export default class Declaration {
 	run ( strongDependencies ) {
 		if ( this.tested ) return this.hasSideEffects;
 
-
-		if ( !this.functionNode ) {
+		if ( this.isClassDeclaration && !this.hasConstructor ) {
+			// class declaration without a constructor has no side-effects
+			this.hasSideEffects = false;
```
Victorystick: A second condition for this to be true is for no _super class constructor_ had side-effects, as a constructor-less class still implicitly invokes the constructor of its super class.

IgorMinar: good point

eventualbuddha: It's there already something else that will look for reassignment of exported bindings, or should we do that here?

bfred-it:
I assume this was replaced by #677

Issue #669, opened 2016-05-30
unnecessary variables while import


QiV:
input file

``` javascript
import {d} from './test';
console.log(d);
```

test.js:

``` javascript
var d = [1,2,3];
var h = noop || d[0];
export { d }
function noop(){
  return 1;
}
```

result of the `rollup index.js -f cjs` code:

``` javascript
'use strict';
var d = [1,2,3];
var h = noop || d[0];
function noop(){
  return 1;
}
```

Maybe this is the expected behavior, but I guess, that there is no need for `h` and `noop`

QiV:
`git clone -b issues/669 git@github.com:QiV/test-rollup.git && cd test-rollup && npm test`

eventualbuddha:
Interesting. It seems to have something to do with the variable declaration, since changing `var h = loop || d[0];` to `var h = d[0];` also forces that statement to remain, even though `d[0]` by itself doesn't.

[repl](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20%7Bd%7D%20from%20'.%2Ftest'%3B%5Cnconsole.log(d)%3B%22%7D%2C%7B%22name%22%3A%22test.js%22%2C%22code%22%3A%22var%20d%20%3D%20%5B1%2C2%2C3%5D%3B%5Cnvar%20h%20%3D%20noop%20%7C%7C%20d%5B0%5D%3B%5Cnexport%20%7B%20d%20%7D%5Cnfunction%20noop()%7B%5Cn%20%20return%201%3B%5Cn%7D%22%7D%5D%7D)

Rich-Harris:
Have had a poke at this, and I think it's a slightly hairy issue that might entail a bit of a rethink about how certain things are handled internally. Before I get into that, there *is* a workaround – if you separate the variable declaration from the initialisation...

```js
// this...
var h; h = noop || d[0];

// not this...
var h = noop || d[0];
```

...then things work as expected.

Rollup has a concept of *aliases*, which ensure that in a situation like this...

```js
function Foo () {}
var foo_proto = Foo.prototype;

foo_proto.method = function () {...}
```

[...]

PR #681, opened 2016-06-05
WIP: better alias handling

Rich-Harris:
See #669. Just a pair of failing tests for now.


Rich-Harris:
Fixed in #902, closing

PR #685, opened 2016-06-05
__esModule for AMD modules and CJS with cyclical dependencies

Rich-Harris:
This is based on #652, but applies it to AMD modules (see https://github.com/rollup/rollup/pull/652#issuecomment-219398441), and moves the block to the top of a CJS file to handle cyclical dependencies.

It also makes the style consistent with other statements inserted by Rollup (single line, single quotes).

Unfortunately I don't think it's possible to handle the cyclical dependency case (https://github.com/rollup/rollup/pull/652#issuecomment-219315359) in AMD modules. Because of that, I think it makes sense to keep the `__esModule` block at the bottom of UMD modules, because augmenting `exports` before any `require` statements take place would yield inconsistent behaviour between AMD and CJS environments. In practical terms, cyclical dependencies between bundles are rather rare, so I don't think this is a major issue.

Before we merge this, we need to be aware that this will break support for IE8. Not that IE8 should guide our decisions, but something to be aware of – for my understanding, what is the actual issue that `__esModule` solves, and is solving it a higher priority than preserving IE8 support? I have to confess I'm not entirely clear on why we need it – its absence has never been an issue for me. (cc @guybedford)

### DIFF
--- a/src/finalisers/amd.js
+++ b/src/finalisers/amd.js
@@ -1,6 +1,7 @@
 import { getName, quoteId } from '../utils/map-helpers.js';
 import getInteropBlock from './shared/getInteropBlock.js';
 import getExportBlock from './shared/getExportBlock.js';
+import esModuleExport from './shared/esModuleExport.js';
 
 export default function amd ( bundle, magicString, { exportMode, indentString }, options ) {
 	let deps = bundle.externalModules.map( quoteId );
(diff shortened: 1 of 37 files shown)


eventualbuddha:
> what is the actual issue that `__esModule` solves, and is solving it a higher priority than preserving IE8 support?

`__esModule` enables us to distinguish a default export from a property named `default`. I believe this is why babel 6 switched to a more strict interop mode. I would consider it a higher priority than IE8 support, especially since you can always use a polyfill to support it.

eventualbuddha:
This seems good to me!

Rich-Harris:
> I would consider it a higher priority than IE8 support, especially since you can always use a polyfill to support it.

fair enough! I suppose we always have the option of writing `exports.__esModule = true` if we needed to add an IE8 safe mode.

developit:
@Rich-Harris Is there an option to turn this off?

developit:
For now I'm using a rather hacky way to remove this: `uglifyjs --pure-funcs=Object.defineProperty`. 
Maybe when `strictMode=false` it'd be fair to skip adding `__esModule`?

PR #688, opened 2016-06-05
Allow specifying multiple output targets.

### LINKED ISSUE #655: Support multiple configurations in a single rollup.config.js
With the current way config files work, creating different bundles from a single code base involves creating a lot of config files with a lot of duplicated configuration in them. I can somehow work around the duplication by importing a base config and overriding properties as needed, but I still end up with all these files and `import base from './rollup.base.js';` at the top of each.

It would be much nicer if I could put all my configurations into a single file, e.g. `targets:  [ {entry: 'src/index.js', format: 'amd', dest: 'dist/amd.js'}, ...]`

eventualbuddha:
Closes #655.

### DIFF
--- a/bin/runRollup.js
+++ b/bin/runRollup.js
@@ -143,6 +143,16 @@ function bundle ( options ) {
 			return bundle.write( options );
 		}
 
+		if ( options.targets ) {
+			var result = null;
+
+			options.targets.forEach( function ( target ) {
+				result = bundle.write(target);
+			});
+
+			return result;
+		}
+
 		if ( options.sourceMap && options.sourceMap !== 'inline' ) {
 			handleError({ code: 'MISSING_OUTPUT_OPTION' });
 		}
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -27,6 +27,7 @@ const ALLOWED_KEYS = [
 	'plugins',
 	'preferConst',
 	'sourceMap',
+	'targets',
 	'treeshake',
 	'useStrict'
 ];
--- a/test/cli/multiple-targets/_config.js
+++ b/test/cli/multiple-targets/_config.js
@@ -1,0 +1,4 @@
+module.exports = {
+	description: 'generates multiple output files when multiple targets are specified',
+	command: 'rollup -c'
+};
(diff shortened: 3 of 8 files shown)


bin/runRollup.js
```diff
@@ -143,6 +143,16 @@ function bundle ( options ) {
...
 		}
 
+		if ( options.targets ) {
+			var result = null;
+
+			options.targets.forEach( target =>
+				result = bundle.write( target )
+			);
+
+			return result;
```
eventualbuddha: This really should be done like so:

```js
return Promise.all(
  options.targets.map( target => bundle.write( target ) )
);
```

However, since this is in `bin/` and the `es6-promise` library is bundled and not available as a node dependency, I can't pull it in. What do you think we should do, @Rich-Harris?

bin/runRollup.js
```diff
@@ -143,6 +143,16 @@ function bundle ( options ) {
 			return bundle.write( options );
 		}
 
+		if ( options.targets ) {
```
eventualbuddha: Should we be doing more configuration validation here? Somewhere else?

bin/runRollup.js
```diff
@@ -143,6 +143,16 @@ function bundle ( options ) {
...
+			var result = null;
+
+			options.targets.forEach( function ( target ) {
+				result = bundle.write(target);
+			});
+
+			return result;
```
eventualbuddha: This really should be done like so:

```js
return Promise.all(
  options.targets.map( function ( target ) { return bundle.write( target ); } )
);
```

However, since this is in `bin/` and the `es6-promise` library is bundled and not available as a node dependency, I can't pull it in. What do you think we should do, @Rich-Harris?

Rich-Harris: Come to think of it I think we can safely remove es6-promise. Node 0.12 and above support it, and I couldn't give a crap about 0.10 or IE (it's easy to polyfill if someone really wants it). Let's just use native `Promise.all` and I'll raise a separate issue about removing es6-promise

bin/runRollup.js
Rich-Harris: I'd go with YAGNI – we can add more validation if it later turns out we need it

PR #695, opened 2016-06-07
undo non-string ID change

Rich-Harris:
Reverts #692 (albeit manually, because I want to keep the `Map` stuff). On further reflection it doesn't really make any sense, because the *input* to `resolveId` is always a string. So this doesn't actually mean we can guarantee no collisions. Not entirely sure what I was thinking.

Alternative proposal: those plugins could use an ID that's an invalid filename, such as NUL followed by one of the many characters that are forbidden in Windows filenames. `createFilter` in rollup-pluginutils could automatically exclude these, meaning things like rollup-plugin-node-resolve would never bother with them.


eventualbuddha:
Seems good. I was going to ask why there previous change was needed.

PR #696, opened 2016-06-07
ensure name suggestions are legal identifiers

Rich-Harris:
This fixes a bug exposed by https://github.com/rollup/rollup-plugin-babel/issues/55 – it's possible for a module to export a binding that is a reserved word, and for another module to import that reserved word, without ever writing a syntax error...

```js
// helpers.js
var _typeof = x => typeof x;
export { _typeof as typeof };

// main.js
import * as helpers from './helpers.js';
console.log( helpers.typeof( 1 ) );
```

(The export from `helpers.js` might be against spec, I'm not sure. That would be a separate issue if it is.)

This PR ensures that the use of `helpers.typeof` doesn't cause the `var _typeof` to be rewritten as `var typeof`.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -1,4 +1,5 @@
 import { blank, forOwn, keys } from './utils/object.js';
+import makeLegalIdentifier from './utils/makeLegalIdentifier.js';
 import run from './utils/run.js';
 import { SyntheticReference } from './Reference.js';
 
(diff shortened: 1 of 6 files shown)


eventualbuddha:
Looks good!

Victorystick:
@Rich-Harris @eventualbuddha I'm pretty sure the only export name that mustn't be a valid variable name is `default`.

PR #698, opened 2016-06-07
Incremental rebuilds

Rich-Harris:
This builds on #658 (and is a PR to merge this into that branch) – a few small tweaks:

* I think the property should be `cache` rather than `bundle` – it's a lot more explicit about what it does, and I think it'll make more sense when we get round to serializing the cache object to disk for fast cold builds
* ASTs and sourcemap chains are preserved – this will allow us to skip parsing in many cases, and should mean sourcemaps are unaffected
* The tests were a little brittle for me – because they relied on resetting the state of the input files, any changes that resulted in tests failing would also cause the state of the test itself to get borked up. So now modules are served from memory rather than the filesystem when testing – makes life a bit easier

If @rollup/collaborators are on board I think we should merge this and #658 and release it, making it clear that this is still experimental. Once that's done, we can crack on with the following:

* [ ] Implementing `rollup --watch`. I think there should be a separate `rollup-watch` package that should handle the file watching, because in my experience file watchers are all horrible and it would be better not to pollute Rollup itself. If someone tries to do `rollup -w` without having installed `rollup-watch` (whether locally or globally, ideally locally) we can do that for them after a prompt. That's the best of both worlds in my view
* [ ] Handling stateful plugins (such as modular-css – see https://github.com/rollup/rollup/pull/658#issuecomment-224120273). I've done a quick survey of existing plugins, and stateful ones are definitely in the minority (I could only find modular-css). I think the best solution here (while promoting statelessness as far as possible) is to add a `this.disableCache()` method to the `transform` context to signal that a given file should *always* be transformed, cache or no cache
* [ ] Handle shadow dependency graphs – e.g. if a CSS file as `@import` statements that aren't converted to JavaScript `import` statements, we need a mechanism (e.g. `this.addDependency(id)`) for keeping track so that the file watcher can trigger rebuilds when those files change
* [ ] Serialization. This isn't *quite* as straightforward as jumping a JSON blob to `.rollup-cache.json` or whatever, because we need to invalidate the cache if the config changes (for example). Would be great to have though

...plus probably some other stuff.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -19,8 +19,8 @@ import { dirname, isRelative, isAbsolute, relative, resolve } from './utils/path
 
 export default class Bundle {
 	constructor ( options ) {
-		if ( typeof options.bundle === 'object' ) {
-			this.cachedModules = options.bundle.modules.reduce((modules, module) => {
+		if ( typeof options.cache === 'object' ) {
+			this.cachedModules = options.cache.modules.reduce((modules, module) => {
 				modules[module.id] = module;
 				return modules;
 			}, {});
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -19,11 +19,11 @@ import { dirname, isRelative, isAbsolute, relative, resolve } from './utils/path
 
 export default class Bundle {
 	constructor ( options ) {
-		if ( typeof options.cache === 'object' ) {
-			this.cachedModules = options.cache.modules.reduce((modules, module) => {
-				modules[module.id] = module;
-				return modules;
-			}, {});
+		this.cachedModules = new Map();
+		if ( options.cache ) {
+			options.cache.modules.forEach( module => {
+				this.cachedModules.set( module.id, module );
+			});
 		}
 
 		this.plugins = ensureArray( options.plugins );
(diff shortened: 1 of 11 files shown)


tivac:
Sorry I'm causing problems :disappointed:

TrySound:
Wow, awesome!

Rich-Harris:
@tivac far from it! You're helping us avoid implementing this from inside an ivory tower, and your suggestion about how to handle helper modules was 💯 . I think it's important we can find a way to accommodate all different types of plugin.

TrySound:
Finally tried rollup in my work with big project. Works perfectly. There is not sourcemap problem anymore. Wooh.

tivac:
@Rich-Harris I was able to solve most of the shadow dep graph issues by modifying the rollup plugin to actually inject the CSS dependency graph into the JS via the `transform` (code [here](https://github.com/tivac/modular-css/blob/bd178c41c5456aa55f6c3d0f901b544daa73c798/src/rollup.js#L42-L44)).

Now the only thing I need is some way for my plugin to be notified when a file has been changed, so it can dump it's cache for that file. I could ditch the caching aspect but I'd really rather not as the CSS processing is a non-trivial amount of work that I'd prefer to not duplicate on each build.

I do that in the [browserify plugin](https://github.com/tivac/modular-css/blob/bd178c41c5456aa55f6c3d0f901b544daa73c798/src/browserify.js#L123-L127) by listening to the `update` event that watchify fires whenever a file is changed.

PR #699, opened 2016-06-08
rollup --watch

Rich-Harris:
This is very basic and unfinished, but it appears to work – if you have [rollup-watch](https://github.com/rollup/rollup-watch) installed, you can run `rollup --watch` to watch all the files in your bundle and rebuild it (incrementally) when those files change.

rollup-watch is very rough – no logging or error handling etc. Just wanted to throw something together so we could start kicking the tyres.

All feedback welcome!


bin/runRollup.js
```diff
@@ -127,7 +128,20 @@ function execute ( options, command ) {
 	});
 
 	try {
-		bundle( options ).catch( handleError );
+		if ( command.watch ) {
+			let watch;
```
TrySound: `var` :)

Issue #700, opened 2016-06-08
Transpile files in /bin


Rich-Harris:
I keep forgetting that I can't use ES2015 syntax in the CLI. It's annoying. Should probably just build it, same as the library itself is built. Bonus: might be able to move some dependencies to devDependencies for faster install and startup

PR #706, opened 2016-06-09
Build/bundle CLI

Rich-Harris:
See #700


.gitignore
```diff
@@ -7,3 +7,4 @@ _actual
 coverage
 .commithash
 .idea
+bin/rollup
```
TrySound: Untrack file to ignore.

Rich-Harris: oops! done, thanks

bin/src/runRollup.js
```diff
@@ -1,15 +1,17 @@
-require( 'source-map-support' ).install();
+import { resolve } from 'path';
+import relative from 'require-relative';
+import handleError from './handleError';
+import SOURCEMAPPING_URL from './sourceMappingURL.js';
```
TrySound: sourceMappingUrl.js

snuggs:
@Rich-Harris curious about this line. Why can't we import? https://github.com/rollup/rollup/blame/master/bin/src/runRollup.js#L6

Rich-Harris:
@snuggs IIRC it's because things used to get terribly confusing if you had an import that was relative to the output bundle – Rollup would try and resolve it relative to the source file and all hell would break loose (but if the import was relative to the source file, the `require` statement in the CommonJS bundle would be incorrect). We had the same thing with Acorn and I did some really nasty stuff like [this](https://github.com/Rich-Harris/acorn/blob/02dc124b80ed12a9b71a74b0f809f078333afc09/rollup/config.bin.js#L4-L7). Was fixed in #754, so we should be able to fix this now...

Rich-Harris:
> Was fixed in #754, so we should be able to fix this now...

https://github.com/rollup/rollup/pull/886

PR #707, opened 2016-06-09
rewrite top-level this as undefined

Rich-Harris:
We didn't catch that this wasn't already happening (pretty sure it did at one point), because Babel was doing it for us in the test (and we needed Babel, because we need to test behaviour inside arrow functions...)

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -41,7 +41,7 @@ export default class Statement {
 		// find references
 		const statement = this;
 		let { module, references, scope, stringLiteralRanges } = this;
-		let readDepth = 0;
+		let contextDepth = 0;
 
 		walk( this.node, {
 			enter ( node, parent, prop ) {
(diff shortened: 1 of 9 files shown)


eventualbuddha:
Looks good.

cognitom:
To bundle the code for non-Node environments, sometimes `this` is needed on top-level. For someone like me, this is a simple workaround for it, kind of:

```javascript
doSomething(that)
```

Add [intro option](https://github.com/rollup/rollup/wiki/JavaScript-API#introoutro):

```javascript
const code = bundle.generate({
  intro: 'const that = this'
}).code;
```

Then the output:

```javascript
const that = this
doSomething(that)
```

Rich-Harris:
For anyone else landing here -> #858

cognitom:
Cool!

PR #711, opened 2016-06-10
Import search path

ksherlock:
This adds a new `-I` cli option (and `importSearchPath` key for the rollup.config.js) to specify module search path(s).  (I appreciate that plugins can handle it as well but it's easier to do it via a command-line flag than dealing with plugins and config files.)


TrySound:
Search paths are very specific usecase. It shoudn't be part of core functionality if could be handle via separate module.

Rich-Harris:
Thanks for the PR @ksherlock! @TrySound is right though – this isn't something that should be in the core library, it's something that needs to happen via a plugin. [rollup-plugin-includepaths](https://github.com/dot-build/rollup-plugin-includepaths) should fit the bill.

Issue #409, opened 2016-01-01
Input source maps support


guybedford:
Not trying to make more work, but also came across this possibility. When using the `load` hook, each file could potentially have its own originating source map to chain into the Rollup optimization.

(in other news just completed the jspm Rollup work here!)

![screen shot 2016-01-01 at 21 25 32](https://cloud.githubusercontent.com/assets/598730/12071921/ac93e62c-b0ce-11e5-9537-f819926558b1.png)

guybedford:
Not needing to make more work here, and I know there are certainly other priorities. Just to note this as a possibility, feel free to close.

Rich-Harris:
I think this is a reasonable thing to consider including. A workaround in the meantime might be to use [sorcery](https://github.com/Rich-Harris/sorcery) – it *should* (never any certainties with sourcemaps, in my experience...) be able to combine the sourcemap Rollup generates with pre-existing ones.

guybedford:
Well that's a convenient project to have on hand!

When @crisptrutski did input source maps work for Traceur, it ended up being a relatively simple implementation at https://github.com/google/traceur-compiler/pull/1608/files.

Working through all the edge cases of this stuff is a pain I know though.

guybedford:
Any news on a feature like this? Input source maps would be really useful to ensure we're getting the right source map generation in jspm (currently there are offset issues due to the multi-compilation steps). If this project is using `source-map`, it is designed to work very easily with input source maps, as shown in the PR link above.

I can potentially see if I can work on something along these lines as well. Just let me know your thoughts.

Victorystick:
At first I thought we'd only need to patch `transform.js` with the following, but it didn't seem to get the source maps generated properly.

```diff
export default function transform ( source, id, transformers ) {
	let sourceMapChain = [];

	if ( typeof source === 'string' ) {
		source = {
			code: source,
			ast: null
		};
	}

+	// If the source already has a source map defined,
+	// add it to the chain of source-maps to collapse.
+	if ( source.map ) {
+		sourceMapChain.push( source.map );
+	}

	let originalCode = source.code;
	let ast = source.ast;
```

Rich-Harris:
I guess the question is how far do you go? Suppose I import some-lib.js, which has an accompanying sourcemap that includes foo.js, bar.js and baz.js, which *also* have sourcemaps pointing to foo.coffee, bar.coffee and baz.coffee? Should Rollup follow those chains as far as they go, or just the first level?

[...]

PR #715, opened 2016-06-12
Make it possible for plugin loaders to provide sourcemaps.

Permutatrix:
With this pull request, a loader can return either a `code` string or a `{ code, map }` object, just like a transformer.

The sourcemap can reference whatever sources it wants, or it can provide no sources, in which case it will be assumed to have one source which is the same as the module's ID.

One source can be referenced by multiple modules' sourcemaps; if no sourcemap provides the source content or if some sourcemaps provide source content while others don't, that will work, but an error will be thrown if multiple sourcemaps provide conflicting content for one source.

Sources, source roots, and module IDs are all assumed to function like normal paths. I'm not sure that's a good way to handle them, since a plugin could produce a sourcemap referencing a URI or maybe even some exotic thingamajig none of us have even heard of. [Mozilla's algorithm for finding sourcemap sources is a fair bit more sophisticated.](https://github.com/mozilla/source-map/blob/master/lib/util.js#L125)

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -46,12 +46,11 @@ export default class Bundle {
 				.concat( resolveId )
 		);
 
-		this.load = first(
-			this.plugins
-				.map( plugin => plugin.load )
-				.filter( Boolean )
-				.concat( load )
-		);
+		const loaders = this.plugins
+			.map( plugin => plugin.load )
+			.filter( Boolean );
+		this.hasLoaders = loaders.length !== 0;
+		this.load = first( loaders.concat( load ) );
 
 		this.transformers = this.plugins
 			.map( plugin => plugin.transform )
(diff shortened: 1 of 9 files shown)


Rich-Harris:
Thanks for the PR @Permutatrix. Could you elaborate a bit on why you need this change? Just want to make sure this is a necessary addition before making the sourcemap code even more complicated :)

Permutatrix:
@Rich-Harris People have been asking for [gulp-rollup v2](https://github.com/mcasimir/gulp-rollup/pull/16#issuecomment-156668598) for ages now, and I'm finally implementing it. This puts Rollup in the middle of the build process, so there's no telling what kinds of sources and transformations went into producing the input files. (There could be multiple Rollup steps for all we know. Probably not very useful, but interesting, and certainly impossible without this change if you want sourcemaps.)

Rich-Harris:
You mean to say that Gulp expects each plugin to account for sourcemaps generated by prior transformations, rather than composing them itself? Well, I'm glad I don't use Gulp!

But since some unfortunate people *do* use Gulp, I guess this change makes sense – and is presumably a step towards https://github.com/rollup/rollup/issues/409

Rich-Harris:
Released as 0.31.2. Thanks 👍

Permutatrix:
@Rich-Harris gulp's "thing" is that it doesn't really do anything itself and lets plugins do all the work. It doesn't even know what a sourcemap _is_&mdash;preparing and outputting them is handled by a plugin. Goofballs like me are willing to pay these prices because it makes weird tasks just as easy as normal ones.

PR #718, opened 2016-06-13
Allow `sourceMapFile` as an option.

### LINKED ISSUE #717: Unable to generate sourcemap into a separate file
I'm unable to generate sourcemap files which are not inline. Setting the `sourceMapFile` key inside the config (according to the wiki) leads to an error:

 ```js
// rollup.config.js
export default {
    entry: 'src/main.js',
    format: 'iife',
    moduleName: 'RollupTest',
    dest: 'dist/bundle.js',
    sourceMap: true,
    sourceMapFile: 'dist/bundle.js.map',
    ...
}
```
```
Unexpected key 'sourceMapFile' found, expected one of: acorn, banner, cache, dest, entry, exports, external, footer, format, globals, indent, intro, moduleId, moduleName, noConflict, onwarn, outro, plugins, preferConst, sourceMap, targets, treeshake, useStrict
```

```js
// src/main.js
var foo = 42;

function main() {
    console.log(foo);
    return foo;
}

function woa() {
    return 5;
}

export default main;
```

Without `sourceMapFile` I'm able to bundle the above file with the folllowing output:
```js
var RollupTest = (function () {
    'use strict';

    var foo = 42;

    function main() {
        console.log(foo);
        return foo;
    }

    return main;

}());
//# sourceMappingURL=bundle.js.map
```

But the bundle.js.map itself is missing. No sourcemaps are rendered.

The command `rollup -m inline` writes the sourcemap correctly into bundle.js

eventualbuddha:
Fixes #717.

I didn't add a test for this because I wasn't really sure where it should go.

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -28,6 +28,7 @@ const ALLOWED_KEYS = [
 	'plugins',
 	'preferConst',
 	'sourceMap',
+	'sourceMapFile',
 	'targets',
 	'treeshake',
 	'useStrict'
--- a/test/test.js
+++ b/test/test.js
@@ -76,7 +76,7 @@ describe( 'rollup', function () {
 			return rollup.rollup({ entry: 'x', plUgins: [] }).then( function () {
 				throw new Error( 'Missing expected error' );
 			}, function ( err ) {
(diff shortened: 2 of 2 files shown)


Rich-Harris:
👍
 
Any new keys need to be added to [this test](https://github.com/rollup/rollup/blob/master/test/test.js#L75-L81) (I've done it)

PR #721, opened 2016-06-16
Added travis ci support to test against modern node versions

amilajack:


TrySound:
@amilajack Add appveyor too please

amilajack:
@TrySound never used it before. I'll look into it and try adding it. What are the differences between it and travis ci?

TrySound:
Travis is linux, appveyor is windows. It's almost the same.

Rich-Harris:
thanks – I added Node 6 to the Appveyor config but removed Node 5 from the Travis config (4/6 aka LTS/stable are priorities, can't imagine something working in 4 and 6 but not 5, so may as well save ourselves waiting around for an extra build to happen)

PR #724, opened 2016-06-17
Handle options.external is a function.

operandom:
Rollup generate TypeError when options.external is a function in rollup.config.js

bin/rollup:746:4
```
TypeError: (options.external || []).concat is not a function
```

### DIFF
--- a/bin/src/runRollup.js
+++ b/bin/src/runRollup.js
@@ -95,8 +95,15 @@ const equivalents = {
 };
 
 function execute ( options, command ) {
-	let external = ( options.external || [] )
-		.concat( command.external ? command.external.split( ',' ) : []  );
+	let external = command.external ?
+		typeof options.external === 'function' ?
+		((fn, a) => {
+			return function (id) {
+				return fn() || a.indexOf(id) !== -1;
+			};
+		})(options.external, command.external.split(',')) :
+		(options.external || []).concat(command.external.split(',')) :
+		options.external;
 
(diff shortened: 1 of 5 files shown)


eventualbuddha:
What is the new API you're proposing? Can you give an example?

operandom:
Hi @eventualbuddha !

I didn't think I was adding something new: I just set the _external_ property with a function as described [here](https://github.com/rollup/rollup/wiki/JavaScript-API#external) in my config file but `rollup -c` generated an error.

I saw that the config file options are merged with the cli command but don't handle the fact that _external_ could be a function as described in the wiki. I just changed the code to handle this case with a wrapper function.

TrySound:
There is a commit which should introduce this feature
https://github.com/rollup/rollup/commit/40217fd2e1a487bc99268f6684fe7febb68a15fb

TrySound:
However yeah, cli wan't updated. @operandom can you write test for this?
/cc @Rich-Harris

operandom:
@eventualbuddha @TrySound I think there is something strange about this cli option : if you remove the external command ([here](https://github.com/rollup/rollup/blob/master/test/cli/external-modules/_config.js#L3)), the test still pass because they are handled by rollup directly. So the only useful way to use _external_ for me is to pass some relative paths.

operandom:
Sorry, wrong button -_-

Rich-Harris:
thanks!

PR #726, opened 2016-06-17
Add build target for ES modules

### LINKED ISSUE #725: Entry file in jsnext:main is not included in npm package
Rollup specifies a jsnext:main field pointing to src/rollup.js, but this file is not included in the published npm package, as of v0.30.0.

The src directory was removed from the package.json files array in 1f97f4e4e228b22f962f2ad04220273673bbbcf5, but it should probably be added back.

maxdavidson:
Fixes #725 by adding a build target for ES modules, and pointing the `jsnext:main` field to it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.31.2",
   "description": "Next-generation ES6 module bundler",
   "main": "dist/rollup.js",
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Thanks!

Issue #672, opened 2016-06-01
Namespace object methods and this


Victorystick:
Consider [this example](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22es6%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20*%20as%20ns%20from%20'.%2Ffoo.js'%3B%5Cn%5Cnconsole.log%28%20ns.self%28%29%20%29%3B%5Cnconsole.log%28%20ns%5B'self'%5D%28%29%20%29%3B%22%7D%2C%7B%22name%22%3A%22foo.js%22%2C%22code%22%3A%22export%20function%20self%20%28%29%20%7B%5Cn%20%20%20%20return%20this%3B%5Cn%7D%22%7D%5D%7D). If a function is accessed through the namespace object in such a way that Rollup cannot statically determine which function it is, and subsequently invoked, `this` will be bound to the namespace object. I do not think that is supposed to happen.

If people start relying on this behaviour, their code will break once modules are natively supported. Can we prevent it, and how? 

``` js
// main.js
import * as ns from './foo.js';

console.log( ns.self() );     // => window or undefined
console.log( ns['self']() );  // => ns

// foo.js
export function self () {
    return this;
}
```

Rich-Harris:
I assume the spec is unambiguous on this point – that the namespace object is *not* the context in which the function is called? (Had a quick look but I'm bad at reading spec documents.)

eventualbuddha:
Yeah, we should change the way we're calling them. I recommend following babel's approach when we have to use a member expression:

```js
function self () {
    return this;
}

var ns = Object.freeze({
    self: self
});

console.log( self() );
console.log( (0, ns['self'])() );
```

Victorystick:
@eventualbuddha That'll work for dynamic access like in my example, but what about when were passing the namespace object to for example an external function?

```js
import * as ns from '...';

externalFunction(ns);
```

It still needs to be able to dynamically look up properties and call methods of the object. How should such scenarios be handled?

bfred-it:
@Rich-Harris if `wontfix` then should we close thi and #728?

guybedford:
It seems unlikely a new approach for handling this will be happening anytime soon. If there are new suggestions though, happy to reopen discussion.

PR #728, opened 2016-06-17
Add a failing test for #672.

eventualbuddha:
I'm not sure where the namespace stuff is handled. My proposed solution was to rewrite all namespaces where we couldn't simply eliminate the namespace object from the method call from `ns['self']()` to `(0, ns['self'])()`.

### DIFF
--- a/src/Statement.js
+++ b/src/Statement.js
@@ -63,6 +63,14 @@ export default class Statement {
 					module.magicString.overwrite( node.start, node.end, 'undefined' );
 				}
 
+				// rewrite `foo[bar]()` as `(0,foo[bar])()` if `foo` is an imported namespace
+				if ( node.type === 'CallExpression' && node.callee.type === 'MemberExpression' && node.callee.object.type === 'Identifier' && node.callee.computed ) {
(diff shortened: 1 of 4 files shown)


Rich-Harris:
I'm also not sure off the top of my head where this would go, but big 👍 on the general solution – never would have occurred to me, much better than faffing about with `.call`. (Though now I'm kind of surprised that `(ns['self'])()` is called with `ns` as context – seems illogical. Oh well)

Rich-Harris:
Have implemented this. Though now I'm reading @Victorystick's [comment](https://github.com/rollup/rollup/issues/672#issuecomment-226903900) and I'm wondering if this is the right approach. This approach won't catch namespaces that are passed around (whether as an argument to a function or as a namespace re-exported as a named export, etc).

Given that this probably isn't a mistake many people are making, and given that it's pretty easily fixed, is it really worth bothering with? Maybe an alternative approach would involve warning if `this` is encountered in a function that's a member of an imported namespace, but otherwise ignoring it?

guybedford:
Closing this, and leaving #672 open to track any future progress on the wontfix.

Issue #468, opened 2016-01-25
Confusing es6 format name


TrySound:
https://github.com/rollup/rollup/wiki/JavaScript-API#format
I think we should change `es6` format name to `es2015` as standard name. es6 is informal way to specify new standard features, but it shouldn't be used in documentation and especially in api.
/cc @Rich-Harris @Victorystick @eventualbuddha

rosskevin:
Agreed, I noticed that today as I switched my other nomenclature to es2015, the `format` option rejects `es2015` as a value.

Victorystick:
Yes, we'd do best to move away from the `es6` name. Perhaps we should go with `esm`; ECMAScript modules? With `es6` and `es2015` as aliases?

TrySound:
@Victorystick Why only three chars? There is `iife` with four chars. So `ecma` seems better.

Victorystick:
> Why only three chars?

`ecmascript-modules` seemed a bit long. :stuck_out_tongue: I thought we'd emphasise the _modules_ part. Since neither CommonJS, or IIFE has any mention of _module_, I suppose `ecma` makes sense. :+1:

eventualbuddha:
Having `ecma` be the real name and `es6` begin issuing a warning seems good to me.

arturparkhisenko:
Disscussion about formats in [twitter of Axel Rauschmayer](https://twitter.com/rauschma/status/693149204261294083) @rauschma

Rich-Harris:
What about just `es`? To me that makes more sense as an abbreviation of `ecmascript`. (Also, I prefer green bikesheds...)

TrySound:
@Rich-Harris Hm.. sure. Using standard abbreviation instead of "legacy" js seems much better.

yaycmyk:
What about 'native'?

TrySound:
Native is too big and incorrect. We decide `es` as new standard.

Victorystick:
:+1: `es`

:-1: bikeshedding

callumlocke:
:+1: es

I don't think this is bikeshedding, it's API design, names matter

TrySound:
It's convention like callback style for node.

eventualbuddha:
Let's do `es`. I think `native` suffers from association with native code
(i.e. compiled code).

stevemao:
> I think we should change es6 format name to es2015 as standard name. es6 is informal way to specify new standard features, but it shouldn't be used in documentation and especially in api.

[...]

PR #730, opened 2016-06-19
Add es format as alias of es6

TrySound:
Ref #468


Rich-Harris:
Agree that we should do this, but I think we should be more aggressive and change `es6` -> `es` as soon as the option is specified, and print a deprecation warning. Will make life a bit easier in the long run I think. PR inbound

Rich-Harris:
#731

PR #731, opened 2016-06-19
Deprecate `es6` format

Rich-Harris:
See #468. Alternative approach to #730 – rather than accommodating either `es6` or `es`, this changes `es6` to `es` straight away and prints a deprecation warning. Bigger upfront change but more maintainable in the long term.


test/test.js
```diff
@@ -348,15 +368,9 @@ describe( 'rollup', function () {
 				( config.skip ? it.skip : config.solo ? it.only : it )( dir, function ( done ) {
 					process.chdir( path.resolve( CLI, dir ) );
 
-					if (os.platform() === 'win32') {
-						config.command = "node " + path.resolve( __dirname, '../bin' ) + path.sep + config.command;
-					}
+					const command = 'node ' + path.resolve( __dirname, '../bin' ) + path.sep + config.command;
```
TrySound: Can't you just pass `config.command` as the last argument?

Rich-Harris: Huh – yeah, I suppose you probably could. Just seems a bit weird to pass non-path things (i.e. `--flags` and other options) to `path.resolve` – am worried it might end up mangling the options somehow

snuggs: `path.join('node ', path.resolve( __dirname, '../bin' ), config.command)`

@Rich-Harris @TrySound Functional programming FTW. 

Net/Net same result at the end of the day.

TrySound: `node ` is not part of that path.

snuggs: @TrySound confused brah. I see the words you are writing but can't put together the point.

Again... net/net is the same. What am I missing here?
capture d ecran 2016-09-06 a 03 24 24

TrySound: Semantics. `node` and `../bin` are different paths and shouldn't be normalized together.

snuggs: I guess...
```javascript
// This
'node ' + path.resolve( __dirname, '../bin' ) + path.sep + config.command

// Gives the same result as the following
'node ' + path.resolve('../bin' ) + path.sep + config.command // no need for __dirname as this is default of path.resolve.
`node ${path.resolve( __dirname, '../bin' )}/${config.command}`
path.join('node ', path.resolve('../bin' ), config.command)
```
Just stared at this with a few of my colleagues who said:
```
// "He's correct about them being in two different paths but he's "normalizing" them merely by concatenating using plus signs.
// No different than a join query in a database. So "semantics" is a good answer "shouldn't be normalized together" is not.

// Tell your friend if I were him i'd be more concerned about using `path.sep` AND ../bin. Would be more interested to // see how this works on a winDOZE machine with him using a POSIX separator. Unless path.resolve is resolving the path.sep as well.
```

I'll take your word for it tho @TrySound you're smarter than me. :-) <3 👍

Issue #722, opened 2016-06-16
Warn on same export


TrySound:
Would be great to warn if there is same exports in module. This example is not explicit and could be dangerous.
```js
// mod1
export const param = 0;
// mod2
export const param = 1;
// mod3
export * from './mod1';
export * from './mod2';
```

eventualbuddha:
To be clear, this is allowed by the spec and has well-defined semantics. However, you're right that if could indicate a bug. Warning sounds reasonable.

Rich-Harris:
Released as 0.33

PR #735, opened 2016-06-19
Fail on duplicating in export * from

TrySound:
Ref #722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # rollup changelog
 
+## 0.32.0
+
+* Deprecate `es6` format in favour of `es` ([#468](https://github.com/rollup/rollup/issues/468))
+* Add correct `jsnext:main` build ([#726](https://github.com/rollup/rollup/pull/726))
+
 ## 0.31.2
 
 * Allow `load` plugins to provide sourcemap ([#715](https://github.com/rollup/rollup/pull/715))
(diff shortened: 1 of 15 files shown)


TrySound:
What's wrong with it?

eventualbuddha:
Is this spec behavior? Last I knew this was allowed.

TrySound:
Yeah, forgot to add it optionally. But is there a side effect to fail tests?

Rich-Harris:
What if this was a warning instead? I can't imagine anyone is likely to set `failOnExportAllDup` just for the hell of it – if it caused a warning but didn't fail, then people would be more likely to find potential bugs without it causing the build to fail unnecessarily.

(test was failing because another one was out of date btw – have fixed)

TrySound:
Yeah warning is good. But I'd like to have some linting effect to prevent merge or pull request with potential bug. Maybe we could collect all matches and throw them in the end? Also the name of option is not quite good, can you suggest something?

TrySound:
Yeah, `onwarn` is good.

TrySound:
Oh, heh :) Didn't realize we already have `onwarn` option ;)

Rich-Harris:
👍  I changed the message to reflect the fact that conflicting namespaces are permitted, if discouraged (previously it said that modules *could not* export the same names, because we were treating it as an error)

TrySound:
Okay. Thanks, Rich.

Issue #736, opened 2016-06-21
rollup --config 'module-from-npm'


bfred-it:
For the sake of sharable/reusable configurations, it'd be nice to be able to require a config file from npm:

```sh
npm install rollup-config-buble-uglify --save-dev
rollup --require-config 'buble-uglify' input.js > output.js
```

Obviously this comes at the expense of configurability, but it'd be nice to have some general-use boilerplate.

It's already kinda possible, so this issue is more about discussing `rollup-config-*` type of modules.

```sh
# current solution
rollup --config 'node_modules/rollup-config-buble-uglify/index.js' input.js > output.js
```

Rich-Harris:
Maybe a convention like `rollup -c node:buble-uglify`?

Rich-Harris:
Have opened a PR for this https://github.com/rollup/rollup/pull/739

bfred-it:
Perfect! I'm going to test it today, I had already published a simple `rollup-config-buble` to npm

Rich-Harris:
available in 0.32.1

bfred-it:
:heart:

PR #739, opened 2016-06-21
Use config files installed in `node_modules`

Rich-Harris:
See #736. This PR allows you to do this...

```bash
rollup -c node:some-plugins-i-like -i src/main.js -o dist/bundle.js
```

...and as long as either `some-plugins-i-like` or `rollup-config-some-plugins-i-like` is a package that exports a [config file](https://github.com/rollup/rollup/wiki/Command-Line-Interface#using-a-config-file), Rollup will use it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .DS_Store
-node_modules
-!test/node_modules
+/node_modules
 .gobble*
 dist
 _actual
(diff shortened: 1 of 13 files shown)


eventualbuddha:
Looks good.

bfred-it:
This seems to be working 👍

Issue #733, opened 2016-06-19
Import with side-effects ignored if affected thing not exported.


mbostock:
Here’s an interesting case with [d3-selection](https://github.com/d3/d3-selection) and [d3-selection-multi](https://github.com/d3/d3-selection-multi), of which the latter mutates the selection.prototype of the [selection](https://github.com/d3/d3-selection#selection) class exported by d3-selection.

If I Rollup a bundle that does not export selection, then Rollup thinks that the selection class is unused, and so ignores the import of d3-selection-multi:

```js
import {selection, select} from "d3-selection";
import "d3-selection-multi";

select("body").append("h1")
    .text("Hello, world!")
    .styles({"text-align": "center"}); // TypeError! selection.styles is not a function
```

But if I export selection, it works:

```js
import {selection, select} from "d3-selection";
import "d3-selection-multi";

export {selection}; // 🌶

select("body").append("h1")
    .text("Hello, world!")
    .styles({"text-align": "center"}); // 👍 
```

Of course, there’s no reason to export anything when creating an IIFE bundle of your application and not a library, so it’s not a great workaround…

Related d3/d3-selection-multi#6, /cc @LeartS.

mbostock:
Here’s the live example:

http://bl.ocks.org/mbostock/648fa00c52c6ad141ab73ac020af5e84

Rich-Harris:
Thanks – you have a real knack for finding the edge cases 😀  Have opened #738

mbostock:
Thank you! I greatly appreciate you taking a look at this.

PR #740, opened 2016-06-22
WIP: failing test for #733

Rich-Harris:


Rich-Harris:
Added test to #902, it passes. Closing

PR #751, opened 2016-06-25
FIX: --no-strict command line option not map bug

futurist:
I've found `--no-strict` option is not mapping from cli, this PR fix it.

Also added test for this.


Rich-Harris:
thanks! 🍺

PR #760, opened 2016-06-30
Use fs.realpathSync() rather than path.resolve() on the config path

hh10k:
It's possible for the require.extensions['.js'] override to fail on Windows because path.resolve(...) might return "c:\..." while fs.realpath(...) returns "C:\...", and the exact match fails.

### DIFF
--- a/test/cli/config-cwd-case-insensitive/_config.js
+++ b/test/cli/config-cwd-case-insensitive/_config.js
@@ -1,0 +1,13 @@
+var os = require( 'os' );
+
+function toggleCase ( s ) {
+  return ( s == s.toLowerCase() ) ? s.toUpperCase() : s.toLowerCase();
+}
+
+module.exports = {
+  skip: os.platform() !== 'win32',
+	description: "can load config with cwd that doesn't match realpath",
+	command: 'rollup -c',
+  cwd: process.cwd().replace( /^[A-Z]:\\/ig, toggleCase ),
+	execute: true
+};
--- a/test/cli/config-cwd-case-insensitive/_config.js
+++ b/test/cli/config-cwd-case-insensitive/_config.js
@@ -8,6 +8,6 @@ module.exports = {
   skip: os.platform() !== 'win32',
 	description: "can load config with cwd that doesn't match realpath",
 	command: 'rollup -c',
-  cwd: process.cwd().replace( /^[A-Z]:\\/ig, toggleCase ),
+  cwd: __dirname.replace( /^[A-Z]:\\/i, toggleCase ),
 	execute: true
 };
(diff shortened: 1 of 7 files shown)


hh10k:
I've used Rollup from the command line many times and never had a problem, but found this was a problem when launching a task from Visual Studio Code.

TrySound:
Write a test please

hh10k:
Test added.

It seems the same bug affects the sourcemaps/relative-paths, so I'll give that one a go too.

hh10k:
The sourcemaps/relative-paths problem was with the test itself.  I've removed the use of process.cwd(), and instead explicitly set the directory before each test is run, avoiding the issue.

All tests now pass on Windows.

test/cli/config-cwd-case-insensitive/_config.js
```diff
@@ -0,0 +1,13 @@
+var os = require( 'os' );
+
+function toggleCase ( s ) {
+  return ( s == s.toLowerCase() ) ? s.toUpperCase() : s.toLowerCase();
+}
+
+module.exports = {
+  skip: os.platform() !== 'win32',
```
TrySound: Should be passed on all systems. And something wrong with your indentation.

hh10k: The test is specific to Windows.  I can remove the `skip` and it should still pass on OSX and Linux, but it's meaningless.

Can you be more specific about what indentation?  I guess maybe placing all of a ternary statement on a single line?  Where can I find the Rollup coding style guidelines?

TrySound: Do you think this object looks good?
```js
{
	description: "can load config with cwd that doesn't match realpath",
	command: 'rollup -c',
  cwd: __dirname.replace( /^[A-Z]:\\/i, toggleCase ),
	execute: true
};
```

TrySound: Test shouldn't be skipped because we need to make sure that your fix won't break another OS.

hh10k: Oh, sorry I didn't notice that you're using actual tabs.  My bad!  Will be fixed in next PR.

hh10k: All the other tests should verify that my fix doesn't break other operating systems.  This test is ensure the config paths that start with `c:\` work equally well as `C:\`, which is a feature that other OSs don't have (at least, those supported by Node).

The name is probably a bit misleading now, it took some experimentation before I understood what the problem really was.

Rich-Harris:
awesome, thanks @hh10k!

PR #773, opened 2016-07-07
Give ongenerate/onwrite more access

tivac:
This proposal came out of the conversation in #742 to try and pass a bit more information along to `ongenerate`/`onwrite` so that they can be more useful. Specifically trying to make it so that `ongenerate` is able to add values to the output of `.generate()` so that `onwrite` or API users can see it. This would allow me to clean up some gross stuff I'm doing in the `modular-css` rollup plugin, and hopefully will allow clearer differentiation between the use-cases for `ongenerate` and `onwrite`.

@Rich-Harris @TrySound would love your thoughts on whether or not this improves things. It's a small change & all tests still pass w/o modification.

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -60,7 +60,7 @@ export function rollup ( options ) {
 				if ( plugin.ongenerate ) {
 					plugin.ongenerate( assign({
 						bundle: result
-					}, options ));
+					}, options ), rendered);
 				}
 			});
 
@@ -79,7 +79,8 @@ export function rollup ( options ) {
 				}
 
 				const dest = options.dest;
-				let { code, map } = generate( options );
+				let output = generate( options );
+				let { code, map } = output;
 
 				let promises = [];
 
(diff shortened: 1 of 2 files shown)


src/rollup.js
```diff
@@ -79,7 +79,9 @@ export function rollup ( options ) {
 				}
 
 				const dest = options.dest;
-				let { code, map } = generate( options );
+				let output = generate( options );
+				let code = output.code;
+				let map = output.map;
```
TrySound:
```js
let { code, map } = output;
```

Rich-Harris:
Yeah, this seems to basically make sense. So you're thinking that a plugin would do some work in `ongenerate`, attach it to the bundle, then the same plugin could use the attached property inside `onwrite`?

Was thinking about the async case – if `ongenerate` were to attach a promise or a promise-generating function, that could be used by either `onwrite` or whatever custom build script is being used. Does that sound workable?

Any chance you could add a couple of tests to this branch? Thanks

tivac:
@Rich-Harris 

1. **Async usage**
    
 Yup, promises-as-added-properties makes the async `ongenerate` case workable without complicating the consumer code unless they need to access it. Since `onwrite` are already async-supporting via `Promise.all` it's pretty natural.

 I posted an example of how I'd do that exactly that in the modular-css rollup plugin over here: https://github.com/rollup/rollup/pull/742#issuecomment-230965890

2. **Tests**

 Yes, definitely. Wanted to get a yay/nay on the basic idea before fleshing it out fully!

Rich-Harris:
Nice, in it goes! Thanks

PR #780, opened 2016-07-10
support paths config

### LINKED ISSUE #754: `paths` option
Scenario: your bundle depends on an external library, but you don't want to do this sort of thing:

```js
import react from 'https://npmcdn.com/react@15.0.1/dist/react.min.js';

// or, e.g. in /bin/my-lib
import myLib from '../dist/my-lib.js'
```

In the second case, you can end up in [all sorts of contortions](https://github.com/Rich-Harris/acorn/blob/02dc124b80ed12a9b71a74b0f809f078333afc09/rollup/config.bin.js#L4-L7) if the relative path from the bundle isn't the same as the relative path from the source file.

Proposed solution: use an external module ID, as normal, but then:

```js
// rollup.config.js
export default {
  entry: 'src/main.js',
  dest: 'bundle.js',
  paths: {
    'react': 'https://npmcdn.com/react@15.0.1/dist/react.min.js',
    'my-lib': '../dist/my-lib.js' // relative to bundle
  },
  // ...
};
```

Rich-Harris:
closes #754. Allows you to specify the path of an external module, making it straightforward to (for example) create an AMD bundle that loads specific libraries from a CDN

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -52,6 +52,12 @@ export default class Bundle {
 		this.hasLoaders = loaders.length !== 0;
 		this.load = first( loaders.concat( load ) );
 
+		this.getPath = typeof options.paths === 'function' ?
+			( id => options.paths( id ) || this.getPathRelativeToEntryDirname( id ) ) :
+			options.paths ?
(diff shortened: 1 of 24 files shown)


eventualbuddha:
👍

PR #781, opened 2016-07-10
Introduce eslint to tests.  Complaints are fixed, but none of them major.

hh10k:
This includes the tab changes that I was going to do in #760.

Probably the most controversial thing is relaxing the use of unused function arguments, but it's only restricted to the tests directory.  I don't understand why you would want to delete arguments when they're useful documentation about the function's interface.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,6 @@
     "env": {
         "es6": true,
         "browser": true,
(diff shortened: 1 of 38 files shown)


Rich-Harris:
nice, thanks 👍

PR #792, opened 2016-07-23
Improve code readability with prefering const

TrySound:


eventualbuddha:
I happen to disagree with the "prefer const" approach. Here's a good exploration of why: https://madhatted.com/2016/1/25/let-it-be

That said, I'd prefer to have consistency.

TrySound:
I like the way of making all variables immutable. This let us to be sure parts of code are mean the same in whole file. This simplifies learning code and just figure out what's going on.

eventualbuddha:
That's part of my problem with `const`: it doesn't mean the value is immutable, just the binding. So you get assurance that the binding is unchanged but days nothing about the value. Also, as the article says, I find it to be a refactoring hazard. If all you're doing is using it to avoid reusing bindings for unrelated values… maybe that's worth it if people don't know not to do that?

TrySound:
I found really helpful to save primitives (including references) from changing. Even if some collections are changes they won't change their meaning. But yeah it's just my opinion and my experience.

PR #800, opened 2016-07-28
Bugfix: avoid infinite recursion in Bundle.sort()

phter:


eventualbuddha:
Can you add a test case that would exhibit the infinite recursion you mention? We'd like to be able to verify that this fixes a real problem.

phter:
I ran into this bug while working on a rather complex library, so no simple test case, sorry.
But please just look at the current code of findParent(), it seems simple enough:
The function never returns a value, yet it contains the line
`if ( findParent( dependency ) ) return;` 
And it never checks if a module has already been visited, which doesn't go well with cyclical structures.

phter:
npm install
npm run build
--> Maximum call stack size exceeded at findParent ...
Do I get a t-shirt now?

[rollup-test.zip](https://github.com/rollup/rollup/files/389045/rollup-test.zip)

TrySound:
@phter Nice. Can you add it in PR?

phter:
@TrySound Like this?

phter:
@TrySound Uh, I messed up the branches of my forked repo, now some bogus commits appear to be part of the PR. That's annoying. Does it matter, though?

TrySound:
@phter Yep. Can you rebase and squash them (you don't need to create new PR for this)?

phter:
@TrySound ok, my git skills are somewhat lacking... but it seems to have worked.

TrySound:
Nice. Thanks. Will check it out soon.

TrySound:
https://github.com/rollup/rollup/commit/6ee27c6a7dcf5619074bc9c686386d0e948aa0d1

PR #804, opened 2016-07-28
Add explicit entry resolve error message

### LINKED ISSUE #802: Better error message for nonexistent entry
If you accidentally misconfigure rollup by using a nonexistent entry, perhaps by an incorrect resolution through a plugin, it may not be immediately apparent why rollup is failing - the error message is `Error: Could not load null: path must be a string`. Maybe the error message could note that it was the entry that could not be loaded? It's not clear where `null` is coming from either.

```js
import {rollup} from 'rollup';

rollup.rollup({
  entry: '/i-really-hope-you-dont-have-this-path'
}).catch((err) => console.log('message', err.message));
```

skeggse:
Fixes #802

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -88,6 +88,7 @@ export default class Bundle {
 		// of the entry module's dependencies
 		return this.resolveId( this.entry, undefined )
 			.then( id => {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Much nicer, thank you!

Issue #753, opened 2016-06-27
Not quite clean error


TrySound:
```
Error: Error transforming c:/_host/project/src/components/map.cmt.js with 'buble' plugin: Error transforming c:/_host/project/src/components/map.cmt.js: Error transforming c:/_host/project/src/components/map.cmt.js with 'eslint' plugin: Warnings or errors were found
```

There is an error only in eslint but it affects also next buble.

PR #814, opened 2016-07-31
Pretty error

TrySound:


TrySound:
Ref https://github.com/rollup/rollup/issues/753

PR #815, opened 2016-07-31
Preserve comments ahead of import statements.

rberner:
This should allow comments to stay above the imported module code when bundled. I did this so that my file name and code coverage specific comments would be preserved.


TrySound:
@rberner Can you please add test case for this?

Rich-Harris:
Have added a test. Thanks @rberner!

Issue #813, opened 2016-07-30
Circular exports cause error "Cannot convert undefined or null to object"


paul-go:
Upon trying to find a work around for #810, it appears that if the exports are distant, we get another error:

```
// a.js
import * as x from "./x.a.js";
x.b();
x.c();

// x.a.js
export * from "./b.js";
export * from "./c.js";

// b.js
import * as x from "./x.b.js";

export function b() {}

// x.b.js
export * from "./a.js";
export * from "./c.js";

// c.js
import * as x from "./x.c.js";

export function c() {}

// x.c.js
export * from "./a.js";
export * from "./b.js";
```

```
Cannot convert undefined or null to object
TypeError: Cannot convert undefined or null to object
    at /Users/Paul/.npm-packages/lib/node_modules/rollup/src/Bundle.js:221:6
    at Array.forEach (native)
    at /Users/Paul/.npm-packages/lib/node_modules/rollup/src/Bundle.js:216:30
    at process._tickCallback (internal/process/next_tick.js:103:7)
    at Module.runMain (module.js:577:11)
    at run (node.js:348:7)
    at startup (node.js:140:9)
    at node.js:463:3
```

PR #821, opened 2016-08-01
Failing test for #813

TrySound:


phter:
#830 will fix this, though I'm not completely sure, it's the right thing to do.

Issue #723, opened 2016-06-16
Speed up incremental builds


Rich-Harris:
From https://github.com/rollup/rollup-watch/issues/5:

> First off, congrats on the release. This is a much awaited feature.

> The build times however are not too great. I'm bundling react and a couple of internal files and my build times look like:

> 2715ms > 2065ms > 2052ms > ...

> 700ms is definitely an improvement, but not by much. What needs to be done to improve the rebuild speed?

We should spend some time profiling to see where the bulk of the work is happening, and where the low-hanging fruit is. A couple of thoughts though:

* it's possible that tree-shaking doesn't have any effect on a library like React, in which case perhaps we can disable it on a per-module basis
* in most cases, the bindings that are being imported from React aren't changing. So if there was a way to capture the result of the tree-shaking step for a given module *indexed by which exports are used* then we could re-use the analysis

Getting rudimentary caching and file watching was only really the first step, these are the sorts of optimisations we need to start thinking about now.

PaulBGD:
Disabling treeshaking cuts the times by about 150ms for the initial build and rebuilds.

Victorystick:
@PaulBGD Is that using the`--no-treeshake` option?

PaulBGD:
In the config, but yeah.

TrySound:
About cold start. Maybe we should support this pattern and follow it also in gobble?
https://github.com/avajs/find-cache-dir

TrySound:
@Rich-Harris Some benchmarking. Runs in acorn build:main (two targets), window8, node0.12
```
Phase 1 - discovery
start
[ 0, 013008154 ]
end
start
[ 0, 416093818 ]
end

Phase 2 - binding
start
[ 0, 001319085 ]
end
start
[ 0, 004428150 ]
end

Phase 3 - marking (mostly running)
start
[ 0, 001532979 ]
end
start
[ 0, 007409536 ]
end

Phase 4 - sorting and deconflicting
start
[ 0, 001402426 ]
end
start
[ 0, 000995164 ]
end
```

TrySound:
Seems like phase two can be successfully cached two.

TrySound:
Phases 2-4 don't need to cache they are extremely fast.
Generating takes about half or third time.
```
[21:07:43] Starting 'module:js'...
0 339.007502
[21:07:44] Finished 'module:js' after 1.02 s
[21:07:45] Starting 'module:js'...
0 379.969004
[...]

PR #823, opened 2016-08-02
Module render performance tweak

TrySound:
Ref https://github.com/rollup/rollup/issues/723

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -447,7 +447,7 @@ export default class Module {
 	}
 
 	render ( es ) {
-		let magicString = this.magicString.clone();
+		let magicString = this.magicString;
 
 		this.statements.forEach( statement => {
 			if ( !statement.isIncluded ) {


ericf:
This was not safe, and started breaking my build.

Previously with `v0.34.3` my `format: 'es'` build worked correctly, then it started failing after this change. It was changing `React.createElement` to `React__default.createElement` but never defining `React__default`.

TrySound:
@ericf Thank you for report. Will be fixed ASAP.

TrySound:
@ericf Can you produce some minimal test case? Can't figure out how to build your repo on windows.

ericf:
@TrySound I figured out the issue, and it makes sense why this change would cause it. I am calling `rollup()` once, but calling `bundle()` twice. I created this script before the targets feature was added: https://github.com/yahoo/react-intl/blob/master/scripts/build-lib.js#L31-L41

I will update my script to be more correct to fix this.

yairEO:
@ericf - well? did the merge break your code or all is good?

Rich-Harris:
@yairEO the change in this PR was undone in https://github.com/rollup/rollup/commit/1d7c853ce6fabec207401a53d8210ea25f08d72e

PR #830, opened 2016-08-05
Initialize member in constructor, fixes #813

### LINKED ISSUE #813: Circular exports cause error "Cannot convert undefined or null to object"

phter:
This might do the trick.
But since fetchAllDependencies() and fetchModule() call each other, this may just hide the real bug / introduce new ones... not sure.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -218,7 +218,6 @@ export default class Bundle {
 				this.moduleById.set( id, module );
 
 				return this.fetchAllDependencies( module ).then( () => {
-					module.exportsAll = blank();
 					keys( module.exports ).forEach( name => {
 						module.exportsAll[name] = module.id;
 					});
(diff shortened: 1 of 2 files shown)


TrySound:
Cool, man! Thanks

PR #835, opened 2016-08-07
Leave it up to resolveId to normalize the entry path.

Permutatrix:
Normalizing the entry ID before passing it to any plugins is redundant, because if it's a path, it will be normalized in resolveId like any other ID. If it isn't a path, normalizing it before passing it to resolveId may mangle it in a surprising way, such as by replacing backslashes with forward slashes.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -34,7 +34,7 @@ export default class Bundle {
 			}
 		});
 
-		this.entry = normalize( options.entry );
+		this.entry = options.entry;
 		this.entryId = null;
 		this.entryModule = null;
 
--- a/test/function/does-not-mangle-entry-point/_config.js
+++ b/test/function/does-not-mangle-entry-point/_config.js
@@ -1,0 +1,26 @@
+var path = require( 'path' );
+var fs = require( 'fs' );
+var assert = require( 'assert' );
+
+var modules = {
+	'x\\y': 'export default 42;',
+	'x/y': 'export default 24;'
+};
+
+module.exports = {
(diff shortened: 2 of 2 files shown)


TrySound:
@Permutatrix Surprising test? :)

Permutatrix:
[You mean like this?](https://github.com/Permutatrix/rollup-plugin-hypothetical/blob/3928408328c4c2031c1ecf026d27bc593a4ec93b/test/test.js#L164) That test fails without this PR and succeeds with it. Or do you want me to add a new test to the PR?

TrySound:
Yep, it the PR. We should prevent accidental adding it again.

Permutatrix:
Alright, there you go. Are tests a requirement for every PR?

TrySound:
@Permutatrix If it modifies functionality, yes. Can this PR break something? I mean should I publish middle version?

Permutatrix:
@TrySound It's a long shot, but there could be a very poorly configured setup out there somewhere that looks something like this:

```js
// rollup.config.js
import hypothetical from 'rollup-plugin-hypothetical';

export default {
  entry: 'x\\y',
  plugins: [hypothetical({
    files: {
      'x/y': `
        console.log("What a contrived example!");
      `
    },
    leaveIdsAlone: true
  })]
};
```

Any change that fixes behavior has the potential to break things that relied on the incorrect behavior, though. I wouldn't bump the minor version number for this myself, but it's obviously up to you.

TrySound:
Thanks.

TrySound:
0.34.7

PR #841, opened 2016-08-09
[WIP] Optimize dynamic access of namespace imports with literal keys

eventualbuddha:
This doesn’t fully work yet since it causes the erroneous-nested-member-expression test to regress. Any thoughts how to make these play nice together?

This would fix https://github.com/rollup/rollup-plugin-babel/issues/65.


Rich-Harris:
Closing in favour of #1240

PR #886, opened 2016-08-30
replace require kludge in runRollup.js

Rich-Harris:
See https://github.com/rollup/rollup/pull/706#issuecomment-243290485

### DIFF
--- a/bin/src/runRollup.js
+++ b/bin/src/runRollup.js
@@ -1,9 +1,8 @@
 import { realpathSync } from 'fs';
+import { rollup } from 'rollup';
 import relative from 'require-relative';
 import handleError from './handleError';
 import SOURCEMAPPING_URL from './sourceMappingUrl.js';
-
-const rollup = require( '../dist/rollup.js' ); // TODO make this an import, somehow
 
 import { install as installSourcemapSupport } from 'source-map-support';
 installSourcemapSupport();
@@ -58,7 +57,7 @@ export default function runRollup ( command ) {
 			config = realpathSync( config );
 		}
 
-		rollup.rollup({
+		rollup({
 			entry: config,
 			onwarn: message => {
 				if ( /Treating .+ as external dependency/.test( message ) ) return;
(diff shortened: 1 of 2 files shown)


snuggs:
Awe @Rich-Harris you should~~n't~~ have! :-) <3 👍

bin/src/runRollup.js
```diff
@@ -217,7 +216,7 @@ function bundle ( options ) {
 		handleError({ code: 'MISSING_INPUT_OPTION' });
 	}
 
-	return rollup.rollup( options ).then( bundle => {
+	return rollup( options ).then( bundle => {
```
snuggs: So will `rollup.rollup(...)` be a deprecated API? Perhaps the `dist` is named rollup but is merely a module wrapper around `rollup`.

Thanks for all the knowledge. I'm new here but willing to help @Rich-Harris

snuggs:
LGTM 🚢  :shipit:

bin/src/runRollup.js
Rich-Harris: If you import the namespace `rollup = require('rollup')` or `import * as rollup ...` then you have to do `rollup.rollup`, but if you import it directly (`import { rollup } ...`) then you don't – I guess it's just a matter of personal preference. I'd generally expect people to use `rollup.rollup` in CommonJS land and `rollup` in ES module land (same as `fs.readFile(...)` as opposed to `readFile(...)`) but it's a matter of preference.

The `rollup.rollup` duplication was just me amusing myself, I got a kick out of imagining everyone trying to figure out why the [Magical Mystery Tour](https://www.youtube.com/watch?v=kTOyVcs2jkI) was going round and round in their head while they were coding...

snuggs: @Rich-Harris more like the Christmas Hat Cat. I love that guy. SO unfortunate about his neck tho. https://www.youtube.com/watch?v=159ERq6Kf9Y

/cc @GGAlanSmithee

PR #890, opened 2016-09-01
Add options as second arg to .transformBundle

### LINKED ISSUE #867: transformBundle plugin hook should pass format option
I'm doing this plugin:

https://github.com/futurist/rollup-plugin-es3

using transformBundle plugin hook, but want to only transform `cjs` format, seems it's lack the option from plugin to check which format the code from. Maybe:

```js
    transformBundle: function (code, [format??] ) {
      if(format=='cjs');  //do some thing
    }
```

Any idea?

tivac:
Currently only piping through `format`, could certainly pass others if necessary. Maybe all options should be passed through?

Fixes #867


Rich-Harris:
Thanks, have released as 0.34.13.

> Maybe all options should be passed through?

Much better to add them as needed, in my view – that way people don't end up relying on undocumented and untested things that we later decide to change

PR #902, opened 2016-09-06
[WIP] Rewrite

### LINKED ISSUE #901: Incorrectly removing code
bool.js:

```js
export default true
```

index.js:

```js
import bool from "./bool"

const hs = document.documentElement.style

if (bool) {
    hs.color = "#222"
}

console.log("hey")
```

command:

```
rollup index.js
```

output:

```js
console.log("hey")
```

😢

Rich-Harris:
Over the weekend I got a chance to scratch an itch I've had for a while. There are some surprising bugs with the current version of Rollup, such as https://github.com/rollup/rollup/issues/901, that are rather difficult to fix with the existing codebase. On top of that, our tree-shaking is too coarse (it only operates on top-level statements/declarations) and too conservative (it sees potential effects where there are none).

This PR intends to lay the foundations for much more sophisticated static analysis, which will allow us to fix those bugs (I just checked – it fixes #901) and eventually do a much better job of tree-shaking libraries like Lodash, which have a lot of statements that look like they have effects but don't.

It's basically a complete rewrite – sorry – so I'll try to give a high-level overview. Previously, each `Bundle` was composed of several `Module` objects, each of which had an array of `Statement` objects – top-level statements and declarations. Now, there's no `Statement` – instead, we analyse the AST nodes themselves.

In order to do that, the AST nodes are 'enhanced' with various properties and methods. Each node's `initialise` method is called, causing declarations (variable, function or class) to register themselves with the current scope. (Each node is responsible for initialising its children, which means that for example a `'x' !== 'x' ? a() : b()` expression can stop `a()` being initialised.) Later, we `bind` each node, such that each reference is linked to the declaration it references.

The final step before rendering is to `run` each node that has effects. (An 'effect' means that it impacts the outside world in some way – e.g. `a = 1` only has an effect if `a` is exported or used elsewhere, but `window.a = 1` always has an effect.) This means walking over the AST and marking statements so that they get included in the final bundle. If we walk over an `Identifier` that references a declaration, we 'activate' the declaration – so if we encounter `foo()` and Rollup judges that the `foo` function has effects (if it's unsure, it errs on the side of caution), both `foo()` and `function foo () {...}` will be included.

Key to this is the idea that we can track the values of nodes. Because we know that `foo` references a function declaration, it's easy to know what code gets run when `foo()` happens. If instead we had `var foo; foo = condition ? a : b`, we know that there are two possible values for `foo` – if either `a()` or `b()` would have effects, so does `foo()`. If `foo` was a global or something imported from an external module, Rollup would assume it has effects.

This all turns out to be less hairy than you might imagine, though I'll readily confess that, as with most large rewrites, the purity of the original idea isn't quite reflected in the implementation. There are some rough edges, some unnecessarily verbose code, and doubtless some bugs. It isn't *quite* ready for merge just yet – need to test the performance, try it out with some more real-world code, etc (plus there are two failing tests related to detecting when cyclical dependencies are in the wrong order), but if @rollup/collaborators are on board then I'd like to merge it before too long otherwise it will block all other work.

----

## Next steps

Assuming there are no showstoppers here – and I'm reasonably confident with this general approach – phase 2 of the operation will be to add more value-tracking logic. For example, this...

```js
var string = 'x';
string = string.replace( 'x', 'y' );
```

...doesn't have an effect, but because Rollup sees the call expression and doesn't know that `String.prototype.replace` is effect-free, it includes both lines in the output regardless of whether `string` is used. And in this case...

```js
class Foo {
  a () { ... }
  b () { ... }
}

const x = new Foo();
x.a();
```

...it should be possible to remove the `b` method if Rollup understands a) properties and b) classes/prototypes.


.eslintrc
```diff
@@ -27,9 +27,18 @@
...
         "node": true
     },
-    "extends": "eslint:recommended",
+    "extends": [
+        "eslint:recommended",
+        "plugin:import/errors",
+        "plugin:import/warnings"
+    ],
     "parserOptions": {
         "ecmaVersion": 6,
```
TrySound: Could be seven?

Rich-Harris: yeah, will change that

TrySound:
What about caching? With buble this way is ok because we have incremental build. But rollup will process all hundreds files. I can try home to compare branches.

Rich-Harris:
> What about caching?

Right now, ASTs aren't reused between runs, meaning Acorn has to re-parse each module. That should be fixable, and once it is I'd expect the caching to have the same impact on incremental builds as before.

TrySound:
We need to update acorn
```
Message:
    TypeError: Class constructor DestructuringErrors cannot be invoked without 'new' in project\src\app\dataServices\faults.srv.js
    at Parser.pp$3.parseMaybeAssign (project\node_modules\rollup\dist\rollup.js:3157:54)
```

TrySound:
Should be fine with the last version
https://github.com/ternjs/acorn/blob/master/src/expression.js#L115

TrySound:
Hm.. acorn need to release... So long.

TrySound:
Let's enable classes transpilling for now. Do same **** for angular2 because they like decorators which requires reflection polyfill which requires transpilling because of the same "do not call class".

TrySound:
And now I got
``
---
5162 :     Node.prototype.bind.call( this, scope );
5163 :   };
5164 :
5165 :   AssignmentExpression.prototype.hasEffects = function hasEffects ( scope ) {
5166 :     const hasEffects = this.isUsedByBundle() || this.right.hasEffects( scope );
                 ^^^^^^^^^^
hasEffects is already declared (5166:8)`
``
Is that a bug in node or somthing else?

Rich-Harris:
I don't quite follow the `class` thing, why do we need do transpile them? They're supported natively in Node 4 upwards.

> Is that a bug in node or somthing else?

Hmm, possibly, what Node version are you on? A `const` that shares a name with the ID of the function expression it's inside shouldn't be an error.

TrySound:
v6.3.1

We can't invoke classes without `new` keyword. acorn contains `DestructuringError.call()` which was removed but not released yet.

TrySound:
Enabled transpilling only for acorn. It works! Finally. Now two times slower.

Rich-Harris:
Weird, was working fine for me. 

> Now two times slower

TrySound:
As you said ast is not cached yet. This is the result. I'm checking on that big project.

TrySound:
I think we should fix it before merge to not make rollup worse.

TrySound:
Ah, and I mean incremental build. Not regular.

Rich-Harris:
> I think we should fix it before merge to not make rollup worse.

Agree. Done 83ccb9725374e0fde9d07043959c397b15d26c67

TrySound:
Not much faster, This branch still slower than before.
```
Starting 'app'...
Finished 'app' after 520 ms
Starting 'app'...
Finished 'app' after 523 ms
Starting 'app'...
Finished 'app' after 552 ms
Starting 'app'...
Finished 'app' after 662 ms
Starting 'app'...
Finished 'app' after 530 ms
Starting 'app'...
Finished 'app' after 533 ms
Starting 'app'...
Finished 'app' after 546 ms
Starting 'app'...
Finished 'app' after 601 ms
```

```
Finished 'app' after 915 ms
Starting 'app'...
Finished 'app' after 1.02 s
Starting 'app'...
Finished 'app' after 866 ms
Starting 'app'...
Finished 'app' after 835 ms
Starting 'app'...
Finished 'app' after 837 ms
Starting 'app'...
Finished 'app' after 1.13 s
Starting 'app'...
Finished 'app' after 816 ms
Starting 'app'...
Finished 'app' after 795 ms
Starting 'app'...
Finished 'app' after 841 ms
Starting 'app'...
Finished 'app' after 821 ms
Starting 'app'...
Finished 'app' after 823 ms
Starting 'app'...
Finished 'app' after 804 ms
Starting 'app'...
Finished 'app' after 800 ms
Starting 'app'...
Finished 'app' after 804 ms
```

TrySound:
https://github.com/ternjs/acorn/releases/tag/4.0.0

Rich-Harris:
> Not much faster, This branch still slower than before.

Yeah, I'd expect it to be slightly slower at this point, I haven't looked into performance yet – just focusing on getting correct output. (Found a couple of bugs that aren't yet in the test suite, fixes to come.) If you have any insight into where most of the time is being spent it will help speed things up 😀 

> https://github.com/ternjs/acorn/releases/tag/4.0.0

Cool – looks like some tests will need to be updated to accommodate it

TrySound:
I'll do some benchmark this evening.

danbucholtz:
Great work team!

src/ast/nodes/VariableDeclaration.js
```diff
@@ -0,0 +1,98 @@
...
+		super.initialise( scope );
+	}
+
+	render ( code, es ) {
+		const treeshake = this.module.bundle.treeshake;
+
+		let shouldSeparate = false;
+		let separator;
+
+		if ( this.scope.isModuleScope && !/forStatement/.test( this.parent.type ) ) {
```
eventualbuddha: Is this supposed to be `ForStatement`?

src/ast/nodes/shared/callHasEffects.js
```diff
@@ -0,0 +1,65 @@
+import flatten from '../../utils/flatten.js';
+import isReference from '../../utils/isReference.js';
+import pureFunctions from './pureFunctions.js';
+import { UNKNOWN } from '../../values.js';
+
+function fnHasEffects ( fn ) {
+	if ( fn._calling ) return true; // prevent infinite loops... TODO there must be a better way
```
eventualbuddha: Use a `Set` to represent the current set of "calling" functions?

src/ast/nodes/VariableDeclaration.js
Rich-Harris: ugh, actually it was supposed to be

```js
const forStatement = /^For(?:Of|In)?Statement/;

// ...
`!forStatement.test(...)
```

good catch! will fix

src/ast/nodes/shared/callHasEffects.js
Rich-Harris: 👍 done

eventualbuddha:
This branch seems to work with decaffeinate and its various projects. However, it does fail to treeshake a helper added by babel, the `set` helper. The simplest case I can come up with is this:

```js
function set() {
  set();
}
```

That function is preserved despite its only reference being inside itself.

Rich-Harris:
> That function is preserved despite its only reference being inside itself.

ah, of course – if a function calls itself Rollup assumes the worse, and it doesn't actually need to. Fix inbound

eventualbuddha:
I just tried to verify that your fix caused `set` not to appear, but it still does. I copy-pasted it into its own file and ran this branch on it and got an exception:

```js
// index.js
var set = function set(object, property, value, receiver) {
  var desc = Object.getOwnPropertyDescriptor(object, property);

  if (desc === undefined) {
    var parent = Object.getPrototypeOf(object);

    if (parent !== null) {
      set(parent, property, value, receiver);
    }
  } else if ("value" in desc && desc.writable) {
    desc.value = value;
  } else {
    var setter = desc.set;

    if (setter !== undefined) {
      setter.call(receiver, value);
    }
  }

  return value;
};
```

```
$ node ~/lib/rollup/bin/rollup index.js
Cannot use 'in' operator to search for 'length' in value
TypeError: Cannot use 'in' operator to search for 'length' in value
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:34:19)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:36:4)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
    at enhanceNode (/Users/donovan/lib/rollup/src/ast/enhance.js:59:3)
Type rollup --help for help, or visit https://github.com/rollup/rollup/wiki
```

It turns out it's trying to enhance the `value` property of this node:

```js
Node {
  type: 'Literal',
  start: 294,
  end: 301,
  value: 'value',
  raw: '"value"' }
```

Rich-Harris:
> It turns out it's trying to enhance the `value` property of this node:

Ah, I know why that is. It doesn't have a lookup table of child nodes for each node type, instead it builds that table as it encounters each node – so the first time it encounters a `VariableDeclaration`, it adds `VariableDeclaration: [ 'declarations' ]` to `keys`, allowing it to skip property enumeration the next time it sees one.

If the first `Literal` it encounters had a non-primitive value, it would mistakenly add `value` to the list of child keys for `Literal` nodes. Should be easily fixed by adding `Literal: []` to https://github.com/rollup/rollup/blob/rewrite/src/ast/keys.js. Not able to do that right now though, am under the cosh.

As to why `set` is continuing to appear, that's a deeper mystery – will have to look into that later.

TrySound:
Benchmark
```
[00:34:14] Starting 'app'...
1628ms: --BUILD--
757ms: transform
151ms: ast
218ms: analyse
97ms: phases234
88ms: treeshake
546ms: --GENERATE--
173ms: render modules
368ms: sourceMap
[00:34:16] Finished 'app' after 2.27 s
[00:34:16] Starting 'app:dev'...
[00:34:16] Finished 'app:dev' after 7.53 ms
[00:34:19] Starting 'app'...
637ms: --BUILD--
0ms: transform
26ms: ast
275ms: analyse
84ms: phases234
80ms: treeshake
623ms: --GENERATE--
165ms: render modules
455ms: sourceMap
[00:34:21] Finished 'app' after 1.39 s
[00:34:30] Starting 'app'...
634ms: --BUILD--
0ms: transform
18ms: ast
264ms: analyse
89ms: phases234
104ms: treeshake
695ms: --GENERATE--
300ms: render modules
391ms: sourceMap
[00:34:31] Finished 'app' after 1.43 s
[00:34:43] Starting 'app'...
537ms: --BUILD--
0ms: transform
8ms: ast
191ms: analyse
88ms: phases234
68ms: treeshake
556ms: --GENERATE--
161ms: render modules
392ms: sourceMap
[00:34:44] Finished 'app' after 1.18 s
[00:34:46] Starting 'app'...
529ms: --BUILD--
0ms: transform
24ms: ast
199ms: analyse
81ms: phases234
65ms: treeshake
552ms: --GENERATE--
159ms: render modules
390ms: sourceMap
[00:34:47] Finished 'app' after 1.17 s
[00:34:48] Starting 'app'...
521ms: --BUILD--
0ms: transform
7ms: ast
184ms: analyse
88ms: phases234
54ms: treeshake
555ms: --GENERATE--
156ms: render modules
396ms: sourceMap
[00:34:49] Finished 'app' after 1.17 s
[00:34:50] Starting 'app'...
597ms: --BUILD--
0ms: transform
13ms: ast
228ms: analyse
104ms: phases234
91ms: treeshake
826ms: --GENERATE--
364ms: render modules
400ms: sourceMap
[00:34:52] Finished 'app' after 1.51 s
```

src/Module.js
```diff
@@ -55,18 +69,15 @@ export default class Module {
 			this.magicString.remove( match.index, match.index + match[0].length );
 		}
 
-		this.comments = [];
-		this.ast = ast;
-		this.statements = this.parse();
-
 		this.declarations = blank();
+		this.type === 'Module'; // TODO only necessary so that Scope knows this should be treated as a function scope... messy
```
Victorystick: Should this really be `===`?

snuggs: If you're strong type checking then 👍 @Victorystick

Rich-Harris: Ha, yeah – should be `this.type = 'Module'`. Fixed

Rich-Harris:
Have released this as 0.35 so that we have a solid base from which to start optimising and addressing any lingering bugs.

@eventualbuddha 

> I just tried to verify that your fix caused `set` not to appear, but it still does

I think I have an understanding of why this is happening now. It's because the function expression references both itself and a function with an unknown value (`MemberExpression` nodes don't currently know what their possible values are – that's something for the near future), so when in the final stage of tree-shaking Rollup asks all the expressions that *could potentially* have relevant effects (`AssignmentExpression`, `UpdateExpression` and `CallExpression` nodes) whether they do, it gets confused.

The fix I envisage is something I was considering separately as a performance optimisation – only enhancing and initialising nodes as necessary. Ideally, as far as Rollup is concerned, in the final tree-shaking stage that declaration should look like this:

```js
var set = function set ( /* who cares */ ) {
  /* who cares */
};
```

Unless `set` is used by something else there's really no need to traverse that deep.

@TrySound 

Thanks, that's useful. How are you running those numbers? Wonder if it's something we should have in the library all the time (perhaps behind `if ( DEBUG ) ...`, if it has a noticeable impact on the thing its measuring...). Seems like sourcemap generation is low hanging fruit, perhaps we can cache some of that stuff more effectively.

TrySound:
@Rich-Harris I use this, require it and add points every time when code is changed.
https://gist.github.com/TrySound/72caea2645bc6bc9313b97c14b75480f

PR #904, opened 2016-09-07
Faster indexing for key validation

hzlmn:
Old solution run `indexOf` for each key which is O(n^keys length). Instead iterate only once - create `Map` object and then index element with O(1). Could be proved by bench.

### DIFF
--- a/src/utils/arrayToMap.js
+++ b/src/utils/arrayToMap.js
@@ -1,0 +1,6 @@
+export function arrayToMap (list) {
+	return list.reduce((accum, el) => {
+		accum[el] = el;
+		return accum;
+	}, {});
+}
--- a/src/utils/validateKeys.js
+++ b/src/utils/validateKeys.js
@@ -1,14 +1,16 @@
 import { keys } from './object.js';
+import { arrayToMap } from './arrayToMap';
 
 export default function validateKeys ( object, allowedKeys ) {
 	const actualKeys = keys( object );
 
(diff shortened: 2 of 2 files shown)


Rich-Harris:
Thanks for the PR! This is actually one of those interesting cases where intuition about performance isn't quite right – the overhead of creating the map blows the gains out of the water. I've just run the numbers myself, and on Chrome at least (i.e. V8, i.e. Node) the map version is *never* faster – in fact, the `array.indexOf` version is consistently at least twice as fast, whether the functions are hot or cold. The only way the map version comes close to the array version is if you generate the map once and re-use it, but even then it's still slower according to my hastily-produced test: https://gist.github.com/Rich-Harris/963b0bb31151fa1039a6e427c16061a9

Internally, accessing properties of an object isn't actually an O(1) operation if the object is larger than a certain size (I think it's 4 properties, though I'm not an expert). 

Either way, key validation takes a small fraction of a millisecond so it's better not to have the extra code! If you see any other optimisation opportunities though please let us know 😀

hzlmn:
@Rich-Harris interesting, basically in our case keys are cached and I got it about 4 times faster at least in Node env. But I got Your point.

```shell
validateKeysWithArray: 1.101ms
validateKeysWithMap: 0.231ms
validateKeysWithArray 10000 times: 4.421ms
validateKeysWithMap 10000 times: 45.615ms
validateKeysWithCachedMap 10000 times: 1.782ms
```

PR #913, opened 2016-09-10
include dependencies in bundle.modules

### LINKED ISSUE #903: Module Tree output
Would it be possible for rollup to print out a tree of all the modules used and how they are related? Reason for the request is that I have a project where certain dependencies are still getting included, but having difficulty tracking down "why", and thought a module graph or tree would help pinpoint the culprit. Thanks

Rich-Harris:
closes #903

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -315,6 +315,7 @@ export default class Module {
 	toJSON () {
 		return {
 			id: this.id,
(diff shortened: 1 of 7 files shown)


Rich-Harris:
Thanks @TrySound 😀

PR #914, opened 2016-09-10
update to acorn 4

Rich-Harris:


snuggs:
DAMN you're fast @Rich-Harris couldn't even submit the PR. LMAO. Are we working on async/await yet? i've discovered the change we need to make. Looks like merely update to verson 8

Rich-Harris:
@snuggs oh, sweet – didn't know Acorn had support for that already. Yeah, I don't see any reason not to default to `"ecmaVersion": 8` – resulting code won't actually *run* anywhere, but presumably any devs using ES2017 features intend to transpile at some point after the bundle is generated. Can't see any obvious downside to allowing it

snuggs:
Copy that @Rich-Harris. I'll throw it in speedy gonzales. ;-)

And Yeah I been knee deep in the Acorn code to get async/await in rollup. My colleague @btakita said we get async/await + generators in it's a wrap. Full mutany from "them". 

image

PR #922, opened 2016-09-11
Fix "process.hrtime() only accepts an Array tuple." error.

Permutatrix:
I was upgrading the version of Rollup used by gulp-rollup when suddenly almost all of my tests started failing on Travis with the error `process.hrtime() only accepts an Array tuple.`.

Turns out Rollup had been updated again while I was working, introducing this error. Tracking that down was fun. So here's a fix&mdash;enjoy.

### DIFF
--- a/src/utils/flushTime.js
+++ b/src/utils/flushTime.js
@@ -1,7 +1,8 @@
 const DEBUG = false;
 const map = new Map;
 
-let time, toMilliseconds;
+let time;
+let toMilliseconds;
 
 if ( typeof process === 'undefined' ) {
 	time = function time ( previous ) {
@@ -11,7 +12,7 @@ if ( typeof process === 'undefined' ) {
 
 	toMilliseconds = function toMilliseconds ( time ) {
 		return time;
-	}
+	};
 } else {
 	time = function time ( previous ) {
 		return previous === undefined ? process.hrtime() : process.hrtime( previous );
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thanks! That's weird – what version of Node is that with? On Node 6, `process.hrtime(undefined)` works exactly as expected (i.e. the same as `process.hrtime()`.

Rather than redefining `toMilliseconds`, how about just changing [this function](https://github.com/Permutatrix/rollup/blob/55c23a3f23a8047b1dfee8186373949fbcb9bde8/src/utils/flushTime.js#L12-L17) to something like this?

```js
time = function time ( previous ) {
	if ( previous ) {
		const hrtime = process.hrtime( previous );
		return hrtime[0] * 1e3 + hrtime[1] / 1e6;
	}

	return process.hrtime();
};
```

Permutatrix:
It's Node 4.

It seems rather odd to me to have one function do pretty much unrelated things depending on whether or not you pass it an argument. I'm not sure what to call the split-up `time` function without sounding redundant with `timeStart` and `timeEnd`... I think `timeStartHelper` and `timeEndHelper` would be descriptive.

Permutatrix:
Alright, I split up `time` using those names.

Rich-Harris:
LGTM, thanks 👍 

> It seems rather odd to me to have one function do pretty much unrelated things depending on whether or not you pass it an argument

I agree, but the Node.js people apparently don't 😀 I guess we were just following their lead

Rich-Harris:
Released 0.35.7

PR #926, opened 2016-09-11
Correctly deshadow re-assigned module functions

### LINKED ISSUE #910: 0.35.0 - Name collisions
I've got a project that uses Three.js (been building fine in several prior rollup versions) and it's now failing in a browser as at line 40907 is:

```
Object.defineProperties( Material.prototype, {
	wrapAround: {
		get: function () {
			console.warn( 'THREE.' + this.type + ': .wrapAround has been removed.' );
		},
		set: function ( value ) {
			console.warn( 'THREE.' + this.type + ': .wrapAround has been removed.' );
		}
	},
	wrapRGB: {
		get: function () {
			console.warn( 'THREE.' + this.type + ': .wrapRGB has been removed.' );
			return new Color();
		}
	}
} );
```

This is failing because `Material` was most recently defined as:

```
var Material = Object.freeze({
	ShadowMaterial: ShadowMaterial,
	SpriteMaterial: SpriteMaterial,
	RawShaderMaterial: RawShaderMaterial,
	ShaderMaterial: ShaderMaterial,
	PointsMaterial: PointsMaterial,
	MultiMaterial: MultiMaterial,
	MeshPhysicalMaterial: MeshPhysicalMaterial,
	MeshStandardMaterial: MeshStandardMaterial,
	MeshPhongMaterial: MeshPhongMaterial,
	MeshNormalMaterial: MeshNormalMaterial,
	MeshLambertMaterial: MeshLambertMaterial,
	MeshDepthMaterial: MeshDepthMaterial,
	MeshBasicMaterial: MeshBasicMaterial,
	LineDashedMaterial: LineDashedMaterial,
	LineBasicMaterial: LineBasicMaterial,
	Material: Material
});
```

Note that `Material` itself has a property referring to the earlier `Material` (a function).

It would seem this is meant to be a transpiled version of:

```
export { ShadowMaterial } from './ShadowMaterial.js';
export { SpriteMaterial } from './SpriteMaterial.js';
export { RawShaderMaterial } from './RawShaderMaterial.js';
export { ShaderMaterial } from './ShaderMaterial.js';
export { PointsMaterial } from './PointsMaterial.js';
export { MultiMaterial } from './MultiMaterial.js';
export { MeshPhysicalMaterial } from './MeshPhysicalMaterial.js';
export { MeshStandardMaterial } from './MeshStandardMaterial.js';
export { MeshPhongMaterial } from './MeshPhongMaterial.js';
export { MeshNormalMaterial } from './MeshNormalMaterial.js';
export { MeshLambertMaterial } from './MeshLambertMaterial.js';
export { MeshDepthMaterial } from './MeshDepthMaterial.js';
export { MeshBasicMaterial } from './MeshBasicMaterial.js';
export { LineDashedMaterial } from './LineDashedMaterial.js';
export { LineBasicMaterial } from './LineBasicMaterial.js';
export { Material } from './Material.js';
```

from `src/materials/material.js` - although I'm not sure how it ended up with the name `Material`.

0.35.0 seems like a big step in the right direction with regard to future-proofing and is great work in general (Thanks!). I'm so sorry to be dropping bug reports like this!

nolanlawson:
This fixes #910. I've added a failing test that succeeds once the fix is in place. I also tested this PR on PouchDB, and it fixes https://github.com/pouchdb/pouchdb/pull/5673#issuecomment-246186805 .

Apparently what was happening here was that functions re-assigned upon import were not causing proper deshadowing in other variables inside of subfunctions. So in the test, this:

```js
// problematicFunc.js
function problematicFunc() {
  return 'foobar';
}
export { problematicFunc };
```

```js
// main.js
import { problematicFunc as otherFunc } from './problematicFunc';
function innerFunc() {
  function problematicFunc () {
    return otherFunc();
  }
  return problematicFunc();
}
var res = innerFunc();
export default res;
```

Would get (incorrectly) compiled to:

```js
function problematicFunc() {
  return 'foobar';
}
function innerFunc() {
  function problematicFunc () {
    return problematicFunc();
  }
  return problematicFunc();
}
var res = innerFunc();
export default res;
```

In the example above, this would cause `problematicFunc()` to spin indefinitely.

After this fix, the compiled code becomes:

```js
function problematicFunc() {
  return 'foobar';
}
function innerFunc() {
  function problematicFunc$$1 () {
    return problematicFunc();
  }
  return problematicFunc$$1();
}
var res = innerFunc();
export default res;
```

### DIFF
--- a/src/ast/scopes/ModuleScope.js
+++ b/src/ast/scopes/ModuleScope.js
@@ -19,14 +19,15 @@ export default class ModuleScope extends Scope {
 		forOwn( this.module.imports, specifier => {
 			if ( specifier.module.isExternal ) return;
 
-			if ( specifier.name === '*' ) {
-				specifier.module.getExports().forEach( name => {
-					names.set( name, true );
-				});
-			} else {
+			specifier.module.getExports().forEach( name => {
+				names.set(name);
+			});
+			if ( specifier.name !== '*' ) {
 				const declaration = specifier.module.traceExport( specifier.name );
 				const name = declaration.getName( true );
-				if ( name !== specifier.name ) names.set( declaration.getName( true ) );
+				if ( name !== specifier.name ) {
+					names.set(declaration.getName( true ));
+				}
 			}
 		});
 
--- a/test/function/namespacing-in-sub-functions/_config.js
+++ b/test/function/namespacing-in-sub-functions/_config.js
@@ -1,0 +1,8 @@
+var assert = require( 'assert' );
+
+module.exports = {
+	description: 'correctly namespaces sub-functions (#910)',
+	exports: function ( exports ) {
+		assert.equal( exports, 'foobar' );
+	}
+};
--- a/test/function/namespacing-in-sub-functions/main.js
+++ b/test/function/namespacing-in-sub-functions/main.js
@@ -1,0 +1,11 @@
+import { problematicFunc as otherFunc } from './problematicFunc';
+function innerFunc() {
+	function problematicFunc () {
+		return otherFunc();
+	}
+	return problematicFunc();
+}
+
+var res = innerFunc();
+
+export default res;
--- a/test/function/namespacing-in-sub-functions/problematicFunc.js
+++ b/test/function/namespacing-in-sub-functions/problematicFunc.js
@@ -1,0 +1,5 @@
+function problematicFunc() {
+	return 'foobar';
+}
+
+export { problematicFunc };


src/ast/scopes/ModuleScope.js
```diff
@@ -19,14 +19,15 @@ export default class ModuleScope extends Scope {
...
-			} else {
+			specifier.module.getExports().forEach( name => {
+				names.set(name);
+			});
+			if ( specifier.name !== '*' ) {
 				const declaration = specifier.module.traceExport( specifier.name );
 				const name = declaration.getName( true );
-				if ( name !== specifier.name ) names.set( declaration.getName( true ) );
+				if ( name !== specifier.name ) {
+					names.set(declaration.getName( true ));
```
eventualbuddha: nit: there should be spaces inside the parentheses

*(why isn't there a lint rule for this?)*

Rich-Harris: > *(why isn't there a lint rule for this?)*

There isn't a combination of spacing rules in ESLint that doesn't result in weirdness, I've found. Good catch – will fix locally now that I've hastily merged it

nolanlawson: Sorry about that! Yeah, I just ran `npm run eslint` and assumed everything was kosher.

eventualbuddha: > There isn't a combination of spacing rules in ESLint that doesn't result in weirdness, I've found.

I just noticed that as well when trying to enable `space-in-parens`. I might take a stab at writing a custom rule…

PR #927, opened 2016-09-11
Reinstate support for Node 0.12

Rich-Harris:
Emoji reactions please... Turns out this wasn't at all hard to do, and has minimal impact on the build. So I'm 👍 if it makes life easier for Ember folks.

Ember is dropping support for 0.12 on Jan 1 2017, so we could probably do likewise then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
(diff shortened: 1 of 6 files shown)


snuggs:
BRUH you can't tell me it was this simple. We're really on to something here @Rich-Harris

PR #933, opened 2016-09-13
Only remove EmptyStatement nodes directly inside blocks.

### LINKED ISSUE #931: Empty statements unexpectedly removed.
I expected the following code to be unchanged by Rollup:

```js
var result = 0;
(function(test) {
  if (test === 0);
  else ++result;
})(0);
console.log(result);
```

However, the resulting output drops the semicolon from the if block, resulting in invalid JavaScript:

```js
var result = 0;
(function(test) {
  if (test === 0)
  else ++result;
})(0);
console.log(result);
```

```
  else ++result;
  ^^^^
SyntaxError: Unexpected token else
```

You can workaround this issue by using an empty block instead of an empty statement:

```js
var result = 0;
(function(test) {
  if (test === 0) {}
  else ++result;
})(0);
console.log(result);
```

Naturally, the problem also applies to else-if and else:

```js
var result = 0;
(function(test) {
  if (test === 0) {}
  else if (test === 1);
  else;
})(0);
console.log(result);
```

Which results in:

```js
var result = 0;
(function(test) {
  if (test === 0) {}
  else if (test === 1)
  else
})(0);
console.log(result);
```

```
  else
  ^^^^
SyntaxError: Unexpected token else
```

This example is slightly more complicated than it should be (using the IIFE) because #930 was causing the simpler test case to be removed entirely.

eventualbuddha:
When used as part of another statement, such as an `IfStatement`, they can be semantically meaningful.

Fixes #931


eventualbuddha:
Released in v0.35.10.

PR #934, opened 2016-09-13
Correctly deshadow star imports when conflicting

nolanlawson:
As @Benjamin-Dobell pointed out in https://github.com/rollup/rollup/issues/910#issuecomment-246220533, there's another case where deshadowing was not working correctly.

If you do `import * as Foo` in one place, and then `import { Foo }` in another, then those two `Foo` declarations aren't checked for namespace collisions. In fact, the problem is with the `import * as Foo`, because `Foo` doesn't get added to the list of usages and therefore isn't checked when doing `getSafeName()`.

I fixed that issue; the two new added tests fail before the fix but succeed after. (In fact they throw an error if you try to execute them.)  Unfortunately there's an aesthetic regression in this PR in that it can too-aggressively deshadow, as shown by the fact that I had to change one of the `form` tests to use `foo$1()` instead of `foo()` even though the `$1` isn't necessary.

However, since this PR fixes the functionality at the expense of making the output a little less pretty, I'm still offering it up. Please let me know if there's a good way to fix the `form` regression; I couldn't find one.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -208,6 +208,15 @@ export default class Bundle {
 
 				declaration.name = getSafeName( declaration.name );
 			});
+			// special case - for `import * as Foo`, we need to make sure that Foo
+			// gets its own variable because it will eventually be rendered as
+			// `var Foo = Object.freeze(...)`
+			forOwn( module.imports, ( importee, importeeName ) => {
+				if ( importee.name === '*' ) {
+					delete module.imports[ importeeName ];
+					module.imports[ getSafeName(importeeName) ] = importee;
+				}
+			});
 		});
 
 		this.scope.deshadow( toDeshadow );
--- a/test/function/namespacing-collisions-2/main.js
+++ b/test/function/namespacing-collisions-2/main.js
@@ -1,0 +1,7 @@
+import { Something } from './Something';
+import { SomethingAgain } from './SomethingAgain';
+import { Material } from './Material';
+import { MaterialAgain } from './MaterialAgain';
+
+var result = [Material(), MaterialAgain(), Something(), SomethingAgain()]
+export default result;
--- a/test/form/namespace-optimization-b/_expected/amd.js
+++ b/test/form/namespace-optimization-b/_expected/amd.js
@@ -1,12 +1,12 @@
 define(function () { 'use strict';
 
-	function foo () {
+	function foo$1 () {
 		console.log( 'foo' );
 	}
 
 	function a () {
-		foo();
-		foo();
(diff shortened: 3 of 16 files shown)


src/Bundle.js
```diff
@@ -208,6 +208,15 @@ export default class Bundle {
 
 				declaration.name = getSafeName( declaration.name );
 			});
+			// special case - for `import * as Foo`, we need to make sure that Foo
+			// gets its own variable because it will eventually be rendered as
+			// `var Foo = Object.freeze(...)`
+			forOwn( module.imports, ( importee, importeeName ) => {
+				if ( importee.name === '*' ) {
+					delete module.imports[ importeeName ];
+					module.imports[ getSafeName(importeeName) ] = importee;
```
nolanlawson: there is probably a more elegant way to do this than deleting the key and re-adding it, but I couldn't think of one.

test/function/namespacing-collisions-2/main.js
```diff
@@ -0,0 +1,7 @@
+import { Something } from './Something';
+import { SomethingAgain } from './SomethingAgain';
+import { Material } from './Material';
+import { MaterialAgain } from './MaterialAgain';
+
+var result = [Material(), MaterialAgain(), Something(), SomethingAgain()]
```
snuggs: Is there a particular reason we are choosing `var` over `const` @nolanlawson ?

nolanlawson: I don't really use const/let in most of the code I write (I know, go find the pitchforks 😛) and this is just test code, so I figured it was okay. We could apply the eslint rules to the test code if you'd like!

eventualbuddha: Since rollup supports 0.12 (again) we have to use buble on anything that doesn't run natively. Most of the tests use `var` if I remember correctly.

snuggs: OH RIGHT `0.12` 👿  Le sigh @eventualbuddha @nolanlawson

Rich-Harris:
🎉 thanks @nolanlawson! I made one small change (https://github.com/rollup/rollup/commit/bed6f0650da4f54ab8b8f7b4e3d64d99a12c922c) before merging – we only need to deconflict *reified* namespaces, which means that the `foo$1` in the changed test can safely stay as `foo`

nolanlawson:
OK, that's great! Your version looks much tidier than mine, and I'm happy the `foo$1` issue could be fixed. Thanks for the quick merge!

Issue #939, opened 2016-09-14
omit getInteropBlock


emilos:
Hey,

thanks for the great tool. I'm trying to include rollup in a bigger app and I'd like to rewrite the files to es6 step by step. I'd like to start with views first, to transpile the files one by one, and later migrate more files to finally end up with a bundle. I was wondering if the following would make sense:

Here's a simplified case, given this input:
```javascript
import View from 'core/view';
export default View.extend({});
```
I'd expect to have the following result:

```javascript
define(['core/view'], function (View) {
      'use strict';

      return View.extend({});
});
```
however the result I'm getting right now is:
```javascript
define(['core/view'], function (View) { 'use strict';
    View = 'default' in View ? View['default'] : View;
    var main = View.extend({});
    return main;
});
````

I'd like to omit the `getInteropBlock` step, to generate the code from the example 1 if possible for now. Would that make any sense or does that have some downsides? I'll happily create a PR if it makes sense.

Rich-Harris:
I could get behind an `interop: false` option. A PR would be great, thank you!

There's a separate issue for the second part (avoiding the `return main` thing): https://github.com/rollup/rollup/issues/826

PR #945, opened 2016-09-16
add: interop option

emilos:
Hey,

https://github.com/rollup/rollup/issues/939

I've added the interop option, so that there's a possibility to disable it if necessary. I have three small concerns:

1) I needed to add a `/*eslint import/no-unresolved: 0*/` comment to the `interop-false/main.js` file to mark the dependency as unresolved, but that doesn't seems to be pretty and causes the comment to be in the result files, is there anything we can do to remove it in this case?

2) the interop works a little bit differently in the `src/finalisers/cjs.js` file and the implementation is split into few lines, do you have any suggestions for this part? It would be great if you could provide me with some background on this one.

3) should I add it to the docs? What's the best place?

Best,
Emil

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 [these people](https://github.com/rollup/rollup/graphs/contributors)
+Copyright (c) 2016 [these people](https://github.com/rollup/rollup/graphs/contributors)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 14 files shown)


Rich-Harris:
Thanks, this is great! The ESLint comments shouldn't be necessary because of the `.eslintrc` file in the project root – as long as it passes CI, we're copacetic. I removed them before merging because it's best to keep test files as noise-free as possible. In the CJS file, it's just a case of avoiding the first branch, [like so](https://github.com/rollup/rollup/commit/303333b31019ad92d7bda54ad2a558e126ed9d6e) – as long as it doesn't think it has to worry about default exports, it won't treat them specially.

I took the liberty of updating the docs while I'm here – https://github.com/rollup/rollup/wiki/JavaScript-API#interop. Thanks for bringing that up, I always forget docs 😁 

👍

Rich-Harris:
Oh and this has been released as 0.35.12

Issue #587, opened 2016-04-03
default export handling in CJS and IIFE bundles


bfred-it:
I'm sure this has been talked about before but I can't find it. :-/

The way `default` exports behave when there are additional exports is counterintuitive.

```js
export default function () {}
```

Becomes in IIFE and CJS respectively:

```js
var doStuff = (function () {
	'use strict';
	function index () {}
	return index;
}());
```
```js
'use strict';
function index () {}
module.exports = index;
```

So far so good, even though I wish the code was shorter.

But then when adding a named export:

```js
export default function () {}
export function other() {}
```

Becomes:

```js
'use strict';

function index () {}
function other() {}

exports['default'] = index; // Boom, broken
exports.other = other;
```
```js
(function (exports) {
	'use strict';

	function index () {}
	function other() {}

	exports['default'] = index; // Boom, broken
	exports.other = other;

}((this.doStuff = this.doStuff || {}))); // Also, why this check only here? In some cases the behavior changes completely.
```

---

The core problem is that the if with a single export I can just use
```js
window.doStuff()
```

adding another one it becomes an unintuitive and unideal
```js
window.doStuff['default']()
```

Same goes for the CJS export, which without [this issue](https://github.com/rollup/rollup/issues/496) sorted out, a browserify user would have to use 

```js
var doStuff = require('do-stuff')['default'];
```

Or even

```js
import doStuff from 'do-stuff';
doStuff['default']();
```

---

rollup is an amazing tool to [bundle libraries](https://github.com/frostney/rollup-babel-lib-bundler), but using proper ESM exports with it is not advised and defeats the purpose of the ES6-style rollup output, because I have to do this instead:

```js
function doStuff() {}
Object.defineProperty(doStuff, "other", {
	value:  function () {}
});
export default doStuff;
```

I understand the reasoning behind it, because this isn't possible for example:

```js
var value = 5;
Object.defineProperty(value, "other", {
	value:  function () {}
});
export default value;
```

But it still remains a big issue for, what I believe is, the most common use case (i.e. exporting functions and objects as default). For others perhaps rollup could revert to the current behavior and output a warning.

Victorystick:
@bfred-it I get what you're saying, and I feel your pain, but I don't understand if you have any concrete suggestions for improvements. Can you elaborate? Or do you just want to have an open discussion about these issues?

bfred-it:
[...]

PR #966, opened 2016-09-19
Avoid exports:auto warning on format:es

bfred-it:
This improves on https://github.com/rollup/rollup/issues/587#issuecomment-216528129


Rich-Harris:
Thanks! Released as 0.36.1

PR #967, opened 2016-09-19
Include default exports used within the module but not by importers.

Permutatrix:
[They're currently omitted.](http://rollupjs.org/?shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybW9kdWxlTmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMm1vZHVsZXMlMjIlM0ElNUIlN0IlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjAlN0IlMjBiJTIwJTdEJTIwZnJvbSUyMCcuJTJGbW9kdWxlLmpzJyUzQiU1Q24lNUNuY29uc29sZS5sb2coYigpKSUzQiU1Q24lMjIlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIybW9kdWxlLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMGEoKSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZygneCcpJTNCJTVDbiUyMCUyMHJldHVybiUyMDUlM0IlNUNuJTdEJTVDbiU1Q25hKCklM0IlNUNuJTVDbmV4cG9ydCUyMGZ1bmN0aW9uJTIwYigpJTIwJTdCJTVDbiUyMCUyMHJldHVybiUyMGEoKSUyMCUyQiUyMDEwJTNCJTVDbiU3RCU1Q24lMjIlN0QlNUQlN0Q=)

PR #972, opened 2016-09-21
Fix for #970.

### LINKED ISSUE #970: Broken bundle generated for D3
I'm trying to generate a custom bundle of D3 v4 using rollup:

https://github.com/Turbo87/rollup-d3-issue

Running `npm run build` (or `rollup -c`) will produce a bundled file at `dist/d3.js` which looks fine on first glance. It includes the following function call:

``` js
defaultLocale({
  decimal: ".",
  thousands: ",",
  grouping: [3],
  currency: ["$", ""]
});
```

but the `defaultLocale()` function is not defined anywhere in the bundle even though it's in the original code (at https://github.com/d3/d3-format/blob/v1.0.2/src/defaultLocale.js).

Note that if I add `export { formatDefaultLocale } from 'd3-format';` to the `d3.js` file we can work around this issue. This is not a good long-term solution though.

Cross-posted this as https://github.com/d3/d3-format/issues/28

mbostock:
Fixes #970.

### DIFF
--- a/src/ast/nodes/ExportDefaultDeclaration.js
+++ b/src/ast/nodes/ExportDefaultDeclaration.js
@@ -76,8 +76,12 @@ export default class ExportDefaultDeclaration extends Node {
 			super.render( code, es );
 		} else {
 			if ( treeshake ) {
-				if ( functionOrClassDeclaration.test( this.declaration.type ) && !this.declaration.activated ) {
-					code.remove( this.leadingCommentStart || this.start, this.next || this.end );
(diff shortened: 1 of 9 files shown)


mbostock:
My fix is probably wrong but the tests pass. 😕 My thought was that if a default export’s declaration is activated, it means we have to include the declaration in the output (but not the export)? /cc @Turbo87

Permutatrix:
My pull request #967 addresses this as well. I'm not sure why it isn't merged yet.

mbostock:
Closing in favor of #967.

PR #989, opened 2016-09-26
ADD: fallback for Object.freeze

futurist:
gracefully fallback for `Object.freeze` to make it more compatible in browser that don't support it.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -86,7 +86,7 @@ export class SyntheticNamespaceDeclaration {
 			return `${indentString}${name}: ${original.getName( es )}`;
 		});
 
-		return `${this.module.bundle.varOrConst} ${this.getName( es )} = Object.freeze({\n${members.join( ',\n' )}\n});\n\n`;
+		return `${this.module.bundle.varOrConst} ${this.getName( es )} = (Object.freeze || Object)({\n${members.join( ',\n' )}\n});\n\n`;
 	}
 }
 
--- a/test/form/prefer-const/_expected/amd.js
+++ b/test/form/prefer-const/_expected/amd.js
@@ -4,7 +4,7 @@ define(['external', 'other', 'another'], function (external, other, another) { '
 	const b = 2;
 
 
-	const namespace = Object.freeze({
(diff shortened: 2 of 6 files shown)


Rich-Harris:
Thanks for the PR @futurist – I'm wondering if maybe this should be behind an option? Otherwise everyone has to suffer for IE8's sins. I *think* there might be one or two other places where Rollup's output doesn't support IE8 (actually can't remember off the top of my head), so if there were a `legacy: true` option we'd be able to reuse it.

That said it is an open question whether we should support IE8 at all... it might be better to use the `intro` or `banner` option to include a polyfill, rather than adding more stuff to Rollup's codebase. Would welcome other people's thoughts.

Victorystick:
In [some situations](http://rollupjs.org/#%7B%22options%22%3A%7B%22format%22%3A%22cjs%22%2C%22moduleName%22%3A%22myBundle%22%2C%22globals%22%3A%7B%7D%7D%2C%22modules%22%3A%5B%7B%22name%22%3A%22main.js%22%2C%22code%22%3A%22import%20*%20as%20counter%20from%20'.%2Fcounter.js'%3B%5Cn%5Cnexport%20%7B%20counter%20%7D%3B%22%7D%2C%7B%22name%22%3A%22counter.js%22%2C%22code%22%3A%22export%20let%20value%20%3D%200%3B%5Cn%5Cnexport%20function%20increment()%20%7B%5Cn%20%20value%2B%2B%3B%5Cn%7D%22%7D%5D%7D) related to the use of `Object.freeze` Rollup requires that the JS engine supports inline getters, which IE8 does not. We'd have to keep that in mind if we intend to support browsers like IE8 through some `legacy`/`ES3` flag.

mourner:
@Rich-Harris an option like this would be very valuable to us at https://github.com/Leaflet/Leaflet, because we still support IE8+ (and a ton of developers rely on this), and adding polyfills may be an option for apps, but it's not very viable for library authors that want to keep it as unobtrusive as possible.

Rich-Harris:
Have added this behind `options.legacy` and released as 0.36.3 – thanks

PR #999, opened 2016-09-29
reduce instead forEach to update options 

vforvova:
Hi,

I explored the repo and I found serial `options` updates  by plugins in `Bundle` constructor. I suppose this case is more suitable for `reduce`, not for `forEach`. With `reduce` `options` will be rewritten once.

I think this code looks more declarative and shows that meaning of this block is more about `options` update and less about `plugins` iteration. As side effect there is performance micro-optimisation.

/cc @Rich-Harris @TrySound @Victorystick

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -31,11 +31,10 @@ export default class Bundle {
 
 		this.plugins = ensureArray( options.plugins );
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
thanks 👍

PR #1012, opened 2016-10-01
Add module id when warning about top level this

FredyC:
When during a bundling I get a cryptic message stating `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten`, it's not very helpful to find actual problematic module especially since it can occur anywhere inside node_modules.

Adding module ID to the message should help locate and possibly remove the problem.

### DIFF
--- a/src/ast/nodes/ThisExpression.js
+++ b/src/ast/nodes/ThisExpression.js
@@ -7,7 +7,7 @@ export default class ThisExpression extends Node {
 		if ( lexicalBoundary.isModuleScope ) {
 			this.alias = this.module.bundle.context;
 			if ( this.alias === 'undefined' ) {
-				this.module.bundle.onwarn( 'The `this` keyword is equivalent to `undefined` at the top level of an ES module, and has been rewritten' );
+				this.module.bundle.onwarn( `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten (in ${this.module.id})` );
 			}
 		}
 	}
--- a/src/ast/nodes/ThisExpression.js
+++ b/src/ast/nodes/ThisExpression.js
@@ -1,4 +1,5 @@
 import Node from '../Node.js';
+import getLocation from '../../utils/getLocation.js';
 
 export default class ThisExpression extends Node {
 	initialise ( scope ) {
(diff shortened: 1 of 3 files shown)


Victorystick:
Good idea! Could you include the line number as well?

FredyC:
I've fixed tests. Did not really expect there will be one checking exact format of the message. Had to do it more precisely instead of checking for a fixed message because the path to the file is dynamic. I hope it's enough like that.

@Victorystick That would be lovely. I tried looking at what data are available there, but could not find information about line number there :(

Victorystick:
To get the line and column location of an AST node just use `utils/getLocation`! :+1:

FredyC:
@Victorystick Ok I did try it like this, but I don't think it's correct. It returns `line 4` in test, but it should be line 3. I am not sure what other number then `this.module.ast.end` use.

src/ast/nodes/ThisExpression.js
```diff
@@ -7,7 +8,9 @@ export default class ThisExpression extends Node {
 		if ( lexicalBoundary.isModuleScope ) {
 			this.alias = this.module.bundle.context;
 			if ( this.alias === 'undefined' ) {
-				this.module.bundle.onwarn( `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten (in ${this.module.id})` );
+				const location = getLocation(this.module.code, this.module.ast.end);
```
Victorystick: All AST nodes have `start` and `end` properties. Use `this.start` instead of `this.module.ast.end`. (That'll always give you the last line of the bundle.)

FredyC: That doesn't seem about right because the `start` seems to be always zero in the test. Even if the actual problem occurred one line 4.

FredyC: I did try now using either `this.start` or `this.module.start`, both are undefined.

Rich-Harris:
Thanks! I fixed the location thing, and tweaked the message a bit so that it includes a link to https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined. 

> Had to do it bit differently instead of checking for a fixed message because the path to the file is dynamic

For #1033 I added a utility, `relativeId`, that takes a module ID and makes it relative to the CWD if possible, which fixes that problem and makes the message more readable – I've incorporated it here.

Rich-Harris:
Released as 0.36.2

FredyC:
Thanks @Rich-Harris. So you think that using `module.ast.end` is correct to find line number? It doesn't work that well as test file apparently has issue at a different spot than the one recovered by this. Might need some tweaking.

Rich-Harris:
> So you think that using module.ast.end is correct to find line number?

No, I changed it to use `this.start` – https://github.com/rollup/rollup/commit/40400224431729bec1970a32c0b78bb694094ccf

FredyC:
Ah I've missed that. Looks great, amazing.

PR #1013, opened 2016-10-02
Add missing semicolon.

### LINKED ISSUE #1004: Allow importing of modules that don't use semicolons
There is a trend in the JS community to [abandon](http://standardjs.com/rules.html#semicolons) [semi](http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding)-[colons](http://inimino.org/~inimino/blog/javascript_semicolons).

When appending a file that starts with an opening bracket (basically most external dependencies) after a file that relies on not using semi colons, the code breaks:
```js
// included from node_mobule/baz/no_semi_colons.js
const someString = "bla"
export default someString

// included from node_mobule/foo/bar.js
(function (global) {
  // blabla
})(typeof window === 'object' ? window : global);
```
The previous example will result in the following error: `Uncaught TypeError: someString is not a function`

Is there a way to avoid this?

### LINKED ISSUE #1009: custom bundle including d3-hierarchy breaks in 0.36.1
First off, thanks for all the hard work that's been put into this library.

Updating to 0.36.1 has broken my build, which uses [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve) to include the [d3-hierarchy](https://github.com/d3/d3-hierarchy#treemap) treemap generator. Here is a stripped down example repository that reproduces the error: https://github.com/davelandry/rollup-d3

mbostock:
Fixes #1004.
Fixes #1009.

### DIFF
--- a/src/ast/nodes/ExportDefaultDeclaration.js
+++ b/src/ast/nodes/ExportDefaultDeclaration.js
@@ -66,6 +66,7 @@ export default class ExportDefaultDeclaration extends Node {
 						return; // don't render children. TODO this seems like a bit of a hack
 					} else {
 						code.overwrite( this.start, this.declaration.start, `${this.module.bundle.varOrConst} ${name} = ` );
+						if ( code.original[ this.end - 1 ] !== ';' ) code.insertRight( this.end, ';' );
 					}
 				}
 			} else {
@@ -84,6 +85,7 @@ export default class ExportDefaultDeclaration extends Node {
 				}
 			} else {
 				code.overwrite( this.start, this.declaration.start, `${this.module.bundle.varOrConst} ${name} = ` );
+				if ( code.original[ this.end - 1 ] !== ';' ) code.insertRight( this.end, ';' );
 			}
 			// code.remove( this.start, this.next || this.end );
 		}
--- a/test/form/removes-existing-sourcemap-comments/_expected/amd.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/amd.js
@@ -2,7 +2,7 @@ define(function () { 'use strict';
 
 	var foo = function () {
 		return 42;
-	}
+	};
 
 	console.log( foo() );
 
--- a/test/form/removes-existing-sourcemap-comments/_expected/cjs.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/cjs.js
@@ -2,6 +2,6 @@
 
 var foo = function () {
 	return 42;
-}
+};
 
 console.log( foo() );
--- a/test/form/removes-existing-sourcemap-comments/_expected/es.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/es.js
@@ -1,5 +1,5 @@
 var foo = function () {
 	return 42;
-}
+};
 
 console.log( foo() );
--- a/test/form/removes-existing-sourcemap-comments/_expected/iife.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/iife.js
@@ -3,7 +3,7 @@
 
 	var foo = function () {
 		return 42;
-	}
+	};
 
 	console.log( foo() );
 
--- a/test/form/removes-existing-sourcemap-comments/_expected/umd.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/umd.js
@@ -6,7 +6,7 @@
 
 	var foo = function () {
 		return 42;
-	}
+	};
 
 	console.log( foo() );
 


Permutatrix:
I'm not one of the maintainers, but they'll probably ask you to add a test for this. I'd put it under the test/function/ directory, since it fixes an issue in the actual behavior of rolled-up output.

Rich-Harris:
Unfortunately it's not just default exports that have this problem – see e.g. [one](http://rollupjs.org/?version=0.36.1&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwZm9vJTIwJTdEJTIwZnJvbSUyMCcuJTJGZm9vLmpzJyUzQiU1Q24lNUNuKGZ1bmN0aW9uJTIwYmFyKCklMjAlN0IlNUNuJTIwJTIwY29uc29sZS5sb2coJTIwJ2JhciclMjApJTNCJTVDbiU3RCkoKSUzQiU1Q24lNUNuY29uc29sZS5sb2coJTIwZm9vKCklMjApJTNCJTIyJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmZvby5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJleHBvcnQlMjB2YXIlMjBmb28lMjAlM0QlMjBmdW5jdGlvbiUyMCgpJTIwJTdCJTVDbiUyMCUyMHJldHVybiUyMDQyJTNCJTVDbiU3RCUyMiU3RCU1RCU3RA==), [two](http://rollupjs.org/?version=0.36.1&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJy4lMkZmb28uanMnJTNCJTVDbiU1Q24oZnVuY3Rpb24lMjBiYXIoKSUyMCU3QiU1Q24lMjAlMjAlMkYlMkYlMjBsb2wlMjBub3BlJTVDbiU3RCkoKSUzQiUyMiU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJmb28uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyY29uc29sZS5sb2coJTIwJ3RoaXMlMjBleHByZXNzaW9uJTIwaXMlMjBub3QlMjBhJTIwZnVuY3Rpb24nJTIwKSUyMiU3RCU1RCU3RA==). Given that 'you should write broken code and expect the engine to fix it' is an attitude that's actively promoted in the JS community via projects like Standard, we'll need to accommodate those situations as well. Two options:

* Add semi-colons after all expression statements and non-class/function declarations. Easiest to do, will catch all cases, but might be seen as a little passive-aggressive...
* Only add a semi-colon when necessary – i.e. if the last included expression statement or declaration lacks one and the opening character of the next module is `[` or `(` (are there any others? As someone who endeavours to *not* write broken code I don't need to memorise when I need to include semis). More complicated, and easier to end up missing cases, but would pacify the 'don't worry, it's *fiiinnne*, they're optional' crowd.

Hmm.

(Mini-rant: yes, it *is* broken code. ASI means that the engine inserts semi-colons where necessary to form a correct program – in other words, the program without semi-colons is not correct. QED. Fight me.)

mbostock:
Personally I would prefer that any generated code include the semicolons explicitly rather than rely on ASI. So the first option sounds reasonable?

Rich-Harris:
If it has the @mbostock seal of approval, that's good enough for me! Have implemented option 1 in #1037, will close this in favour of that. Thanks

montogeek:
@mbostock Just saw this on my bundle diff, good call!

Issue #179, opened 2015-10-12
Detect whether CallExpressions (may) have side-effects before including them


Rich-Harris:
All `CallExpression` nodes that aren't inside a function are assumed to have side-effects, and automatically included. This behaviour is over-cautious in situations like this:

``` js
// main.js
import { foo } from './foo.js';
foo();

// foo.js
export const foo = () => console.log( 'foo' );

export const Unused = (function () {
  function Unused () {
    this.willNeverRun = true;
  }

  Unused.prototype = {
    runThisCode () {
      console.log( 'jk. this code will never run' );
    }
  };

  return Unused;
})();
```

The `Unused` class is never instantiated, but we included it anyway because it's declared inside an IIFE (i.e a `CallExpression`). This isn't an academic problem, it's how Babel transpiles classes (for example).

Similarly, a call to a function that either a) doesn't have side-effects, or b) only has side-effects that apply to things we don't need, should not be included, and should not cause the relevant declaration to be included:

``` js
export const num = 5;
export const squared = square( num );

function square ( num ) {
  return num * num;
}
```

The function call on the second line means we have to include that line, plus the `num` and `square` declarations, regardless of whether we're importing either `num` or `squared` from this module, even though we can statically determine that `square` has no side-effects.

It may also be possible to determine whether functions are capable of mutating their arguments.

Victorystick:
Statically analyzing JavaScript is made  _very_ difficult by its various features. Take the `square` function for instance, which multiplies `num` with itself. If `num` is a primitive value type, it can't be modified and there should be no side-effects. But if `num` is an object that defines `valueOf` all bets are off since arbitrary code could be executed, twice. There could be both side-effects and mutation.

Victorystick:
If only JavaScript had a way of telling a function was pure...

Rich-Harris:
Oh, I didn't say it'd be easy! I think it's reasonably to be pragmatic though and say that if you've defined `valueOf` to have side-effects, you deserve to have your programming license revoked. (Maybe if `options.paranoid === true` we skip these checks or something.)

If we assume that the developer hasn't poked about with `valueOf` or whatever, we can determine that both the IIFE and `square` are pure. I suspect there are a lot of cases where it's not possible to know if a function is pure (e.g. inside our locally declared `foo` function we call `bar`, imported from an external module, so we have to treat `foo` as though it has side-effects), but there are definitely *some* cases where we can confidently remove code.

mbostock:
[...]

PR #1017, opened 2016-10-03
Removes `new Foo(...)` when Foo is a pure function

Pauan:
This PR adds in the following optimization (which fixes https://github.com/rollup/rollup/issues/179#issuecomment-250903867):

When using `new Foo(...)` and the `Foo` function is "pure", then the `new Foo(...)` is removed.

A function is pure if it meets all of the following conditions:
- It is called with `new Foo(...)`
- The only side effects it contains are of the form `this.foo = bar` or `this[foo] = bar` (the expressions `foo` and `bar` can be anything, but they may not contain side effects)
- It does not have any other side effects
- It is an ES5 function (e.g. not an arrow function)

This detection is currently very limited: it requires the assignment to occur at the top level of the function, and it does not support nested assignment (e.g. `this.foo.bar = qux`)

The detection can be improved later, this just covers the common case.

It only treats the `Foo` function as pure if it is called as `new Foo(...)`. The reason is because `Foo(...)` will cause a runtime error (because it tries to assign to `undefined`), therefore removing the `Foo(...)` would change the behavior of the program.

All of the tests pass, but I had to change the `form/unmodified-default-exports` test, because this optimization (correctly) removed the `Foo` function in that test.

### DIFF
--- a/src/ast/nodes/CallExpression.js
+++ b/src/ast/nodes/CallExpression.js
@@ -27,7 +27,7 @@ export default class CallExpression extends Node {
 	}
 
 	hasEffects ( scope ) {
-		return callHasEffects( scope, this.callee );
+		return callHasEffects( scope, this.callee, false );
 	}
 
 	initialise ( scope ) {
--- a/src/ast/nodes/NewExpression.js
+++ b/src/ast/nodes/NewExpression.js
@@ -3,6 +3,6 @@ import callHasEffects from './shared/callHasEffects.js';
 
 export default class NewExpression extends Node {
 	hasEffects ( scope ) {
-		return callHasEffects( scope, this.callee );
+		return callHasEffects( scope, this.callee, true );
 	}
 }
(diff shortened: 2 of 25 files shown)


Pauan:
@Rich-Harris I don't mean to be pushy, but I've been waiting for over 2 months, and this pull request would help out a lot with PureScript programs.

Rich-Harris:
@Pauan sorry for the delay – been away from this project a while, slowly working through the backlog – and thanks for the PR. Long term I'd like Rollup to have a more complete understanding of how context works in JavaScript, but this is a good solution in the meantime. Released as 0.37.2.

Pauan:
@Rich-Harris Awesome, thanks! I really appreciate all the work that you've done on Rollup, it's very solid. :+1:

I agree that a deeper understanding would be good, in particular understanding the `prototype` and class inheritance would be beneficial.

Pauan:
@Rich-Harris Here is an example of how massively this optimization affects PureScript:

* [Before](https://github.com/purescript/purescript/issues/2207#issuecomment-250905093)
* [After](https://github.com/purescript/purescript/issues/2207#issuecomment-268628306)

That's a 283% reduction in size (from `11.3 kB` to `4.0 kB`)! So, thanks again.

Rich-Harris:
@Pauan that's outstanding! Thanks for sharing those figures, and thanks for the PR!

Issue #1007, opened 2016-09-30
`this` implies global scope in code, but not at run-time


ntilwalli:
More of a question than an issue... I'm trying to bundle `superagent` with rollup which is written in commonjs. It has a module which appends methods directly to the `exports` object which all use `this` directly. Those methods are then mixed-in (copied over) to a different object in other modules. 

In the original module the `this` looks like it's referencing the global scope so rollup sets `this` to `undefined` and at run-time calls to those methods error out because this has been explicitly reset
When I modify the code to assign the methods to a tmp object and assign that to the `exports` and then copy those methods from the that tmp object, since it no longer seems to be referencing the global scope `this` doesn't get assigned to `undefined` and all works fine

The offending module is here: https://github.com/visionmedia/superagent/blob/master/lib/request-base.js

The methods are copied over here: https://github.com/visionmedia/superagent/blob/master/lib/client.js#L524

What's the best approach to deal with this?

ntilwalli:
Is the only way to PR the `superagent` repo so that module has more ES6 conformant module semantics?

ntilwalli:
Would it be possible to add `allowGlobalThisReferences` configuration flag which accepts files where `this` is not set to undefined?

TrySound:
Did you try `context` option?

ntilwalli:
@TrySound Does `context` allow me to set the option for one file only?  I want `this` to be left alone so the functions can be later set on the `prototype` of other functions during run time, where they should have standard `this` semantics...  What would I set `context` to?

Rich-Harris:
Closing via #1023

ahelmberger:
I'm sorry, I tried to get my head around this for two hours, but still didn't get it to work with superagent. My config looks like this:

~~~
const { rollup } = require('rollup');
const nodeResolve = require('rollup-plugin-node-resolve');
const commonjs = require('rollup-plugin-commonjs');
const babel = require('rollup-plugin-babel');

module.exports = {
  entry: 'app/main.js',
  plugins: [
    nodeResolve({ jsnext: true, main: true, browser: true, preferBuiltins: false }),
    commonjs(),
    babel()
  ],
  moduleContext: {
    './node_modules/superagent/lib/request-base.js': 'window'
  },
  targets: [{
    dest: 'dist/main.js',
    format: 'iife',
    sourceMap: true,
    useStrict: true
  }]
};
~~~

[...]

PR #1023, opened 2016-10-05
Add support for module-specific context with 'moduleContext' option

ntilwalli:
This PR is based on the issue raised here: https://github.com/rollup/rollup/issues/1007

I've posted it for comment and possible merge.

@TrySound offered that `options.context` could address this, but I didn't like that the `context` option applies to the whole bundle when I only want to override the context for one specific file/module in the bundle.  This PR adds a module-specific context override option (`moduleContext`) for targeted `this` overriding.  It falls back to the `context` option in case no `moduleContext` entry is given for a given module.  The option accepts an object.  The entries are key-value pairs, where the key is the path relative to the directory of the file given in `options.entry` and the value is the context override value.

Example: 
```
{
  entry: './index.js',
  moduleContext: {
    'node_modules/superagent/blah.js': 'wutwut'
  }
}


Rich-Harris:
Thanks for the PR, this makes sense. Other similar options support functions as well as objects (e.g. `globals`, `external`) so for consistency this should too – have made that change.

Rich-Harris:
Released as 0.36.2

Issue #1010, opened 2016-10-01
Maximum call stack error with cache after error


jo-sm:
I'm building a React app, and if I use the `cache` and get a syntax error, I get a `RangeError: Maximum call stack size exceeded` on the next Rollup attempt. This is the call stack:

```
[2016-10-01T10:31:40.253Z] Hot Swap: SyntaxError: /assets/js/BasicComponent.js: Expected corresponding JSX closing tag for <h1> (5:32)
  3 | export default class BasicComponent extends React.Component {
  4 |   render() {
> 5 |     return <div><h1>Hello world!</h2></div>;
    |                                 ^
  6 |   }
  7 | }
  8 | 
    at Parser.pp.raise (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/location.js:22:13)
    at Parser.pp.jsxParseElementAt (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/plugins/jsx/index.js:462:12)
    at Parser.pp.jsxParseElementAt (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/plugins/jsx/index.js:440:30)
    at Parser.pp.jsxParseElement (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/plugins/jsx/index.js:481:15)
    at Parser.parseExprAtom (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/plugins/jsx/index.js:14:21)
    at Parser.pp.parseExprSubscripts (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:270:19)
    at Parser.pp.parseMaybeUnary (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:250:19)
    at Parser.pp.parseExprOps (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:180:19)
    at Parser.pp.parseMaybeConditional (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:157:19)
    at Parser.pp.parseMaybeAssign (/node_modules/rollup-plugin-babel/node_modules/babel-core/node_modules/babylon/lib/parser/expression.js:120:19)
[...]

PR #1025, opened 2016-10-06
deepClone causes 'RangeError: Maximum call stack size exceeded' error…

talmobi:
I think the issue may be with deepClone, rollup v.0.34.13 seems to be the latest version that should work, after that the deepClone commit was added that seems to cause RangeError: Maximum call stack exceeded errors.

I think this commit (between v0.34.13 and v0.35.0) is the reason for the errors: https://github.com/rollup/rollup/commit/83ccb9725374e0fde9d07043959c397b15d26c67

test case: https://github.com/talmobi/wrollup_issue_2

more discussion of the issue: https://github.com/rollup/rollup-watch/issues/7#issuecomment-251746728

and: https://github.com/rollup/rollup/issues/1010

suggested temporary fix (or use rollup v 0.34.13): https://github.com/talmobi/rollup/commit/3eceff612fe96ddb2b99be5d6e8e12cc9bc1b9e9

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -39,7 +39,8 @@ export default class Module {
 		timeStart( 'ast' );
 
 		this.ast = ast || tryParse( code, this.comments, bundle.acornOptions, id ); // TODO what happens to comments if AST is provided?
-		this.astClone = deepClone( this.ast );
+		// this.astClone = deepClone( this.ast );
+		this.astClone = this.ast
 
 		timeEnd( 'ast' );
 


Rich-Harris:
Thanks for the detailed reproduction and for narrowing it down to `deepClone`. For whatever reason, it's not failing for me at the moment, but I definitely have experienced something similar (unfortunately, it was at a time when I couldn't stop what I was doing to investigate). Will keep an eye out for it.

It's necessary to clone the AST because it gets mutated during bundling – reusing the same tree results in bugs.

rogervila:
any update on this?

shouze:
@Rich-Harris I confirm we experiment the same issue.

Rich-Harris:
closing in favour of https://github.com/rollup/rollup/pull/1205 – thanks

PR #1042, opened 2016-10-10
Support cycles in deepClone

ochafik:
Besides adding support for cycles (introduced by tsc through the [Node.parent](https://github.com/Microsoft/TypeScript/blob/08eafd3a28a33842511f96e83582317ce74ec6ef/lib/typescriptServices.d.ts#L423) property), this code helps reuse other node instances that are present in multiple locations (e.g. [Identifier](https://github.com/Microsoft/TypeScript/blob/08eafd3a28a33842511f96e83582317ce74ec6ef/lib/typescriptServices.d.ts#L433) nodes, empty `[]` array, etc).

Note: ES6 [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) uses same-value semantics for keys (ie. its an IdentityHashMap in Java terms)

### DIFF
--- a/src/utils/object.js
+++ b/src/utils/object.js
@@ -20,20 +20,24 @@ export function assign ( target, ...sources ) {
 
 const isArray = Array.isArray;
 
-// used for cloning ASTs. Not for use with cyclical structures!
-export function deepClone ( obj ) {
+// used for cloning ASTs.
+export function deepClone ( obj, identityMap = new Map() ) {
 	if ( !obj ) return obj;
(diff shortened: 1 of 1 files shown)


wearhere:
I'm not sure this is a total fix for #1010—with this patch in place, I no longer get a call stack error, but rollup just never finishes bundling.

Rich-Harris:
Closing in favour of https://github.com/rollup/rollup/pull/1205 – thanks

Issue #1052, opened 2016-10-13
Allow imports to resolve against certain file extensions


bretkikehara:
### issue

A user should be allowed to implicitly import files. In [utils/defaults.js](https://github.com/rollup/rollup/blob/152afb9732975188c7f4bf716c143da549afd432/src/utils/defaults.js#L9-L21), it seems that only files with `*.js` extension will correctly resolve. Thus, rollup doesn't work well with projects with JSX files even if the rollup babel plugin has been enabled to execute the JSX transpile to JS.

While renaming `foo.jsx` to `foo.js` ensures rollup plays well with JSX files and babel can do its job, this is in directly violation with eslint react default rule of [throwing errors for JSX files with the wrong extensions](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md).

Tries to solve:
https://github.com/rollup/rollup/issues/115
https://github.com/rollup/rollup-plugin-babel/issues/44,
### proposal

With PR #1053, module resolution should be user-configurable with fallback of `.js`. 

``` js
rollup.rollup({
  entry: 'main.js',
  extensions: [ '.js', '.jsx' ],
  plugins: [
    babel({ ... }),
  ],
})
```

McPo:
Whats the status on this, running into this issue myself?

Victorystick:
@bretkikehara If you specify the extension in the import, you can use whatever extension you please. @Rich-Harris has given several reasons ([here](https://github.com/rollup/rollup/issues/480#issuecomment-177520954) and further down in the thread) why extensionless imports are bad.

It's very unlikely that we'd add support for specifying extensions to Rollup core, especially since it's so easily solved with a plugin such as [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve). In fact, it's more likely that we'd remove the special-case of handling `.js` extensions altogether.

McPo:
@victorystick if that's the case then I'd be for removing the special case of handling *.js, for consistency.

Victorystick:
If what you want is to resolve certain file-extensions and want to avoid `node-resolve`, you can use a custom plugin.

```js
import extensions from './extensions.js';

export default {
  ...
  plugins: [
    extensions({
      extensions: [ '.jsx', '.foobar', '.etc' ],
    })
  ]
};
```

An implementation could look something like this. See [the Wiki](https://github.com/rollup/rollup/wiki/Plugins#creating-plugins) for more info.

```js
[...]

PR #1053, opened 2016-10-13
allow imports to resolve against certain extensions

bretkikehara:
See #1052 for more information.

For more robust support, what about passing `{ extensions: this.extensions }` instead of passing this.extensions directly? LMK...

### DIFF
--- a/bin/src/help.md
+++ b/bin/src/help.md
@@ -13,6 +13,7 @@ Basic options:
 -i, --input              Input (alternative to <entry file>)
 -o, --output <output>    Output (if absent, prints to stdout)
 -f, --format [es]       Type of output (amd, cjs, es, iife, umd)
+-ext, --extensions       Allowed file extensions
 -e, --external           Comma-separate list of module IDs to exclude
 -g, --globals            Comma-separate list of `module ID:Global` pairs
                             Any module IDs defined here are added to external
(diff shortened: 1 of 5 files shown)


thisconnect:
Isn't that already possible with the https://github.com/rollup/rollup-plugin-node-resolve plugin?

bretkikehara:
while `.jsx` may be resolvable with `node-resolve` plugin, it seems a bit odd to resolve your own source files when the original purpose of `node-resolve` is to resolve 3rd-party modules.

If `node-resolve` is the solution, IMO there is large enough use case to bring this basic functionality into the main `rollup` bundler rather than have every person add `node-resolve` just to use JSX. It adds a barrier to the React community for those who are looking to replace `webpack` or `browersify` as their build system. Larger projects with external imports will undoubtedly need `node-resolve`, but at least the source files will be found and it doesn't look as if rollup is broken. It is easier to troubleshoot 'i need to fix my imports using `node-resolve`' if the JSX source files are at least being processed by rollup. Without this PR, Rollup will ignore JSX so imports will never be processed.

From my own anecdote, this is the reason why I ended up ditching rollup for webpack in project. =/

Victorystick:
Closing, see https://github.com/rollup/rollup/issues/1052#issuecomment-260065475.

PR #1056, opened 2016-10-15
Remove sourceMappingURL only in comments

btd:
I have problem that current code removes `//# sourceMappingURL=....` even from string (because of regexp search). I noticed in code there are 2 places where remove happen - in Module and Bundle classes.

With Module idea is very clear - we iterate by comments only and clean sourcemaps. But with Bundle i did not get original idea why it is need, and it is also breaking my usecase (i am inserting SM with plugin.intro string matching this regexp - point there it is a string not comment). Also there is no test for this part of code. 

Could you please point me how should i better rework this part? or i can remove it at all? (we can reparse end bundle to remove comments, but i do not see why plugin could not do anything they want).

Also it looks it is more correct to remove comment body but rest //. For example we had `//# sourceMappingURL=...` and rest only `//`. Is it right?

Thanks.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -15,7 +15,6 @@ import { mapSequence } from './utils/promise.js';
 import transform from './utils/transform.js';
 import transformBundle from './utils/transformBundle.js';
 import collapseSourcemaps from './utils/collapseSourcemaps.js';
-import SOURCEMAPPING_URL from './utils/sourceMappingURL.js';
 import callIfFunction from './utils/callIfFunction.js';
 import { dirname, isRelative, isAbsolute, normalize, relative, resolve } from './utils/path.js';
 import BundleScope from './ast/scopes/BundleScope.js';
@@ -428,8 +427,7 @@ export default class Bundle {
 		let map = null;
 		const bundleSourcemapChain = [];
 
-		code = transformBundle( code, this.plugins, bundleSourcemapChain, options )
-			.replace( new RegExp( `^\\/\\/#\\s+${SOURCEMAPPING_URL}=.+\\n?`, 'gm' ), '' );
+		code = transformBundle( code, this.plugins, bundleSourcemapChain, options );
 
 		if ( options.sourceMap ) {
 			timeStart( 'sourceMap' );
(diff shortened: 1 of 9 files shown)


btd:
Reapplied to master. @Rich-Harris if you have time can you please check.

Rich-Harris:
Hi @btd – thanks for this PR. The issue was actually fixed in https://github.com/rollup/rollup/commit/ccc7ad7352a1c0fbc4b3c157a7a57be1e1cd3f7d – apologies, been a bit overwhelmed by the outstanding issues/PRs and didn't realise you'd already fixed it. Will close this.

btd:
@Rich-Harris thanks, but regexps is not a fix for this. You can try with ``template strings and put sourceMappingURL string at start``, like [this](http://rollupjs.org/?version=0.38.0&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIydmFyJTIwYSUyMCUzRCUyMCU2MCU1Q24lMkYlMkYlMjMlMjBzb3VyY2VNYXBwaW5nVVJMJTNEYS5iLmZmb28lNUNuJTYwJTNCJTVDbmNvbnNvbGUubG9nKGEpJTNCJTIyJTdEJTVEJTdE).

IvanSanchez:
I wonder if whitespace should be allowed between the start of the line and the sourcemappingURL comment?

btd:
@IvanSanchez in spec i did not found any restrictions. It just said:
```
The generated code may include a line at the end of the source, with the following form:

     //# sourceMappingURL=<url>

```

Nothing about whitespace. This could be interpreted as if this is not restricted, this should be allowed. I also seen in past such pattern used:
```
/*
//@ sourceMappingURL=...
*/
```
This was when initial spec authors did not test it in IE. But anyway you see pattern there, special string `//@ sourceMappingURL` at the start of row.

But problem actually that @Rich-Harris fix is not working for all cases. I already suggested failing case.

PR #1057, opened 2016-10-16
Quote names in module definition if legacy option is set

chge:
In addition to pull request #989.

This adds single quotes around property names inside module definition, if `legacy` option is set, to prevent reserved words to appear in object keys.

Example output:

``` javascript
var foo = (Object.freeze || Object)({
    'default': bar
});
```

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -82,6 +82,9 @@ export class SyntheticNamespaceDeclaration {
 			if ( original.isReassigned ) {
 				return `${indentString}get ${name} () { return ${original.getName( es )}; }`;
 			}
+			if (legacy) {
+				name = `'${name}'`;
+			}
 
 			return `${indentString}${name}: ${original.getName( es )}`;
 		});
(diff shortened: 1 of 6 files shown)


chge:
How to reproduce:
```javascript
// foo.js
import * as bar from './bar';
console.log(bar);
```
```javascript
// bar.js
export default 42;
```
Currently `rollup` outputs:
```javascript
var bar = 42;


var bar$1 = (Object.freeze || Object)({
        default: bar
});

console.log(bar$1);
```

To workaround this using plugin, you should parse generated code and try to reliably distinguish this `default:` from other `default:`'s in `switch` statements, which is challenging.

Additionally, it's possible to do the following:
```javascript
// bar.js
const bar = 42;
export { bar as for };
```
which becomes:
```javascript
const bar = 42;



var bar$1 = (Object.freeze || Object)({
        for: bar
});

console.log(bar$1);
```

Rich-Harris:
Thanks! I made a small tweak so that only *reserved* properties get quoted

PR #1068, opened 2016-10-20
[fix] remove Object.define for legacy; pass into cli; test legacy

futurist:
This PR does below things for `option.legacy=true`
1. Remove `Object.define` esModule line
2. Also pass the `legacy` option to CLI
3. Test for the new code, fix previously legacy test use case.


Rich-Harris:
Thanks 🍻

PR #1069, opened 2016-10-20
Do not output getter defs when in legacy mode

IvanSanchez:
This prevents rollup from outputting object getters for reassigned exports (which will trigger a syntax error when `Object.freeze()` is skipped).

Related to #1068, based on comments from https://github.com/futurist/rollup-plugin-es3/issues/3 , ~~will try to add~~ added a test for this.

### DIFF
--- a/src/Declaration.js
+++ b/src/Declaration.js
@@ -79,7 +79,7 @@ export class SyntheticNamespaceDeclaration {
 		const members = keys( this.originals ).map( name => {
 			const original = this.originals[ name ];
 
-			if ( original.isReassigned ) {
+			if ( original.isReassigned && !legacy ) {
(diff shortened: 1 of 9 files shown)


IvanSanchez:
FYI, without the fix, the new test fails as so:

```diff
  var browserStuff = (Object.freeze || Object)({
-         get browserSpecificThing () { return browserSpecificThing; },
+         browserSpecificThing: browserSpecificThing,
          foo: foo
  });
```

Rich-Harris:
Thank you!

derrickb:
Hi @Rich-Harris I'm running into this problem now, but there doesn't appear to be a `legacy` option anymore. Can we reopen this issue? I need getters/setters removed from my output. Thank you!

PR #1076, opened 2016-10-24
destructure options in Scope constructor

zackschuster:
i couldn't think of a reason you'd opt out of destructuring here; apologies if i'm mistaken somehow


Victorystick:
Looks good.

Issue #982, opened 2016-09-24
rollup fails to resolve entry file when using --watch, seemingly at random


hpfast:
I'm just trying rollup for the first time and cannot figure this error out involving --watch. I've set up a simple project and compile src/scripts/main.js to build/js/main.min.js. As the console output below shows, I save the source file multiple times (sometimes with a change, sometimes without) until eventually I get this error. The number of times it succeeds varies, and I can't tell if waiting longer helps.

I've gradually stripped out all plugins from my rollup.config.js. I was using uglify, replace, babel, eslint, commonjs, and resolve, but the error persists even with the following config:

```
   export default {
     entry: 'src/scripts/main.js',
     dest: 'build/js/main.min.js',
     format: 'iife',
     sourceMap: 'inline'
   };
```

I have rollup-watch installed locally because rollup complains it can't find it globally (I think because on Ubuntu I have to use sudo to install it globally)

<details>
  <summary>

Console output:</summary>

```
 hans@vast  ~/priv/prj/foo  npm run watch-js

> foo@1.0.0 watch-js /home/hans/priv/prj/foo
> rollup -c -w

checking rollup-watch version...
bundling...
bundled in 26ms. Watching for changes...
bundling...
bundled in 12ms. Watching for changes...
bundling...
bundled in 17ms. Watching for changes...
bundling...
bundled in 8ms. Watching for changes...
bundling...
bundled in 14ms. Watching for changes...
bundling...
Could not resolve entry (src/scripts/main.js)
Error: Could not resolve entry (src/scripts/main.js)
    at /home/hans/priv/prj/foo/node_modules/rollup/src/Bundle.js:97:29
    at process._tickCallback (internal/process/next_tick.js:103:7)
Type rollup --help for help, or visit https://github.com/rollup/rollup/wiki
```

</details>

hpfast:
my setup here: https://github.com/hpfast/rollup-error

hpfast:
I wonder if I'm hitting something like this issue in rollup-watch? https://github.com/rollup/rollup-watch/issues/4

I have the problem when using Vim, but not with Leafpad.

Victorystick:
If the problem is editor dependent, I've experienced something similar using Sublime Text. There was an atomic save mode which for every save created a new file and renamed it. It may be that you're experiencing something similar with Vim and Leafpad. Are you having this issue with the latest version of rollup-watch?

hpfast:
Hi,
[...]

PR #1087, opened 2016-10-28
Hack to fix external modules with --watch, as seen in issue #982

EmielM:
When running rollup `--watch`, the module cache triggers a bug when compiling in the second run.

In the first run, `fetchAllDependencies` calls `this.resolveId(..)` that yields `false` for externals. However, it subsequently saves the actual `externalId` in the `resolvedIds` cache.

The second run, this `externalId` is now returned instead of `false` and the `Treating xx as external dependency` path is not hit.

The hack just doesn't copy `resolvedIds` from the module cache letting rollup re-evaluate them on subsequent compiles. I have no idea what the proper fix should be. Should externals be saved on `resolvedIds`?

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -267,7 +267,7 @@ export default class Bundle {
 				return transform( source, id, this.plugins );
 			})
 			.then( source => {
-				const { code, originalCode, originalSourceMap, ast, sourceMapChain, resolvedIds } = source;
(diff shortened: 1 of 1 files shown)


EmielM:
Right, seems to break a test. It fixes my issue, though. Any thoughts?

EmielM:
This seems fixed in the latest version 🎉

PR #1126, opened 2016-11-25
resolve index.js if file is a path

icfr:
resolve index.js when the import file is a directory 
```javascript
// main.js
import store from './store'
```
```
+-- app/
+-- main.js
+-- store/
|   +-- index.js  
```
will load ./store/index.js

```
+-- app/
+-- main.js
+-- store.js
+-- store/
|   +-- index.js  
```
will load ./store.js


icfr:
test fail because i dont see that browser build load a different fs library ...

PR #1143, opened 2016-12-13
Improve tree-shaking of assignment expressions and unused functions

### LINKED ISSUE #1120: Unused code included in output - part 2
Reported by @bclinkinbeard in https://github.com/rollup/rollup/issues/977#issuecomment-262545870

Simplified test case using rollup version 0.36.4

$ cat a.js
```
function bar(umm) {
    umm = hmm();
    console.log("bar");
}
function hmm() {
    return true;
}
function baz() {
    console.log("baz");
}
baz();
```
$ node bin/rollup a.js 
```
function hmm() {
    return true;
}
function baz() {
    console.log("baz");
}
baz();
```
`function hmm` should have been dropped.

### LINKED ISSUE #1142: Recursive function uses itself and not gets removed by tree-shaking
When having a function that calls itself, it is not removed by tree-shaking. Seems this happens because the Node gets `activated`, at least that's were I ended debugging.

## Input
```js
var set = function set(object) {
   object.value = value;

   set(Object.getPrototypeOf(object));
};
```

## Expected output
```js
'use strict';

```

## Current output
```js
'use strict';

var set = function set(object) {
   object.value = value;

   set(Object.getPrototypeOf(object));
};
```

## Example

http://rollupjs.org/?version=0.37.0&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIydmFyJTIwc2V0JTIwJTNEJTIwZnVuY3Rpb24lMjBzZXQob2JqZWN0KSUyMCU3QiU1Q24lMjAlMjAlMjBvYmplY3QudmFsdWUlMjAlM0QlMjB2YWx1ZSUzQiU1Q24lNUNuJTIwJTIwJTIwc2V0KE9iamVjdC5nZXRQcm90b3R5cGVPZihvYmplY3QpKSUzQiU1Q24lN0QlM0IlMjIlN0QlNUQlN0Q=

## Environment

rollup: 0.37
node: 6.9.1
os: macOS 10.11

 I can help to fix it, but I'm not sure where to start. I debugged a testcase with this input-code, but I'm not sure where the whole activation happens, and how the Node could detect that it got activated by itself.

danez:
This fixes tree-shaking for recursive functions which also have side effects as described in #1142.

~~The basic idea is, that if an identifier is within its declaration, it is not marked as `activated`.  Do you think this is a proper solution?~~
I added three new test, and enabled a skipped one, that was testing something similar and works now.

Fixes #1142 
Fixes #1120


kzc:
I think the patch in https://github.com/rollup/rollup/issues/1120#issuecomment-262889664 is a more general solution that covers more cases.

danez:
I integrated all the changes from #1120.

kzc:
LGTM

May I suggest a catchier title to get it merged quicker?

Improve tree-shaking of assignment expressions and unused functions

/cc @Rich-Harris

Rich-Harris:
Excellent, thanks!

PR #1144, opened 2016-12-13
Access global.* using array notation

### LINKED ISSUE #582: Using a scoped package name break UMD module generation
I want to generate a UMD module where the name is a scoped package, e.g. `@something/modulename`. Doing so triggers an error a build time because the generated code is invalid:

```
factory((global.@something/modulename = {}) ... etc.
```

Would it be possible to detect this kind of situation an instead generate something like:

```
factory((global['@something/modulename'] = {}) ... etc.
```

This is a killer for me because we scope all of our internal modules.

### LINKED ISSUE #584: Module name with dash in UMD format produces invalid output
It's probably not a big issue but the following config produces bad output:

``` js
export default config = {
  moduleName: "my-module-with-dashes",
  format:     "umd",
  // ...
}
```

Results in this line in the bundle's UMD head:

``` JS
// ...
  (factory((global.my-module-with-dashes = global.my-module-with-dashes || {})));
// ...
```

You should probably check the module name and use bracket syntax if necessary.

christopherthielen:
Allow arbitrary `moduleName` (including names with dashes like `ui-router` or scoped npm packages like `@angular/core`) to be added to the global object.

Closes #582
Closes #584

### DIFF
--- a/test/form/module-name-scoped-package/_expected/umd.js
+++ b/test/form/module-name-scoped-package/_expected/umd.js
@@ -1,0 +1,13 @@
+(function (global, factory) {
+	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
+	typeof define === 'function' && define.amd ? define(['exports'], factory) :
+	(factory((global['@scoped/npm-package'] = global['@scoped/npm-package'] || {})));
+}(this, (function (exports) { 'use strict';
+
+	let foo = 'foo';
+
+	exports.foo = foo;
+
+	Object.defineProperty(exports, '__esModule', { value: true });
+
+})));
--- a/test/form/module-name-wat/_config.js
+++ b/test/form/module-name-wat/_config.js
@@ -1,0 +1,6 @@
+module.exports = {
+	description: 'properly dereferences properties on the global object regardless of nesting',
+	options: {
+		moduleName: 'foo.@scoped/npm-package.bar.why-would-you-do-this'
+	}
+};
--- a/src/finalisers/iife.js
+++ b/src/finalisers/iife.js
@@ -3,16 +3,23 @@ import { getName } from '../utils/map-helpers.js';
 import getInteropBlock from './shared/getInteropBlock.js';
 import getExportBlock from './shared/getExportBlock.js';
 import getGlobalNameMaker from './shared/getGlobalNameMaker.js';
+import propertyStringFor from './shared/propertyStringFor';
+
+// thisProp('foo.bar.baz') === "this['foo']['bar']['baz']"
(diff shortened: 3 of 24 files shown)


test/form/namespaced-default-exports/_expected/iife.js
```diff
@@ -1,6 +1,6 @@
-this.foo = this.foo || {};
-this.foo.bar = this.foo.bar || {};
-this.foo.bar.baz = (function () {
+this['foo'] = this['foo'] || {};
+this['foo']['bar'] = this['foo']['bar'] || {};
+this['foo']['bar']['baz'] = (function () {
```
christopherthielen: Still processes namespaced exports by splitting on the dots

test/form/module-name-scoped-package/_expected/umd.js
```diff
@@ -0,0 +1,13 @@
+(function (global, factory) {
+	typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
+	typeof define === 'function' && define.amd ? define(['exports'], factory) :
+	(factory((global['@scoped/npm-package'] = global['@scoped/npm-package'] || {})));
```
christopherthielen: allow arbitrary strings, including npm scoped packages.  Not sure this is/should be a requirement, but I added a test anyway because of #582

Rich-Harris:
Thanks @christopherthielen! Any chance you could modify this PR such that valid property names are left as-is? There are advantages to using dot notation, so I think it would be better if properties were only quoted as necessary since they're in the minority.

christopherthielen:
sure, no problem

test/form/module-name-wat/_config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+	description: 'allows module name with dashes to be added to the global object',
+	options: {
+		moduleName: 'foo.@scoped/npm-package.bar.why-would-you-do-this'
```
christopherthielen: module-name-wat

christopherthielen:
Okey dokey, done.

I chose to be naive about when to quote/not quote properties.  The actual rules for javascript property access are more complex: https://mathiasbynens.be/notes/javascript-identifiers .  There are npm  modules you could use to test if a property can be accessed using a dot or not, but it seemed overkill.  This should catch 99% and quote when unsure.  

Also I didn't want to force maintenance of a variable checker on the project.  I did find some (low-stars) [npm](https://www.npmjs.com/package/var-validator)  [packages](https://www.npmjs.com/package/is-var-name) that could be used if you prefer something along those lines.

Rich-Harris:
Excellent, thank you!

Issue #1104, opened 2016-11-10
Pruning of sourcemap comments is too agressive


IvanSanchez:
Be advised this is an **edge case**.

I discovered this while working with my [rollup-plugin-file-as-blob workflows](https://gitlab.com/IvanSanchez/rollup-plugin-file-as-blob). In order to reproduce the bug:

* Have a `*.js` file with a `//# sourceMappingURL=` comment at the end
* Include that file as a `blob://` URL via rollup-plugin-file-as-blob
* Try to parse the resulting code

My plugin returns module code as follows:

```js
export default __$strToBlobUri("Contents of a plain-text file goes here", "text/plain; charset=us-ascii", false);
```

Which, for a javascript file, might look like:

```js
export default __$strToBlobUri("\"use strict\";foobar();\n//# sourceMappingURL=something.js.map\n", "text/plain; charset=us-ascii", false);
```

The code from https://github.com/rollup/rollup/commit/9164f80487ab17165371ddc7b6224250530f82f8 will take that module code, run the regexp, **not notice that the sourcemap comment is inside a string literal**, then strip the closing `"` and the rest of the line, i.e.:

```js
export default __$strToBlobUri("\"use strict\";foobar();\n
```

That will obviously break all kinds of havoc in the parsers.

I think I can work around the bug in my plugin code, given that this is an edge case involving custom rollup plugins. Nonetheless I'd appreciate if rollup handling of sourcemap comments (https://github.com/rollup/rollup/commit/9164f80487ab17165371ddc7b6224250530f82f8) could be a bit less naïve.

IvanSanchez:
I worked around this in https://gitlab.com/IvanSanchez/rollup-plugin-file-as-blob/commit/24d64f5e3e40c3d73e8c6e72100f1c85da5356f5?view=inline. Unless somebody else hits the problem, I guess this can be considered ultra-low priority.

btd:
It looks like https://github.com/rollup/rollup/pull/1056. I got the same problem when worked on css modules

Rich-Harris:
This is fixed now – the regex for matching sourceMappingURL comments has a `^` at the start

btd:
just for reference, fix proposed does not work for such case:
```js
var str = `
//# sourceMappingURL=...
`
console.log(str);
```
I still get failled test in my plugin.

PR #1171, opened 2016-12-22
Failling test for #1104

btd:
@Rich-Harris i am really sorry for being that noisy.

### DIFF
--- a/test/form/removes-existing-sourcemap-comments/_expected/amd.js
+++ b/test/form/removes-existing-sourcemap-comments/_expected/amd.js
@@ -4,6 +4,10 @@ define(function () { 'use strict';
 		return 42;
 	};
 
-	console.log( foo() );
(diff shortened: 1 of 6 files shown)


Rich-Harris:
Ah, of course, I hadn't considered that case. Is this something you're experiencing in the wild (as opposed to a theoretical issue – still needs fixing, but lower priority)?

Rich-Harris:
Ignore that, just saw your comment on #1104 – it *is* happening in the wild

Rich-Harris:
Added your fix back. Sorry for the confusion, I was careless and didn't read your patch properly. Much better than the crude regex approach

btd:
@Rich-Harris Thank you. Yes i got this issue while working on postcss plugin. I have to write generated sourcemap for css to include it to bundle (css written as string literal).

PR #1174, opened 2016-12-22
remove, and warn about, unused imports from external modules

### LINKED ISSUE #595: Unused imports from external modules should be stripped from import statements in ES bundles
The following code...

``` js
// main.js
import { unused } from 'external';

function alsoUnused () {
  unused();
}
```

...could result in the following ES bundle (i.e. it imports the external module in case it has side-effects, but doesn't need `unused` specifically)...

``` js
import 'external';
```

...but instead we get this...

``` js
import { unused } from 'external';
```

...which causes [confusion](http://stackoverflow.com/questions/36417779/rollup-including-unneeded-dependancies-and-missing-a-function-in-output).

Rich-Harris:
fixes #595

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -175,6 +175,21 @@ export default class Bundle {
 
 				timeStart( 'phase 4' );
 
+				// while we're here, check for unused external imports
+				this.externalModules.forEach( module => {
+					const unused = Object.keys( module.declarations )
+						.filter( name => name !== '*' )
+						.filter( name => !module.declarations[ name ].activated );
+
+					if ( unused.length === 0 ) return;
+
+					const names = unused.length === 1 ?
+						`'${unused[0]}' is` :
+						`${unused.slice( 0, -1 ).map( name => `'${name}'` ).join( ', ' )} and '${unused.pop()}' are`;
+
+					this.onwarn( `${names} imported from external module '${module.id}' but never used` );
+				});
+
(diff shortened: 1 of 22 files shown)


kzc:
Does this PR assume that unreferenced dropped imports do not have intentional side effects?

Rich-Harris:
It'll still import the modules, it just skips the bindings, so this...

```js
import { unused } from 'external';

function alsoUnused () {
  unused();
}
```

...becomes this...

```js
import 'external';
```

...instead of this:

```js
import { unused } from 'external';
```

kzc:
Ah okay. I misread the patch.

Perhaps Rollup could support an option to drop unreferenced imports like this:
```js
import { unreferenced } from 'external';
```
and keep imports like this:
```js
// Import an entire module for side effects only, without importing any bindings.

import "my-module";
```
as per https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/import

But then again, the warning might be enough to get the user to drop the import themself.

Rich-Harris:
Not a bad idea. Will leave it in the nice-to-have pile for now until someone asks for it though, in the interests of getting through the backlog

kzc:
Upon more thought, nothing in the ES2015 spec precludes `import { unreferenced } from 'external';` from having a side effect so it can't safely be dropped. So the idea wouldn't work.

Issue #1122, opened 2016-11-23
transformBundle/ongenerate aren't async


tivac:
I realize that `transformBundle`/`ongenerate` are synchronous because `bundle.generate()` is synchronous, but I'm starting to run into cases where that's pretty limiting. Given that `rollup.rollup()` and `bundle.write()` already return a promise would it be weird for `bundle.generate()` and the associated hooks to be changed?

I'd be happy to do that work and send a PR for it for what it's worth!

I realize this is a pretty big breaking change but I'd be happy to further explain why that hook/fn being sync is a problem for me if it'd help. I did some explaining already in [gitter](https://gitter.im/rollup/rollup?at=5835e3e416ee9e640479df46)

Rich-Harris:
As of 0.45, `bundle.generate` returns a promise, so `transformBundle` hooks can too — will close this issue

PR #1177, opened 2016-12-23
WIP: Promisify core

btd:
Hi all.

I am not expecting this to be merged. I would like to gather initial feedback about idea to make possible for `.intro`, `.outro`, `.footer`, `.banner`, `bundle.generate` to return Promises (first 4 methods could return sync values as it now also). If such idea will get more or less positive feedback to be included in rollup, i would like to do the same for other methods (`.transformBundle`, `.on*` methods).

Currently all tests passed.

I think it is clear why this could be helpful. But in any case: it is not rare when you need to do async operation to be included in header (e.g. get commit hash etc) and not all of them could be replaced with Sync equvalents.

I am appriciating any thoughts. Thanks.


tivac:
I filed #1122 because I think it's really weird that they're sync currently. I'm :+1: x :100: on this.

Rich-Harris:
Thanks for the PR @btd. I think making `intro`, `outro`, `footer` and `banner` async makes sense – there are valid use cases, and it could be done in a way that's completely non-breaking.

I'm less sure about making `bundle.generate` async. That *is* a significant breaking change, and as far as I can see the only reason to do it is to make it possible for `transformBundle` to return a promise. In terms of cost-benefit, I'm not convinced it's worth it. But then maybe I just don't understand the reasons why it's necessary?

I'm even less sure about making `onwrite` and `ongenerate` async. They're basically just event handlers, and it seems weird for event handlers to be async. But again, maybe I'm just missing the use case?

tivac:
Making the event handlers async would be a big help for me so I don't have to do the weird "attach a promise to `result`" thing I'm doing atm.

When I was digging into bundle splitting having `transformBundle` be async would've been useful, but I forgot why at this point. 😓

tivac:
@Rich-Harris that'd be ok too.

btd:
I totally agree about `.intro`, `.outro`, `.banner` and `.footer`. About `bundle.generate` it is transitive dependency: all four previous methods (.intro etc) used in `bundle.render` (so it is becoming returning promise), but `bundle.render` used in `bundle.generate` that is why it is returning promise. 

About `.transformBundle` returning promise, i had i problem in past (but i already forgot usecase). It was something with emitting several files. But currently i did this because `bundle.generate` become async.

Rich-Harris:
My thinking was that `intro`, `outro`, `banner` and `footer` could be consolidated in `bundle.build`, so that they'd be ready to go in `bundle.generate`. That's arguably more correct anyway since their output doesn't vary with the `generate` configuration.

btd:
@Rich-Harris sounds good. But it will be still breaking change, because some people could rely on order of plugin methods calls. I will prepare new PR.

How can we start pushing emitting multiple files output?

btd:
Closing this to prepare another one

Issue #1108, opened 2016-11-12
Unused function declarations within functions not removed


kzc:
The unused `function foo` is removed at top level scope, but not at function scope.
```
$ node bin/rollup -v
rollup version 0.36.3
```
```
$ cat wrap.js

function foo() { console.log("FOO"); /* is removed */ }
function baz() { console.log("BAZ"); }
function boo() { /* is removed */ }
baz();

function quux() {
    function foo() { console.log("FOO"); /* should be removed */ }
    function baz() { console.log("BAZ"); }
    function boo() { /* should be removed */ }
    baz();
}
quux();
```
```
$ node bin/rollup wrap.js

function baz() { console.log("BAZ"); }
baz();

function quux() {
    function foo() { console.log("FOO"); /* should be removed */ }
    function baz() { console.log("BAZ"); }
    function boo() { /* should be removed */ }
    baz();
}
quux();
```
Note: this is not a dupe of #977.

Issue #1178, opened 2016-12-24
Unused class declarations within functions not removed


kzc:
rollup version 0.38.0

$ cat inner_class.js 
```
function foo() {
    class C {}
    class UnusedInnerClass {}
    return C;
}
class UnusedClass {}
console.log(foo().name);
```
$ node bin/rollup inner_class.js 
```
function foo() {
    class C {}
    class UnusedInnerClass {}
    return C;
}
console.log(foo().name);
```
Related issue: #1108 Unused function declarations within functions not removed

PR #1179, opened 2016-12-24
drop unused function and class declarations within functions

kzc:
Fixes:  #1108, #1178


Rich-Harris:
outstanding, thank you

PR #1182, opened 2016-12-25
Simplify repository value

vinkla:
Since we're using GitHub we can use [the shorthand syntax](https://docs.npmjs.com/files/package.json#repository).


Rich-Harris:
didn't know about the shorthand, thanks!

Issue #1194, opened 2016-12-29
Better warnings


Rich-Harris:
Warnings are a bit hard to do things with at the moment, because they're just strings, meaning that if you want to (say) suppress all warnings about `this` being undefined, you have to do some regex hackery.

It would be better if the warning was an object similar to an error object, with a warning code, a message, and other applicable information like file, line and column (which would allow the CLI to print a code frame – see also #296).

To minimise breakage, the warning object should have a `toString` method that returns the message.

Rich-Harris:
Further thought: we could probably group certain warnings together. For example, all the implicit external dependency or '`this` is undefined' warnings could be delivered in one go, with all the relevant details. Having a unified way to handle grouped warnings would be better than doing all the necessary book-keeping in an ad-hoc fashion.

yairEO:
I want to see warning, but recently you've added feature which prints to the console simple linting errors. I **do not** want to see those, they are huge and I already have elegant eslint messages. if i'll set `onwarn` to an empty function, I might miss "real" important errors. what should I do?

Rich-Harris:
The new system is a lot more flexible in terms of how warnings are printed. By default it'll show a code frame etc, but because the warning is now an object rather than a string you can do this sort of thing:

```js
onwarn ( warning ) {
  // simple handler
  console.warn( warning.message );
}

onwarn ({ loc, message }) {
  // print location if applicable
  if ( loc ) {
    console.warn( `${loc.file} (${loc.line}:${loc.column}) ${message}` );
  } else {
    console.warn( message );
  }
}

onwarn ( warning ) {
  // skip certain warnings
  if ( warning.code === 'UNUSED_EXTERNAL_IMPORT' ) return;

  // throw on others
  if ( warning.code === 'NON_EXISTENT_EXPORT' ) throw new Error( warning.message );

  // console.warn everything else
  console.warn( warning.message );
}
```

yairEO:
Thank you very much for the fine explanation! would be nice to copy it to the WIKI

Rich-Harris:
Good point! Done https://github.com/rollup/rollup/wiki/JavaScript-API#onwarn

Finesse:
What about adding a helper function which renders a warning beautifully?

Like this:
```
[20:10:00] Plumber found unhandled error:
 SyntaxError in plugin "gulp-better-rollup"
Message:
    ~/lib/jquery.fileDownload.js: Expecting Unicode escape sequence \uXXXX (3:19)

[...]

PR #1199, opened 2016-12-29
[WIP] [BREAKING] Better warnings

Rich-Harris:
See #1194. This PR makes it possible to emit warnings with structured information, including things like the location of the code that triggered the warning together with a code frame (if applicable).

Warnings must have a `code` and a `message` property. For warnings that relate to a specific part of the code, they may also have a `pos` (character index) property, a `loc` property which is a `{ file, line, column }` object, and a `frame` property highlighting the error location. Beyond that, any additional properties can be added (though we're not currently using that anywhere).

A `toString` property that logs the `message` property is added to the object, hopefully minimising the disruption this change causes.


Rich-Harris:
CLI output looks rather different now – it might be a little too bulky for some people's tastes, but rather than spending a long time finessing it I think it's worth releasing this and gathering feedback. There's plenty more we could do – grouping warnings by file or type, allowing different levels of verbosity, etc – but that work can happen later.

PR #1205, opened 2016-12-30
prevent mutation of cached ASTs

### LINKED ISSUE #1153: How keep watch when Error?
My npm dev command:
```
rollup -w -c build/rollup.dev.config.js
```

rollup.dev.config.js
```
import buble from 'rollup-plugin-buble'

export default {
	entry: 'src/main.js',
	format: 'umd',
	moduleName: 'Graph',
	plugins: [
		buble()
	],
	dest: 'dist/ugraph.js'
}
```

If compile success each should be okay, but when compile has error so can't no keep watch file after change file, 

Current solution manual reload npm command when rullopjs compile error

Who can help please!

Rich-Harris:
This fixes the dreaded 'Maximum call stack size exceeded' error that occurs with rollup-watch when a file with an error is subsequently fixed. It does so by preventing mutations of the cached AST, which is reused between incremental builds for files that *don't* have errors.

Fixes #1153, #1010, and https://github.com/rollup/rollup-watch/issues/21.

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -2,7 +2,7 @@ import { parse } from 'acorn/src/index.js';
 import MagicString from 'magic-string';
 import { locate } from 'locate-character';
 import { timeStart, timeEnd } from './utils/flushTime.js';
-import { assign, blank, deepClone, keys } from './utils/object.js';
+import { assign, blank, keys } from './utils/object.js';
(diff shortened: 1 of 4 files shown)


tivac:
🎉

talmobi:
:taco: 👍

wearhere:
woah awesome!

PR #1210, opened 2016-12-31
chmod a+x bin/rollup

### LINKED ISSUE #1209: generated bin/rollup not chmod +x
```
$ bin/rollup
bash: bin/rollup: Permission denied
```
```
$ node bin/rollup 

rollup version 0.39.2
=====================================

Usage: rollup [options] <entry file>
```

Rich-Harris:
fixes #1209

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "posttest-coverage": "remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped.json -b dist && remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped.lcov -t lcovonly -b dist && remap-istanbul -i coverage/coverage-final.json -o coverage/coverage-remapped -t html -b dist",
     "ci": "npm run test-coverage && codecov < coverage/coverage-remapped.lcov",
     "build": "git rev-parse HEAD > .commithash && rollup -c",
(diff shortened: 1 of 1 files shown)


kzc:
This works. Thanks.

kzc:
Would `chmod a+x` work for Windows users?

I might have suggested the same patch but don't know the npm portable way to do this.

Rich-Harris:
It passes CI so at the very least it doesn't break anything...

tivac:
Something like [shelljs/shx](https://github.com/shelljs/shx) would presumably be a safer cross-platform way to do that. Dunno how you'd feel about that extra dependency though.

Rich-Harris:
It only needs to run on the machine from which the package is being published though, no? As long as the file has the correct permissions when it's published, it'll have the correct permissions for everyone who installs the package regardless of OS?

tivac:
npm takes care of setting the `.bin` dir entries to be `+x` already, I wouldn't expect many people to execute `node_modules/rollup/bin/rollup` directly.

The use-case I was thinking of was lost souls like myself that develop on windows 👻

kzc:
There's always `fs.chmodSync(path, mode)`

https://nodejs.org/api/fs.html#fs_fs_chmodsync_path_mode

PR #1227, opened 2017-01-06
`throw` always considered to be a side effect

kzc:
Fixes one of the bugs listed in #305


Rich-Harris:
Ah, yep. This distinction made a bit more sense (though maybe not that much) back when Rollup could only remove top-level statements. Now that it's a lot more granular there's no need for it. Thanks

Issue #1140, opened 2016-12-12
`this.warn(...)` (and others) for plugins


Rich-Harris:
Via https://github.com/rollup/rollup-plugin-babel/pull/85 – would be good to have a standardised way for plugins to provide logging info, warnings, and errors. Mutating the `options` object is a bit mad, but currently the only way.

tivac:
> is a bit mad

https://github.com/tivac/modular-css/blob/master/src/rollup.js#L27-L29

...

😅

Rich-Harris:
Released this in 0.41

PR #1231, opened 2017-01-06
[WIP] [BREAKING] add this.warn method to plugin contexts

Rich-Harris:
See #1140. This adds `this.warn` to plugin `transform` hooks. Usage:

```js
// simple case — a string
this.warn( 'a message' );

// equivalent to
this.warn({ message: 'a message' });

// custom code/other data
this.warn({
  code: 'MY_CUSTOM_CODE', // defaults to 'PLUGIN_WARNING'
  message: 'a message',
  foo: 'bar'
});

// include a character position that the warning pertains to. this
// will cause `pos`, `loc` and `frame` properties to be added
// to the warning object, as with internal `module.warn(...)`
this.warn( 'a message', 42 );
```

In addition to `code`, `pos`, `loc` and `frame` properties, a `plugin` property will be added which is the plugin's name.

TODO:

* [x] `this.error(...)` helper
* [x] Update `toString` method to include plugin name
* [x] Update CLI to print plugin name


Rich-Harris:
Technically a breaking change (it no longer manipulates error messages) so should be a 0.41 release, though unlikely to actually cause breakage in the wild.

tivac:
@Rich-Harris do you have time to document this on the wiki? I'd love to add support but I haven't had a chance to fully grok this yet. Even some light docs would be really useful!

Rich-Harris:
@tivac try this on for size https://github.com/rollup/rollup/wiki/Plugins#warnings-and-errors

tivac:
Thanks!

PR #1232, opened 2017-01-06
Use stdout instead of stderr for rollup-watch non errors

porsager:
This makes it more easy for users and libraries using rollup-watch to react accordingly to rollup-watch output.


Rich-Harris:
Thanks — I was under the impression that stdout was only meant to be used for output, and any logging or diagnostics (whether info, warnings or errors etc) belonged to stderr. Can you elaborate on why this is needed?

porsager:
I'm pretty sure the intention is to keep info, diagnostics etc. in stdout and errors in stderr. There's a more thorough article on it here:
http://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

For instance in my specific use case I would like to listen to stdout and stderr from rollup-watch to show appropriate notifications to the user (so they don't have to watch the terminal). If everything comes as stderr I won't know when to show a big error notification or a small "done bundling" notification.

Rich-Harris:
In the article,

> McIllroy specifically says that `stderr` is for diagnostic messages

— so I guess it's a question of what counts as a diagnostic message. I'd always thought of `stdout` as being for the actual *output* of a program (which doesn't really apply to anything in the case of a task the generates the same file repeatedly like rollup-watch), which wouldn't include lines like 'bundling...'.

I think what we really need here is more configurability. Maybe if there was an `oninfo` option?

porsager:
Yeah, you're right, if the program produces an output just like rollup without a destination then it should go to stdout. 

What I've seen done mostly for programs which main function isn't to produce an output (services) is to log to stdout and error to stderr, same reason console.log goes to stdout and console.error goes to stderr in node. I guess it's because I see rollup-watch as a service that runs rollup 😉 

I can't imagine a solution to output a bundle on stdout from rollup-watch would be that useful either. 

So an oninfo to define like `rollup -c --watch --oninfo stdout`? Or was it something else you had in mind?

Rich-Harris:
It'd be an option in the config file I think, same as `onwarn`:

```js
export default {
  entry: 'src/index.js',
  dest: 'bundle.js',
  format: 'cjs',
  oninfo: info => console.log( info.message )
};
```

porsager:
Hi.. Sorry for the late reply.

As far as i can tell onwarn isn't used for rollup-watch output currently.
If you'd like I can change the pull request to include onwarn and oninfo to work with rollup-watch, but maybe just making the silent flag work with rollup-watch is a better idea :)

olsonpm:
I would be super confused if warn deviated from [node's warn](https://nodejs.org/api/console.html#console_console_warn_data_args)

> The console.warn() function is an alias for console.error().

spectras:
As a sidenote, rollup's current behavior of defaulting to write its output to the standard output matches what is done on almost all text processing tools in the unix world. I believe it should remain this way, because:
1) it's expected by most developers
2) it works well with the ecosystem. It allows standard chains like `rollup ..... | minify ..... | gzip -9 > output.js.gz`

porsager:
@spectras I certainly agree for that case, but this would only be for using rollup watch, where your example wouldn't work anyway 😉

ralfstx:
When running rollup from a build tool like grunt, stderr output will be marked as error:

screenshot

That doesn't look appropriate for a purely informational message. I'd agree that when the output goes to a file, the info messages should go to stdout. I'd expect this behavior as a default rather than having to configure it.

evocateur:
Tools that interpret `stderr` logging as "failure" without actually checking the process for a non-zero exit code are Doing It Wrong™. Logs aren't program output.

porsager:
@evocateur you know that rollup watch runs continously, outputting on stderr at each watch trigger?
The exit code is not really relevant in this discussion.

evocateur:
Indeed, I do know that. Doesn’t change the accuracy of my statement.

If tools want to respond to errors, the watcher process should either exit on error (a `--bail` option), or the tool should run the watcher programmatically and listen for error events. Both solutions require changes, of course, but they would be more correct than switching stdio pipes.

ralfstx:
@evocateur In my example, the tool (grunt) does not interpret the stderr messages as failure, it just prefixes stderr messages from the sub process with red `>>`, which makes it look like an error.

I can't follow the idea of stderr being just another output stream without any semantics. It is called "standard **error**" after all. Using stderr for purely informational messages feels like a misuse. At least, no other tool in my build chain would do that.

spectras:
@ralfstx Using stderr for anything that's not part of the output is a standard unix convention. Most unix tools works that way, and have been working that way for decades.

This is also supported by the ISO C standard, section 7.9.13 which explicitly states that:

> At program startup, three text streams are predefined and need not be opened explicitly
> — standard input (for reading conventional input),
> standard output (for writing conventional output), and
> standard error (for writing **diagnostic output**).

Emphasis mine. Note that  it does not say errors, it says diagnostic, that also includes warnings, progress notifications and debugging information. The fact that some tools in the javascript ecosystem fail to meet the standard should not be a justification to break those that perfectly follow it.

By the way, the canonical approach to this is to have an option (either on command line or in config file) to divert logging to some file or turn it off entirely.

ralfstx:
> Most unix tools work that way, and have been working that way for decades.

@spectras  Most unix tools do not print any diagnostic messages when everything works as expected. So maybe the solution here is a `--verbose` option?

spectras:
@ralfstx That would do indeed. Either way, --quiet or --verbose, it's just a matter of picking a sensible default.

porsager:
It seems like there are 3 scenarios to consider, in which I originally only was interested in fixing `rollup --watch`

## 1 `rollup --watch`
Using watch starts a continuous "service" which I think should do like any other continuous running service. Output info messages (build time, watch fired etc - like server logs) to stdout, and syntax errors etc during build to stderr (like unexpected errors in a server (if the service is set to restart or continue running is besides the point and a domain specific choice)). This is the only way for other tools to interface properly with rollup watch without having to resolve to string matching which is subject to change at any time. It'd also be easier to make the output for human eyes readable. (stdout green, stderr red).

## 2 `rollup` write code result to file defined in config / parameter
I don't care that much for this one, but I think this is the one @ralfstx mentions, and my thoughts on it would be close to how rollup --watch is handled. This is not like some unexpected thing happening for the user of rollup, they'll very likely be aware that they configured rollup to write the result to a file.

## 3 `rollup` output code result to stdout (for piping)
Of course the code should go to stdout in this case and errors on stderr.

I think it'd help if we either create a separate issue for # 2, or at least not mix # 1 and # 2 together.

shellscape:
Hey all. Thanks for participating in this PR. Rollup has made quite a few changes since this was opened. We're going to close this one for now. If the changes that have been made since aren't sufficient to address the concerns here, please open a new issue and we'll start this discussion fresh.

PR #1240, opened 2017-01-07
optimize namespace["foo"]

Rich-Harris:
This is #841 revisited (since that PR was against an ancient version of the codebase and would be hopeless to try and merge). It means that `namespace['foo']` and `namespace.foo` are treated the same way for the purposes of tree-shaking.

### DIFF
--- a/src/ast/nodes/MemberExpression.js
+++ b/src/ast/nodes/MemberExpression.js
@@ -1,14 +1,24 @@
-import isReference from 'is-reference';
 import relativeId from '../../utils/relativeId.js';
 import Node from '../Node.js';
 import { UNKNOWN } from '../values.js';
+
+const validProp = /^[a-zA-Z_$][a-zA-Z_$0-9]*$/;
 
 class Keypath {
 	constructor ( node ) {
 		this.parts = [];
 
 		while ( node.type === 'MemberExpression' ) {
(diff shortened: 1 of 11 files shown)


eiszfuchs:
Hm. 🤔  Since this change, my compilation breaks with
```TypeError: replacement content must be a string```

Did I do something wrong? I find myself unable to hunt down the cause.

Rich-Harris:
Oh! I just saw a similar error, it makes sense that it would be related to that change. If you do manage to create a repro (ideally on [rollupjs.org](http://rollupjs.org/)!) that would be a huge help — either way, will try and get a patch release out shortly

eiszfuchs:
[My code has grown preeeeeetty large](https://code.eiszblu.me/zeug/art-generator/src/13d667ed4b4cc0c0be90500d970956fb26c45425/gulpfile.js#L245) over the last months, I *might* try to isolate and reproduce something like that, but not before I had some sleep. (And I'm using some plugins that *could* be involved, too.)

ericf:
I hit this as well with React Intl's build: https://travis-ci.org/yahoo/react-intl/jobs/191738459#L320

Rich-Harris:
We have a repro! https://github.com/rollup/rollup/issues/1258

Issue #1254, opened 2017-01-12
Babili's sourcemaps causes error in collapseSourcemaps.js


Comandeer:
I've created [simple plugin for babili](https://github.com/Comandeer/rollup-plugin-real-babili/), however [sourcemaps don't work](https://github.com/Comandeer/rollup-plugin-real-babili/issues/1#issuecomment-269867964).

<details>
<summary>Stack trace</summary>

```
TypeError: Cannot read property 'traceSegment' of undefined
      at node_modules/rollup/dist/rollup.js:8902:23
      at Array.forEach (native)
      at node_modules/rollup/dist/rollup.js:8900:8
      at Array.map (native)
      at Link.traceMappings (node_modules/rollup/dist/rollup.js:8897:31)
      at collapseSourcemaps (node_modules/rollup/dist/rollup.js:9026:19)
      at Bundle$$1.render (node_modules/rollup/dist/rollup.js:9611:10)
      at Object.generate (node_modules/rollup/dist/rollup.js:9800:26)
      at rollup.then (tests/index.js:97:26)
```

</details>

It can be simply reproduced by running tests from my plugin, especially [ones dedicated to sourcemaps](https://github.com/Comandeer/rollup-plugin-real-babili/blob/master/tests/index.js#L92-L133).

After some investigation I discovered that Babili often returns last line mapping with only one segment:

<details>
<summary>Example Babili's output</summary>

```javascript
[ [ [ 0, 0, 0, 0, 0 ],
    [ 8, 0, 0, 8, 1 ],
    [ 11, 0, 0, 0 ],
    [ 12, 0, 0, 13 ],
    [ 16, 0, 0, 0 ],
    [ 17 ] ] ]
```

</details>

This causes that [`source` variable inside `Link.traceMappings` is undefined](https://github.com/rollup/rollup/blob/2a01a993efe04b5158aea834577f56140acdfd1f/src/utils/collapseSourcemaps.js#L33), so all efforts to use that variable (especially the very next line of code) just fails.

The simplest fix would be probably to just check if `source` variable is really set before trying to use it.

I also know that this issue seems to be very similar to #961 but my very quick research with GCC shows that the fix does not apply there and GCC still produces invalid sourcemaps.

Rich-Harris:
Released 0.41.5 with the fix — thanks, and sorry for the delay in merging

PR #1255, opened 2017-01-12
Babili's sourcemaps causes error in collapseSourcemaps.js

Comandeer:
This PR fixes issue described in #1254.

Due to Babili returning last line mapping with only one segment inside, `source` variable in `Link.traceMappings` remains undefined and all operations on this variable fails.

The proposed fix just checks if `source` is really set before doing anything with it.


kzc:
The one line fix to src/utils/collapseSourcemaps.js is fine, but rollup shouldn't have a devDependency on babel, so the test is not required - unless it can be rewritten without babel.

Comandeer:
Sorry, I've found test with uglify.js and thought that I could do the same with Babili.

I've simplified test. It doesn't require Babili anymore.

Rich-Harris:
Thanks!

Issue #1015, opened 2016-10-02
bundle.generate() produces sourcemap with absolute paths in sources


asapach:
See systemjs/builder#705

Starting with rollup@0.34 sourcemaps produced by `bundle.generate()` contain absolute paths in `sources` instead of relative URLs when input source maps are provided. Repro case: https://github.com/asapach/rollup-sourcemaps
Here's the example in it's entirety:

``` js
var rollup = require( 'rollup' );

rollup.rollup({
  entry: 'test.js',
  plugins: [{
    resolveId: function(id) {
      return id;
    },
    load: function(id) {
      return {
        code: 'export var foo = "bar";',
        map: {"version":3,"sources":["test.js"],"names":["foo"],"mappings":"AAAA,OAAO,IAAMA,MAAM,KAAZ","file":"test.js","sourcesContent":["export const foo = \"bar\";"]}
      };
    }
  }]
}).then(function (bundle) {
  var result = bundle.generate({
    sourceMap: true,
    moduleName: 'test',
    format: 'iife'
  });
  console.log(result.map.sources);
});
```

On my machine this outputs `[ 'C:/tmp/rollup-sourcemaps/test.js' ]`

It looks like this commit (5b4f70725dbee1a78063415adff3aaf70af1f03d) has introduced native `path.resolve()` instead of custom `resolve()` implementation. And now source path is resolved relative to `cwd`: https://github.com/rollup/rollup/blob/c5d9d347721ad3d168b2ca6df27894be69815d1c/src/utils/collapseSourcemaps.js#L121

This issue manifests itself differently when passing the `dest` option: the source path becomes relative, but still **relative to `cwd`**. Is this behavior intentional? How would you recommend handling this in tools that use `rollup` as a library?

Rich-Harris:
Without a `sourceMapFile` or `dest` option (`dest` is used as `sourceMapFile` if the latter is missing), Rollup doesn't have much choice other than to use an absolute path (which isn't as useless as it sounds – subsequent tooling might manipulate the sourcemap further – though perhaps it should have a `file://` protocol prefix, and perhaps it should cause a warning to be emitted).

If you *do* have a `sourceMapFile` or `dest` option, the resulting sources are relative to that destination. So in your example, if we add this line...

```diff
}).then(function (bundle) {
  var result = bundle.generate({
    sourceMap: true,
+   sourceMapFile: 'bundle.js',
    moduleName: 'test',
    format: 'iife'
  });
[...]

PR #1259, opened 2017-01-13
added repro case for #1015

asapach:


guybedford:
Running the test this outputs on my file system `../../../../../../../../../test.js` instead of the depth in the expectation. I suspect this is because Rollup is using path.resolve on the actual file system which is resulting in the backtracking below the filesystem root being ignored.

guybedford:
Marking as closing for inactivity unless we can clarify the use cases further here.

PR #1276, opened 2017-01-19
README: jsnext:main→module

simon04:
Since https://github.com/rollup/rollup/wiki/jsnext:main is just a redirect to https://github.com/rollup/rollup/wiki/pkg.module and the latter states that `module` should be used.


simon04:
Obsolete due to d8ca8685e335dce05f189d2c514a277d41a7db7e.

Issue #1028, opened 2016-10-06
Default exports being included with "export *"


mairatma:
I started getting this `Conflicting namespaces` error after changing a file (let's say **file2.js**) to re-export everything from another file (**file1.js**), by using two export declarations:

``` js
import def from './file1';

export default def;      // Re-exports the default export.
export * from './file1'; // Re-exports all named exports.
```

Rollup then started complaining that there were `Conflicting namespaces` for the default export, as if **file1.js** was exporting it twice. I researched how `export *` should behave, and everwhere I looked it was treated as if it should only export named files, including transpilers like babel (see how they ignore default exports on purpose when converting ExportAllDeclaration [here](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-es2015-modules-commonjs/src/index.js#L36)).

[...]

PR #1279, opened 2017-01-23
fixes #1278

### LINKED ISSUE #1278: re-export default conflict when using asterisk
[This github repo](https://github.com/olsonpm/rollup-reexport-default-conflict) reproduces the issue (the issue is an erroneous warning, not a failed rollup build)

In short, if a module contains multiple `export * from` statements, where the modules being passed through also contain default exports, then the following warning is shown:

> Conflicting namespaces: dep1.js re-exports 'default' from both dep2.js (will be ignored) and dep3.js

The above will be triggered for

```js
// dep1.js
export * from './dep2';
export * from './dep3';


// dep2.js
export default 'this is dep2';

const namedDep2 = 'this is namedDep2';
export { namedDep2 } ;


// dep3.js
export default 'this is dep3';

const namedDep3 = 'this is namedDep3';
export { namedDep3 } ;
```

Note the rollup output is fine, as **the defaults are correctly withheld**.

olsonpm:
- only add named exports to module.exportsAll

\* I didn't add any tests to this since it's only a warning and is really  just an extension of [the fix](https://github.com/rollup/rollup/pull/1038) for #1028

If you would prefer me to write a test and re-send the PR, I will certainly do that.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -314,7 +314,9 @@ export default class Bundle {
 
 				return this.fetchAllDependencies( module ).then( () => {
 					keys( module.exports ).forEach( name => {
-						module.exportsAll[name] = module.id;
+						if ( name !== 'default' ) {
+							module.exportsAll[name] = module.id;
+						}
 					});
 					module.exportAllSources.forEach( source => {
 						const id = module.resolvedIds[ source ];


Rich-Harris:
Thank you! Sorry for the long delay in merging. Just doing a sweep of easily-mergeable/fixable things then will cut a new release

olsonpm:
Not a problem.

Btw, It's very interesting reading about your design decisions in the libraries you publish.  I appreciate the time you take to explain them.

Rich-Harris:
Thanks — I've found it to be a useful process as it forces you to think about the correct vocabulary etc (the old 'if you can't explain it, you don't really understand it')

PR #1295, opened 2017-02-01
Recommend 'module' field over 'jsnext:main'

lukechilds:


btd:
Dup https://github.com/rollup/rollup/pull/1276

lukechilds:
Oops sorry, missed that.

PR #1311, opened 2017-02-22
put parens around sequences if conditional simplified

### LINKED ISSUE #1273: Rollup >=0.37.0 removes parens and generates invalid js
Code instrumented by Istanbul doesn't work with rollup >=0.37, because it started to remove parens in cases like this:

Input file `test.js`:

```js
let i = 0;
const A = true ? (i++, "abc") : undefined;
console.log(A);
```

Rollup 0.36.4 output `$ rollup -i ./test.js`:

```js
let i = 0;
const A = (i++, "abc");
console.log(A);
```

Rollup 0.37.0 output `$ rollup -i ./test.js`:

```js
let i = 0;
const A = i++, "abc";
console.log(A);
```

kzc:
Fixes  #1273

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "homepage": "https://github.com/rollup/rollup",
   "devDependencies": {
-    "acorn": "^4.0.1",
+    "acorn": "4.0.4",
     "buble": "^0.15.1",
     "chalk": "^1.1.3",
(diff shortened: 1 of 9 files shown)


kzc:
@Rich-Harris ping.

Rich-Harris:
thanks 👍

PR #1316, opened 2017-02-26
Allowing logging message to be of string type

diervo:
Before this change, in plugins like this: 
https://github.com/rollup/rollup-plugin-babel/blob/master/src/index.js#L39
will just get `undefined` in the output when running standalone rollup cli

I haven't have the time to to look very in depth at all the code, but I found a couple of places (both in rollup itself and in external plugins) where the caller is passing a string, so since fixing all the plugins is probably a non-starter, I figure making the API support strings was the easiest.

You might want to change the param name of the function or do a more strict type check maybe.


Rich-Harris:
This makes sense, thank you

PR #1317, opened 2017-02-27
Remove redeclared variables - fixes #1268

### LINKED ISSUE #1268: Treeshake: false results in name collisions
Hi. 

When disabling treeshaking using treeshake:false rollup seems to generate name collisions like in the following scenario:

https://gist.github.com/porsager/9436fe8aa2da156d2f9bfc25443dae0a

porsager:


Rich-Harris:
thank you 👍

PR #1318, opened 2017-02-28
README: recommending to specify `pkg.module`

exarus:
The current propsal suggests pkg.module as a key: https://github.com/nodejs/node-eps/blob/master/002-es6-modules.md
I'm sure that pkg.module should be the recommendation but I'm not sure if `jsnext:main` should be removed from README.
But it seems reasonable to discourage using `jsnext:main`.


btd:
Dup of #1276

exarus:
Indeed. I will be more attentive next time. Thanks. 
Going to close this.

PR #1323, opened 2017-03-04
documentation

vijithassar:


Rich-Harris:
Thanks, whoever you are! (jk, we're in a coffee shop updating the documentation here and on rollupjs.org. Sorry I've been absent from this project a while everyone — normal service will resume shortly...)

vijithassar:
it's actually a bar 😉

PR #1327, opened 2017-03-09
Add support for custom finalisers

### LINKED ISSUE #269: Custom output formats?
I had to do something a lil’ dirty in [d3-bundler](https://github.com/d3/d3-bundler). I call bundle.generate with iife format and then mangle the resulting code:

``` js
function patch(code) {
  if (code.slice(0, 37) !== "(function (exports) { 'use strict';\n\n") throw new Error("unexpected output");
  if (outro = code.slice(-26) !== "\n})((this.exports = {}));\n") throw new Error("unexpected output");
  try { var version = require(path.resolve(entry, "../package.json")).version; } catch (ignore) {};
  return "!function(global, exports) {\n"
      + "  \"use strict\";\n"
      + "\n"
      + code.slice(37, -26)
      + (version ? "  exports.version = " + JSON.stringify(version) + ";\n" : "")
      + "\n"
      + "  if (typeof define === \"function\" && define.amd) global." + argv.name + " = exports, define(\"" + argv.name + "\", [], function() { return exports; });\n"
      + "  else if (typeof module === \"object\" && module.exports) module.exports = exports;\n"
      + "  else global." + argv.name + " = exports;\n"
      + "}(this, {});\n";
}
```

I do this because I have “special” requirements for my UMD module, as D3 is a library:
- I want to generate a named AMD module, not an anonymous one.
- I want to export a global even if AMD is available (but not with CommonJS).
- I want to automatically pull the version number from the package.json.

My hack’s not too bad since the iife format is pretty simple, but do you think this is the right way to do it?

### LINKED ISSUE #1326: Add support for custom finalizers
## Current Behavior

There's no external way known to me to specify finalizers or change their output

## Request Behavior

If `options.format` is a function, treat it as a finalizer, should be a 5 mins change [here](https://github.com/rollup/rollup/blob/e955ea76b1c4296c4f58f456642836489f00a176/src/Bundle.js#L476-L486)

steelbrain:
This PR aims to fix #1326
Also partially fixes #269

- [x] implement feature
- [ ] add specs confirming it works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage
 .commithash
 .idea
 bin/rollup
+npm-debug.log
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -473,7 +473,7 @@ export default class Bundle {
 
 		const indentString = getIndentString( magicString, options );
 
-		const finalise = finalisers[ options.format ];
+		const finalise = typeof options.format === 'function' ? options.format : finalisers[ options.format ];
 		if ( !finalise ) {
 			error({
 				code: 'INVALID_OPTION',


steelbrain:
I couldn't find any docs suggesting how to write new specs and the existing ones have a complicated dir structure so any pointers with that would be appreciated

shellscape:
@steelbrain this one has been sitting for a while, but was curious if you were still interested in getting this added.

steelbrain:
My use case no longer needs this, thanks!

Issue #1341, opened 2017-03-18
Allow undefined imports in getInteropBlock


paulfalgout:
Most interopts I've seen, babel, webpack, and rollup's [cjs interop](https://github.com/rollup/rollup/blob/master/src/finalisers/cjs.js#L42) check for the existence of the imported value prior to checking for `default`.

I tried to read the ecma specs but they're pretty confusing.  If an import statement isn't relative, is this expected to throw an error?

```js
import foo from 'foo-does-not-exist';

// or would it just be undefined?
foo === undefined;
```

Even _if_ it would throw an error in ES2015 modules, if I know my target is UMD anyways, and it is an external dependency, why not allow it to import without an error?

FWIW.. the exact same src will work if you're exporting the commonjs, but it will fail for UMD which seems less than ideal.

The solution is also very simple.

Instead of
```js
foo = 'default' in foo ? foo['default'] : foo;
```
 
it'd be
```
foo = foo && 'default' in foo ? foo['default'] : foo;
```

I have a PR ready to go if this is reasonable.

Rich-Harris:
Released 0.43 with the fix

PR #1342, opened 2017-03-23
Allow undefined import in interop

paulfalgout:
Resolves https://github.com/rollup/rollup/issues/1341

Most interopts I've seen, babel, webpack, and rollup's [cjs interop](https://github.com/rollup/rollup/blob/master/src/finalisers/cjs.js#L42) check for the existence of the imported value prior to checking for `default`.

I tried to read the ecma specs but they're pretty confusing.  Is this expected to throw an error?

```js
import foo from 'file-does-not-exist';

// or would it just be undefined?
foo === undefined;
```

Even _if_ it would throw an error in ES2015 modules, if I know my target is UMD anyways, why not allow it to import without an error?

### DIFF
--- a/src/finalisers/shared/getInteropBlock.js
+++ b/src/finalisers/shared/getInteropBlock.js
@@ -11,7 +11,7 @@ export default function getInteropBlock ( bundle, options ) {
 				return `${bundle.varOrConst} ${module.name}__default = 'default' in ${module.name} ? ${module.name}['default'] : ${module.name};`;
 			}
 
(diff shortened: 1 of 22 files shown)


Rich-Harris:
Thanks very much for the PR, this seems like a good change — should also fix https://github.com/rollup/rollup/issues/1385.

We should probably extract this out into a helper function in cases where there are two or more interop checks in a bundle, but that can wait for later.

roastlechon:
@Rich-Harris when is the next release?

PR #1351, opened 2017-03-30
Fix link

veltman:
It looks like `rollupjs.org/guide/` doesn't exist anymore, fixed this link to point at `rollupjs.org` root.


Rich-Harris:
thanks — the site now redirects `/guide` to `/`, but it should be this way anyway

Issue #1274, opened 2017-01-18
TypeError: Cannot read property 'specifier' of undefined, instead of "A module cannot import itself"


coni2k:
When using "rollup" with a combination with a module that is trying to import itself (in my case "breeze-client"), it cannot reach to **"A module cannot import itself"** error message and fails with **"TypeError: Cannot read property 'specifier' of undefined"** error message.

It happens in the following block, at `return declaration.source === source` line:

	if ( resolvedId === module.id ) {
		// need to find the actual import declaration, so we can provide
		// a useful error message. Bit hoop-jumpy but what can you do
		var name = Object.keys( module.imports )
			.find( function (name) {
				var declaration = module.imports[ name ];
				return declaration.source === source;
			});

		var declaration = module.imports[ name ].specifier.parent;

		module.error({
			code: 'CANNOT_IMPORT_SELF',
			message: "A module cannot import itself"
		}, declaration.start );
	}

And the values of these variables:
`declaration.source`: `commonjs-proxy:breeze-client`
`source`: `breeze-client`

This leads to `name` variable becoming null (or undefined?) and then next line fails with **"TypeError: Cannot read property 'specifier' of undefined"** error message:

    var declaration = module.imports[ name ].specifier.parent;

I have created a small demo app that shows the error. I tried to make the app as simple as possible, however I saw your "Issue Guideline" after preparing the application, and unfortunately it still contains typescript and gulp steps to show you the error.

I already spent some time to prepare this demo app but if you think it's still too much, I can even try to make it simpler.

https://github.com/forCrowd/Labs-RollupJS-InvalidAliasesFix

* Download node packages
* Compile typescript files
* Run the the following gulp tasks to see each case:
 * **build-invalidSettings-buggy**: Uses invalid settings (no alias defined for "breeze-client". Rollup should show "A module cannot import itself" error message, but cannot due to "TypeError: Cannot read property 'specifier' of undefined" error.
 * **build-invalidSettings-fixed**: Again uses invalid settings, but experimentally fixed version of rollup and now rollup shows the correct error message.
 * **build-valid**: This tasks is here just to show that rollup works as expected with the correct settings.

Experimental fix that I did is simply changing this line:

    return declaration.source === source;

to this:

    return declaration.source.endsWith(source);

Also the actual error message can be improved by mentioning the module name like this?:

    message: "A module cannot import itself: " + source

Probably this is not correct way of fixing this error but in case if you think it's fine for the moment, I can send you a pull request.

<details>
[...]

PR #1387, opened 2017-04-26
Fix for cannot-import-self check, adding sanity check

enanox:
Hi, this PR is to tackle issue #1274 brought up by @coni2k. 
Using a couple of null checks the verification produces its intended result.
Feel free to point out any issue about this, as I'm not aware of any underlying implications it would have (and it's my first PR here).

Hopefully this makes sense!
Best regards.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -394,12 +394,12 @@ export default class Bundle {
 									return declaration.source === source;
 								});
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you very much!

enanox:
Awesome, thank you for the heads up @Rich-Harris!

Issue #1352, opened 2017-03-31
Prune external dependencies with unused imports


Rich-Harris:
[REPL](https://rollupjs.org/repl?version=0.41.6&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QiUyMHVudXNlZCUyMCU3RCUyMGZyb20lMjAnZXh0ZXJuYWwnJTNCJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlMkMlMjJtb2R1bGVJZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE) — if you import from external modules but don't use the imports, it would be nice to be able to remove the resulting `require` statements from the generated bundle. This would have to be configured, since by default Rollup has to assume that the external modules have side-effects.

rtsao:
I'm interested in helping make this happen. I'd be happy to start working on a PR.

**How should this option be configured?**

I think something like this would be reasonable:
```js
{
  // The default option. Assume all external modules have side effects and don't remove unused imports
  pureExternalImports: false, 

  // Assume all external modules don't have side effects and remove unused imports
  pureExternalImports: true, 

  // Assume all external modules but 'foo' and 'bar' have imports with side effects
  pureExternalImports: {
     pure: ['foo', 'bar'] 
  },

  // Assume all external modules but 'foo' and 'bar' have pure imports
  pureExternalImports: {
     impure: ['foo', 'bar'] 
  },

  // Note: if both 'pure' and 'impure' options are provided, an error is thrown
}
```

Rich-Harris:
Fixed in 0.43 — thanks

rtsao:
Looks like webpack is supporting a `"pure-module"` field in the package.json: https://github.com/webpack/webpack/tree/feature/pure-module/examples/pure-module

PR #1388, opened 2017-04-28
Prune unused pure external imports

rtsao:
Resolves https://github.com/rollup/rollup/issues/1352

I didn't get a chance to fully get familiarized with the codebase so not sure if this is the best way to implement it, but this seems to work. I'll definitely need to add a few more tests, but figured I'd throw this up in the meantime.

I've made this feature  configurable based on the config I described in  https://github.com/rollup/rollup/issues/1352, reproduce below:

```js
{
  // The default option. Assume all external modules have side effects and don't remove unused imports
  pureExternalImports: false, 

  // Assume all external modules don't have side effects and remove unused imports
  pureExternalImports: true, 

  // Assume all external modules but 'foo' and 'bar' have imports with side effects
  pureExternalImports: {
     pure: ['foo', 'bar'] 
  },

  // Assume all external modules but 'foo' and 'bar' have pure imports
  pureExternalImports: {
     impure: ['foo', 'bar'] 
  },
}
```

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -43,6 +43,20 @@ export default class Bundle {
 
 		this.treeshake = options.treeshake !== false;
 
+		if (options.pureExternalModules === true) {
+			this.isPureExternalModule = () => true;
+		} else if (typeof options.pureExternalModules === 'object') {
+			if (options.pureExternalModules.pure) {
+				const pure = new Set([options.pureExternalModules.pure]);
+				this.isPureExternalModule = id => pure.has(id);
+			} else if (options.pureExternalModules.impure) {
+				const impure = new Set([options.pureExternalModules.impure]);
+				this.isPureExternalModule = id => !impure.has(id);
+			}
+		} else {
+			this.isPureExternalModule = () => false;
+		}
+
 		this.resolveId = first(
 			[ id => this.isExternal( id ) ? false : null ]
 				.concat( this.plugins.map( plugin => plugin.resolveId ).filter( Boolean ) )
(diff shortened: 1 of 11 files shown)


elmerbulthuis:
Great! I have been looking for this for a while! As soon as this pr is accepted i will start using rollup!

nathancahill:
This is huge. Very excited to see this.

rtsao:
Would someone be able to review this? I'd be happy to make whatever changes are needed, but I'm not familiar enough with the codebase to know if I'm taking the correct approach.

@eventualbuddha @Victorystick @vijithassar @TrySound @Swatinem

Rich-Harris:
Thanks so much for this PR, and sorry for the long delay in responding.

I think this is a great idea. API-wise, I think it perhaps makes sense to simplify it to this:

```js
{
  // default
  pureExternalImports: false

  // assume all imports are pure
  pureExternalImports: true 

  // foo and bar are pure
  pureExternalImports: ['foo', 'bar']
}
```

My reasoning:

* Specifying which modules are impure is dangerous, because any new dependencies that are added are automatically assumed to be pure, which could result in incorrect code. With `pureExternalImports: true` you're explicitly opting in to that behaviour
*  It optimises for the common case, which is ensuring that `foo` and `bar` don't remain unnecessarily
* Using an object key as a 'flag' is potentially confusing as it *looks* as though you can use them together somehow

If you did need the `impure` behaviour, it could still be achieved with something like this:

```js
pureExternalImports: Object.keys(pkg.dependencies).filter(id => {
  return id !== 'foo' && id !== 'bar';
})
```

Also, a lot of options in Rollup can be expressed as functions — I think we should do the same here:

```js
{
  pureExternalImports: id => id.startsWith('some-prefix-')
}
```

I'll raise a new PR based on this branch — in the meantime, if anyone objects to this plan then please yell at me!

rtsao:
Awesome work, thanks!

kim366:
Where has this option gone? I cannot find it documented on the site and Rollup tells me this is an unrecognized option

kim366:
I see, thanks

Issue #372, opened 2015-12-22
Code splitting


Rich-Harris:
Previously: #120.

Probably not going to happen any time soon, but at some point it would be nice to add support for code-splitting, i.e. finding the common dependencies of `a.js` and `b.js` so that you can do this sort of thing...

``` html
<!-- a.html -->
<script src='shared.js'></script>
<script src='a.js'></script>

<!-- b.html -->
<script src='shared.js'></script>
<script src='b.js'></script>
```

...and let visitors to `b.html` who have already been to `a.html` use a cached copy of `shared.js` (_mutatis mutandis_ for client-side routing situations with an asynchronous module loader).

First hard question is what the API looks like – my initial thought is that if you specify _x_ entry points you should get _x+1_ bundles back, where the _+1_ is the shared code. Very easy to grok and probably meets most use cases, but not as flexible as some other approaches.

Anyway I'll just leave this here for now.

thisconnect:
I'd love to have control over what is shared. I.e `b.html` and `c.html` may share a large part, that I don't want to have in. I belive that people usually like to decide what should be shared and whatnot.

thisconnect:
Related to automatic finding common dependencies, I am currently trying to setup 2 configs to manually control what goes in the shared common lib and what goes into app.

- app1.js
- common/
  - index.js
  - moduleA.js
  - moduleB.js

Trying to keep it as simple as possible, my approach is currently to build 2 iife one as a lib and the app importing from the lib (common/index.js).

common/index.js
```javascript
export * from './moduleA.js';
export * from './moduleB.js';
```
Is there a better way to export all shared modules, than listing all modules and `export * from`? 

app1.js
```javascript
import Lib from './common/index.js';
```

The lib build works fine (entry `common/index.js`, moduleName `Lib`),
but I can't find a way to exclude the Lib from app.

I am not sure if this approach should be working at all or if it would make sence to put all common code into its own node module, see #401

Victorystick:
`common/index.js` should probably export it's submodules' namespaces,
```js
import * as moduleA from './moduleA.js';
import * as moduleB from './moduleB.js';

export { moduleA, moduleB };
```

And then in `app1.js` import,
```js
import { moduleA, moduleB } from './common/index.js';
```

And hack together a plugin that will ensure that `./common/index.js` isn't included in the bundle. I think that should be possible.

thisconnect:
[...]

PR #1389, opened 2017-04-28
Support multiple configs/bundles in parallel

nathancahill:
This PR adds support for rolling multiple bundles (with their own configs, or at least with their own entry/dest options) in parallel. `rollup.config.js` can now export an array of objects, or a single object.

### DIFF
--- a/bin/src/runRollup.js
+++ b/bin/src/runRollup.js
@@ -86,15 +86,21 @@ export default function runRollup ( command ) {
 					}
 				};
 
-				const options = require( config );
-				if ( Object.keys( options ).length === 0 ) {
-					handleError({
-						code: 'MISSING_CONFIG',
-						message: 'Config file must export an options object',
-						url: 'https://github.com/rollup/rollup/wiki/Command-Line-Interface#using-a-config-file'
-					});
-				}
-				execute( options, command );
+				const configs = require( config );
+				const normalized = Array.isArray( configs ) ? configs : [configs];
+
+				normalized.forEach(options => {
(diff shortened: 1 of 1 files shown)


nathancahill:
This is how I manually handle #372 until there is a better solution in Rollup internals.

Rich-Harris:
This is great, thank you! Even works in `--watch` mode 🤘

jhabdas:
@nathancahill If I understand this correctly I should be able to remove some of the [branch logic](https://github.com/jhabdas/fetch-inject/blob/master/rollup.config.js#L20-L29) used in my rollup config, and instead return an array of rollup configs?

jhabdas:
Here's a pattern some may wish to use to leverage this feature to produce minified bundles without duplicating config:

```js
import uglify from 'rollup-plugin-uglify'
import { minify } from 'uglify-es'
const activeConfigs = [ /* array of rollup config objects */ ]
const minifiedConfigs = activeConfigs.reduce(
  (minifiedConfigs, activeConfig) => minifiedConfigs.concat(
    Object.assign({}, activeConfig, {
      plugins: activeConfig.plugins.concat([uglify({}, minify)]),
      dest: activeConfig.dest.replace('js', 'min.js')
    })
  ),
  []
)
export default activeConfigs.concat(minifiedConfigs)
```

ProLoser:
What should we do if we want to build a bunch of assets in parallel but we want to host 1 webserver that contains multiple landing pages that are artifacts of the parallel builds?

Should I just configure ONE of the parallel builds with the webserver? Will livereload be triggered across parallel builds?

Issue #1416, opened 2017-05-26
foo:  - (2 ** 23) causes a operators[this.operator] is not a function


backspaces:
TL;DR: This construct causes a `operators[this.operator] is not a function`: minInt24x: - (2 ** 23)

Details: Place this in the main.js in the repl:
```
const foo = {
  minInt24: 0 - (2 ** 23),
  // minInt24x: - (2 ** 23) // rollup bug operators[this.operator] is not a function
}
export default foo
```
Then remove the comment.

olsonpm:
Thanks for raising that issue.  Definitely a bug in rollup - hopefully the next release will merge my PR

PR #1419, opened 2017-05-30
allow unknown binary operators

olsonpm:
- fix issue #1416
 - the issue exposes rollup's assumption that it knows
   about all binary operators which is untrue for future
   ES versions.  The fix is to return UNKNOWN when
   the operator is not in the hardcoded hash.
 - the test must have the subtraction symbol before the
   exponent operator in order for the
   <Binary Operator>.getValue code to run

### DIFF
--- a/src/ast/nodes/BinaryExpression.js
+++ b/src/ast/nodes/BinaryExpression.js
@@ -33,6 +33,8 @@ export default class BinaryExpression extends Node {
 		const rightValue = this.right.getValue();
 		if ( rightValue === UNKNOWN ) return UNKNOWN;
 
+		if (!operators[ this.operator ]) return UNKNOWN;
+
 		return operators[ this.operator ]( leftValue, rightValue );
 	}
 }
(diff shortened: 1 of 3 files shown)


olsonpm:
I could have added the same UNKNOWN clause to UnaryExpression, but didn't know whether you would prefer to leave code untouched until an issue arises, or fix potential bugs ahead of time.

Rich-Harris:
Thanks! We should also just add support for the exponentiation operator — will do that separately

olsonpm:
@Rich-Harris - personally I would add the same fix to [UnaryExpression](https://github.com/rollup/rollup/blob/master/src/ast/nodes/UnaryExpression.js#L23), but I also understand the "don't fix what ain't broken" mantra.  What do you think?

Rich-Harris:
I think it's probably fine — I'm not aware of any new unary operators that have been proposed. If any *do* get proposed, we'll have months, maybe years of warning

PR #1434, opened 2017-06-04
apply plugins before checking for options.entry

### LINKED ISSUE #1270: Plugins cannot specify the entry option
Here is an example Rollup configuration:

```js
export default {
  plugins: [
    {
      options: function (rollup) {
        rollup.entry = "foo.js";
      }
    }
  ]
};
```

When using `rollup --config` I get the following error:

```
🚨   You must specify an --input (-i) option
```

If I instead use this config, then it works:

```js
export default {
  entry: "foo.js",
  plugins: [
    {
      options: function (rollup) {
        rollup.entry = "foo.js";
      }
    }
  ]
};
```

Basically, it seems that Rollup is checking whether the `entry` option is specified or not, but it does this check before the plugins are run.

I think Rollup should do the check after the plugins run, because a plugin might set the `entry` in an `options` hook.

Rich-Harris:
closes #1270

### DIFF
--- a/bin/src/runRollup.js
+++ b/bin/src/runRollup.js
@@ -247,13 +247,6 @@ function assign ( target, source ) {
 }
 
 function bundle ( options ) {
-	if ( !options.entry ) {
-		handleError({
-			code: 'MISSING_INPUT_OPTION',
-			message: 'You must specify an --input (-i) option'
-		});
-	}
-
 	return rollup.rollup( options )
 		.then( bundle => {
 			if ( options.dest ) {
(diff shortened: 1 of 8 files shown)


evocateur:
How would a plugin that _wants_ `entry` available to its `options()` callback obtain that value?

I'm trying to write an "incremental build" plugin that aborts running if none of the input (all files under `src`, for example) is newer than _existing_ output(s). Having the `entry` available gives me at least half a chance of figuring out the root directory to start glob+statting from.

Rich-Harris:
@evocateur you'll still have that — the only change here is that Rollup will apply plugins before prematurely throwing a 'missing entry' error

evocateur:
Thanks for the clarification!

PR #1436, opened 2017-06-06
Fix magic-string deprecations

Conduitry:
Prevent some warnings related to magic-string deprecations.

### DIFF
--- a/src/ast/Node.js
+++ b/src/ast/Node.js
@@ -68,7 +68,7 @@ export default class Node {
 
 	insertSemicolon ( code ) {
 		if ( code.original[ this.end - 1 ] !== ';' ) {
-			code.insertLeft( this.end, ';' );
+			code.appendLeft( this.end, ';' );
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


kzc:
Just wondering if someone could explain what do the new `magic-string` `overwrite` options `storeName` and `contentOnly` do?
```js
code.overwrite( this.start, this.end, name, { storeName: true, contentOnly: false } );
```
I read the docs in the `magic-string` project but it's still not clear to me.

Rich-Harris:
`storeName` means that the string being replaced should be considered a name that gets included in the sourcemap's `names` array — so if you have code like this...

```js
function foo () {
  throw new Error('oh no! something went wrong');
}
```

```js
code.overwrite(9, 12, 'bar', { storeName: true });
```

...then when you run the code, the sourcemapped stack trace will show that the error originated inside `foo`, not `bar`. [Demo](https://runkit.com/rich/593f4eb9d98c590011ff710d)

`contentOnly` means that the you're not replacing anything that was prepended or appended to the chunk. [Demo](https://runkit.com/rich/593f4fcf0f4e6200123c9d4a)

Rich-Harris:
Thanks @Conduitry

kzc:
@Rich-Harris Thanks for the great explanation!

Issue #1433, opened 2017-06-04
Error quality down?


talmobi:
Error quality has diminished?

version 0.39.0

```bash
$ npm run build:js -s
Error transforming /Users/mollie/work/solar-marketing-site/scripts/app-contact-form.js with 'buble' plugin: Unexpected token (114:58)
SyntaxError: Error transforming /Users/mollie/work/solar-marketing-site/scripts/app-contact-form.js with 'buble' plugin: Unexpected token (114:58)
110 :         )
111 :       ),
112 :       html('.app-solar-contact-form-checkbox__text', opts.text || '')
113 :     ),
114 :     _input = SolarContactFormInput({ type: 'businessId' }):
                                                                ^
    at Parser.pp$4.raise (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:2221:15)
    at Parser.pp.unexpected (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:603:10)
    at Parser.pp.expect (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:597:28)
    at Parser.pp$3.parseExprList (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:2152:16)
    at Parser.pp$3.parseSubscripts (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1741:35)
    at Parser.pp$3.parseExprSubscripts (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1718:17)
    at Parser.pp$3.parseMaybeUnary (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1692:19)
    at Parser.pp$3.parseExprOps (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1637:21)
    at Parser.pp$3.parseMaybeConditional (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1620:21)
    at Parser.pp$3.parseMaybeAssign (/Users/mollie/work/solar-marketing-site/node_modules/acorn/dist/acorn.js:1597:21)
```

```bash
$ npm run build:js 2>&1 -s | wooster

 >> wooster output <<
 SyntaxError: Error transforming ./s/app-contact-form.js with 'buble' plugin: Unexpected token (114:58)

[...]

PR #1451, opened 2017-06-19
better error logging information

talmobi:
issue ref: #1433

Improved the error logging a little bit so that you can see where an error is located.

old (rollup v0.43.0)
```
$ rollup -c -f iife -i src/app.js -o bundle.js
🚨   (buble plugin) Unexpected token (1:24)
undefined (1:24)
```

new (needs to be `npm version patch`ed if merged)
```
$ /Users/mollie/temp/rollup/bin/rollup -c -f iife -i src/app.js -o bundle.js
🚨   (buble plugin) SyntaxError: Unexpected token (1:24)
src/subdir/mod.js (1:24)
````


Rich-Harris:
thank you!

PR #1457, opened 2017-06-20
Fix error when options.external returns true

ksvitkovsky:
Fix for #1264

Although I do not see any issues and all tests pass, please, check that the changes do not cause other issues (it just feels too easy 😄 ). 

Explanation (as clear as I could come up with) is in last commit message.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -59,7 +59,7 @@ export default class Bundle {
 		}
 
 		this.resolveId = first(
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thanks for the PR! Afraid this isn't the correct fix though — the test only passes because `options.entry` is already fully resolved in the tests. Closing in favour of #1477

PR #1463, opened 2017-06-22
Run same options transform on generate for bundle

diervo:
Some recent changes introduced in `0.43.0` breaks the amd named module generation.

The checkOptions function is applied only on the main entry point (rollup), but not in the generate() function.

Basically when passing this options:
```js
// Passing a set of options
rollup({ entry, plugins, ..otherOptions }).then(() => {
    ... 
   // Passing another set of options
    bundle.generate({ format: 'amd', moduleId: 'something' });
    ...
```
The output does respect the `moduleId` option, nor does it error out.

This change addresses this inconsistency, and it will give the same warning.

We tried to add a tests but the tests makes the assumption that both entry points rollup and bundle will receive the same options object, which is not necessarily imposed in the API.

//cc @pmdartus who coPR this.

### DIFF
--- a/src/rollup.js
+++ b/src/rollup.js
@@ -44,15 +44,7 @@ const ALLOWED_KEYS = [
 	'watch'
 ];
 
-function checkOptions ( options ) {
-	if ( !options ) {
-		throw new Error( 'You must supply an options object to rollup' );
-	}
-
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you!

PR #1464, opened 2017-06-23
update uglify-js to v3

FelixRilling:
This PR updates Uglify-JS from v2 to v3, while also adjusting the code to conform the new API.
Uglify-JS is only used for internal testing, namely in two `_config.js` files.
All tests work after updating; `npm run lint` runs with success.


Rich-Harris:
Thanks! 👍

PR #1470, opened 2017-07-01
Fix closure memory leak in Bundle constructor

dmitrage:
Should fix https://github.com/rollup/rollup/issues/883

"options" argument is captured by closures in Bundle constructor, so passing "cache" leads to chain of bundles in memory. 

Similar bug is described here: https://blog.meteor.com/an-interesting-kind-of-javascript-memory-leak-8b47d2e7f156

I've created gist with test script, but don't known how to properly integrate it here.
https://gist.github.com/dmitrage/fff3d30280688356194c137323ea4e02

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -70,10 +70,11 @@ export default class Bundle {
 		this.hasLoaders = loaders.length !== 0;
 		this.load = first( loaders.concat( load ) );
 
-		this.getPath = typeof options.paths === 'function' ?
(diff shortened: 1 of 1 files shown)


Rich-Harris:
tim-and-eric-mind-blown

You hero! Been scratching my head over this a while — thanks, this is incredibly useful. Have added your gist as a test.

hugomrdias:
Thank you !!!!

Issue #827, opened 2016-08-04
format:iife's output is inconsistent


bfred-it:
`iife` sometimes overrides the chosen `moduleName`, sometimes extends it.
## Override ([rollupjs.org demo](https://rollupjs.org/repl?version=0.43.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMCgpJTIwJTdCJTdEJTNCJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmlpZmUlMjIlMkMlMjJtb2R1bGVOYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTJDJTIybW9kdWxlSWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==))

``` js
export default function () {};
```

``` js
var myBundle = (function () {
    'use strict';
    function main () {};
    return main;
}());
```
## Extend ([rollupjs.org demo](https://rollupjs.org/repl?version=0.43.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwZm9vJTIwJTNEJTIwMSUzQiUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJpaWZlJTIyJTJDJTIybW9kdWxlTmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCUyQyUyMm1vZHVsZUlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=))

``` js
export const foo = 1;
```

``` js
(function (exports) {
    'use strict';
[...]

PR #1472, opened 2017-07-03
[new feature] extend option

Andarist:
Fixes https://github.com/rollup/rollup/issues/827

This addresses inconsistent behaviour between `named` and `default` output. Ideally whenever possible IIFE should output a simple variable, not extending a global object.

As extending global still has its use cases this PR allows for that with a new option - `extend`, also 'namespaced' modules are preserving the old behaviour, as it is not possible to achieve a simple variable for them anyway.

When implementing this Ive also realised that the same thing should be implemented for `umd` and I can do that, aint sure if that should be a matter of another PR or this one. **NOTE:** umd change would affect d3, its however easily fixable - first of all we can add a warning about the changed behaviour, second of all we can target d3 specifically and add `extend` option for them automatically (for a 1 minor version ofc)

Also I was not sure if I should maybe throw when validating the options object - `extend` now is only implemented for `iife` case, and for other module formats like `es` it doesnt even make sense, so maybe the user should be warned that he uses obsolete option.

Didn't yet include any docs about this, can do that. However would prefer to add this at the end, when PR gets approved.

cc @bfred-it @TrySound @mbostock @Victorystick


bfred-it:
> the same thing should be implemented for `umd` and I can do that, aint sure if that should be a matter of another PR or this one

They should probably both behave the same way with globals, so it's best to also change UMD's behavior.

> we can target d3 specifically and add `extend` option for them automatically

I'd rather not have special _behavior_ like that for others. A special _warning_ would be enough. 

> for other module formats like `es` it doesnt even make sense

Right, `moduleName` and `extend` only apply to the IIFE bundle.

Andarist:
> They should probably both behave the same way with globals, so it's best to also change UMD's behavior.

Ok. I'll work on that one soon too, will just add commits to this PR.

Victorystick:
Nice work, but it doesn't look quite ready yet. Take the following trivial example:

```js
export var a = 1;
```

The resulting bundle would leave `myBundle` without a value, since none is returned by the function. You need to end the bundle with a `return exports;`.
```js
var myBundle = (function (exports) {
	'use strict';

	exports.a = 1;
}({}));
```

Andarist:
oh damn, didnt spotted this one! gonna fix it ofc, thanks for the review

Andarist:
Ok, the bug spotted by @Victorystick is fixed. UMD tweaks are also done.

While working on this though I've spotted that [this](https://github.com/rollup/rollup/blob/ef59ea2d8979570aeca6c434986ef1504697fce9/test/form/umd-noconflict/_expected/umd.js#L6) is not working properly. It suffers from the same problem that @Victorystick pointed out to me - nothing is returned from the closure. 

So I fixed that issue too.

Please review the PR and let me know if there is anything else I can do here. I would personally add some warnings (maybe even errors?) for options combinations that are not supposed to work together, such as:
- extend + format other than umd or iife
- umd + noConflict + exportMode: 'none'
Aint sure where such things are usually validated though.

TrySound:
@Andarist Did you see my review?

Andarist:
@TrySound unfortunately not, maybe you have forgotten to send it? I often forget that this new~ review tool doesnt send comments immediately

TrySound:
You are right :)

src/utils/identifier-helpers.js
```diff
@@ -6,13 +6,26 @@ const builtins = 'Infinity NaN undefined null true false eval uneval isFinite is
...
 	str = str
 		.replace( /-(\w)/g, ( _, letter ) => letter.toUpperCase() )
-		.replace( /[^$_a-zA-Z0-9]/g, '_' );
+		.replace( illegalCharacters, '_' );
 
-	if ( /\d/.test( str[0] ) || blacklisted[ str ] ) str = `_${str}`;
+	if ( startsWithDigit(str) || blacklisted[ str ] ) str = `_${str}`;
 
 	return str;
 }
```
Andarist: k, should I maybe rename [map-helpers](https://github.com/rollup/rollup/blob/ef59ea2d8979570aeca6c434986ef1504697fce9/src/utils/map-helpers.js) too?

TrySound: Would be great :)

Andarist:
@TrySound I've adjusted the code with your suggestions. Please take a look.

Rich-Harris:
thanks! releasing as 0.44

dnoyes:
@Rich-Harris @Andarist is there somewhere I can go to see example usage?

Andarist:
example usage is just passing `extend: true` to the config options, unfortunately I didnt test it properly with CLI usage and its not working when passed from CLI arguments, fix is on the way though - https://github.com/rollup/rollup/pull/1482

Issue #1474, opened 2017-07-05
Async transformBundle plugins


caccialdo:
Although the current version of rollup allows for `transform` method returning a `Promise`, the `transformBundle` is currently synchronous only.

cf. https://github.com/rollup/rollup/blob/master/src/utils/transformBundle.js#L11

I have a scenario where I need to run closure compiler (the Java version) in a custom plugin, over hundreds of different bundles. `transformBundle` needing to be synchronous, I'm calling the `.jar` file using `require("child_process").spawnSync`. It works but I cannot easily make the bundling tasks run in parallel.

Is there a reason for having limited the Promise handling to `transform` as part of https://github.com/rollup/rollup/pull/263 ?

Rich-Harris:
Released as 0.45

dnoyes:
After upgrading from 0.44.0 to 0.44.5, I'm now getting:

`> grunt rollup
Running "rollup:target" (rollup) task
Warning: bundle.generate(...) now returns a Promise instead of a { code, map } object Use --force to continue.

Aborted due to warnings.`

Here's the pertinent Gruntfile info:

`
   var babel = require("rollup-plugin-babel");

      rollup: {
            options: {
                format: "umd",
                moduleName: "...",
                plugins: function() {
                    return [
                        babel({
                            presets: [["env", {modules: false}]],
                            plugins: ["external-helpers"]
                        })
                    ];
                }
            },

    grunt.loadNpmTasks("grunt-rollup");
`

And my package.json
`
    "babel-cli": "^6.24.1",
    "babel-plugin-external-helpers": "^6.22.0",
    "babel-polyfill": "^6.23.0",
    "babel-preset-env": "^1.5.2",
    "grunt": "^1.0.1",
    "grunt-rollup": "^1.0.1",
    "rollup": "^0.44.0",
    "rollup-plugin-babel": "^2.7.1",
    "rollup-plugin-node-resolve": "^3.0.0",
`

Are there tips/instructions of what I need to change after upgrading to 0.45.0?

I'm running node 8.1.3 and npm 5.0.3

Rich-Harris:
Unfortunately grunt-rollup will need to be updated to reflect the new API. Specifically, [this](https://github.com/chrisprice/grunt-rollup/blob/master/tasks/rollup.js#L79) would need to change to something like this:

```diff
return rollup.rollup({
  entry: entry,
  external: options.external,
  plugins: plugins,
  context: options.context,
[...]

PR #1475, opened 2017-07-07
Support for async transformBundle

caccialdo:
Attempt at solving the https://github.com/rollup/rollup/issues/1474 issue.

A real-life scenario where this could be useful is if you need to build the same code (same entry) with different parameters and minify the resulting bundles with Google Closure Compiler.

Because Google Closure Compiler is a java program, it needs to be started via a `spawn` call. At the moment, rollup only allows for synchronous `transformBundle` methods with means that we need to use the sync version of `spawn`: `spawnSync`.

This change essentially wraps the `transformBundle` calls into promises, and propagate the behaviour all the way up to the `generate` function. As a result, this means changing the signature of `generate` to return a promise, making this feature a **breaking change**.

I guess the code could be written to be backward compatible but that would mean:

- promise duck-typing the return value of `transformBundle()`.
- making the code more complex with extra branches.
- confusing the user with two possible return types.

### DIFF
--- a/bin/src/runRollup.js
+++ b/bin/src/runRollup.js
@@ -73,10 +73,11 @@ export default function runRollup ( command ) {
 			onwarn: handleWarning
 		})
 			.then( bundle => {
-				const { code } = bundle.generate({
+				return bundle.generate({
 					format: 'cjs'
 				});
-
+			})
+			.then( ({ code }) => {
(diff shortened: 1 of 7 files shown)


Rich-Harris:
Thanks! I added getters for `code` and `map` to the returned promised that throw an error explaining that it's now asynchronous, to try and minimise confusion.

PR #1482, opened 2017-07-10
Fixed --extend flag not being passed correctly from the CLI arguments

Andarist:
Aint sure how to add test for this, although I still feel that this should be released asap as patch version and the test can be added in another PR - if you guide me a little I can provide that ;)

cc @Rich-Harris @bfred-it


Rich-Harris:
thanks!

Andarist:
Any tests that should be added for this? I was not sure where to put those.

Issue #1486, opened 2017-07-10
Error occurs when using `webpack` loader query string as import in rollup


roastlechon:
TypeError occurs when using `webpack` loader query string as import in `rollup`.

```
Cannot use ‘in’ operator to search for ‘default’ in /assets/file-loader-imported-filename.9b951a6598d24f97acccd61c43d8df97.js 
```

Our use case is using a `webpack` loader import like:

```
import mosaic from '!!file-loader?publicPath=/assets/&name=file-loader-imported-filename.[hash].[ext]!@external-module';
```

Because of the interopBlock looking for `default` via `in` keyword, it throws if its a string.  To compromise, I tested and created a PR that uses hasOwnProperty instead to ensure both worlds work.

paulfalgout:
The bugfix is released, correct?

roastlechon:
Looks released as of v0.45.1

PR #1487, opened 2017-07-10
Error occurs when using webpack loader query string as import in rollup

roastlechon:
Addresses #1486 

Fixes the issue with TypeError that occurs when using `'default' in string`


Rich-Harris:
thank you!

Issue #408, opened 2016-01-01
Suppress warning messages


guybedford:
I'm using that `return false` in the resolveId function now and it's working great.

Only it seems to give a message each time _Treating 'jspm_packages/npm/babel-runtime@6.3.19/helpers/classCallCheck' as external dependency_.

It would be great to be able to disable these messages via a `rollup({ warn: false })` option or similar in https://github.com/rollup/rollup/blob/master/src/utils/defaults.js#L32.

Rich-Harris:
You can supply a custom `onwarn` function that will intercept warnings:

```js
var warnings = [];

rollup.rollup({
  entry: 'src/main.js',
  onwarn: function ( message ) {
    if ( /external dependency/.test( message ) ) return;
    console.error( message );
  }
}).then(...);
```

Just realised that isn't in the wiki – will update it.

guybedford:
Excellent, thanks!

ibqn:
@Rich-Harris is there a way to call the default `onwarn` implementation after processing some filtering in the overwritten `onwarn` function?

Because, at the moment when I use the custom `onwarn` function I lose default coloring and warning message format :-(.

Rich-Harris:
@ibqn not currently, no. The source code is [here](https://github.com/rollup/rollup/blob/master/bin/src/logging.js) if you want to crib it. not sure what the best way to expose it would be

ibqn:
@Rich-Harris This is exactly the way I do it here for [Angular2-quick-start](https://github.com/ibqn/ng2-quick-start/commit/d5eb2de849b0273ce9d8510d7d80a25c701de3bf) however this look very ugly;)

Suddenly, this is a well known warning message, which appears during the rollup of the AOT compilation with [Angular2](https://angular.io/docs/ts/latest/cookbook/aot-compiler.html#!#run-rollup).

It would be great if something like this were possible in the `rollup.config.js` file:

```javascript
import { handeWarning } from 'rollup';
...
const warnFilter = /The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten/;

export default {
  ...
  format: 'iife',
  onwarn: warning => {  // overwite the default warning function
      const str = warning.toString();

      if (warnFilter.test(str)) return;
      handeWarning(warning);
  },
  ...
}
```

Rich-Harris:
Maybe an alternative would to be make the functions chainable:

```js
export default {
  ...
  format: 'iife',
[...]

PR #1491, opened 2017-07-13
[WIP] Better warnings etc

Rich-Harris:
Rollup can be a little overzealous with warnings sometimes, but using `--silent` usually isn't ideal. Planning to improve the experience a bit in a few ways:

## Batch warnings

We can group warnings by type, by plugin and by file. Here's a before and after (everything here subject to change...):

screen shot 2017-07-13 at 12 02 35 pm

screen shot 2017-07-13 at 12 06 09 pm

## Allowing custom handlers to use default behaviour

Reimplementing the standard `onwarn` function is no fun. I'm planning to implement this:

```js
onwarn(warning, warn) {
  if (warning.code === 'SOMETHING_I_WANT_TO_IGNORE') return;
  warn(warning);
}
```

`warn` is the default batching behaviour screenshotted above.

## Replace terminal contents when watching (maybe)

At present, any warnings etc are just appended to the terminal, and frankly it can get a bit out of hand. Am wondering about more dashboard-like behaviour when watching — i.e. it fills the terminal tab, and output from previous bundles is replaced when a new bundle is generated. Any thoughts?


jonataswalker:
This will be just great! Thank you.

Rich-Harris:
As noted in https://github.com/rollup/rollup/issues/408#issuecomment-315481848, it might be a good idea to suppress unresolved dependency warnings if a) the dependency is a Node built-in, and b) the target format is `cjs`.

Rich-Harris:
Ok, so this PR got a bit out of hand — a combination of feature creep and distraction. Apologies to anyone trying to keep track.

The big change here is the Rollup now includes a `rollup.watch` method in core, rather than via the [rollup-watch](https://github.com/rollup/rollup-watch) package which will be deprecated. It was necessary to make some changes to the design of rollup-watch in order to batch warnings, and I couldn't think of a particularly good reason to keep the packages separate. Merging them will make development easier and reduces friction.

`rollup.watch` should be considered experimental for the time being.

Rather than splurging out a never-ending stream of warnings, `rollup -w` will now clear the terminal each time the bundle is rebuilt (or bundles, if you have an array of configs). The output is much, much easier to read now, in my opinion. (You may disagree. Let me know! We will fix it.)

Couple of other changes:

* `options.format` is now required. `format: 'es6'` is now an error (it should be `es`)
* The `onwarn` handler, if supplied, receives the default `onwarn` as its second argument. This makes it much easier to filter out unwanted warnings (like `THIS_IS_UNDEFINED`)
* Probably some other stuff that I've forgotten

Will cut a new release (0.46) soon.

jacobmischka:
Could we please get a flag to disable the output coalescence when watching? Or possibly checking if it's in a tty so something like `| cat -` would work to disable it?

I currently run several watchers at once using an `npm-run-all -p` command, so right now webpack output is being eaten by rollup.

Thanks!

arxpoetica:
@Rich-Harris exactly what @jacobmischka said. I run multiple watchers, and rollup now makes all the other text disappear. There needs to be a flag. Rollup is messin' w/ my mojo. ;)

arxpoetica:
One more thingy. It now appears that Rollup won't allow one to `console.log`—I'm assuming it's this feature. I don't have a mechanism now for verifying that I'm getting the right variables into Rollup. Halp!

paulsmirnov:
What about "Allowing custom handlers to use default behaviour" part, isn't it done yet? I thought that was a part of PR.

PR #1494, opened 2017-07-19
Semicolon Removal Fix

### LINKED ISSUE #1275: compilation errors caused by rollup deleting minimized firebase semicolon
I have a firebase project that rollup is failing to build. I determined that rollup was corrupting the auth.js file out of firebase by removing meaningful semicolon. Rolling back rollupjs to version 0.36.1 "fixed" me. You probably want to look at this. `g+oe|0a.h[0]` is the broken substr. 
 Minimized rollup.config.js follows:

```js
import replace from 'rollup-plugin-replace'
export default {
    entry:  'node_modules/firebase/auth.js',
    dest:   './dist/auth.js',
    format: 'iife',
plugins:
   [
      replace({
         'process.env.NODE_ENV': JSON.stringify( 'production' )
      })
   ]
}
```

jeffjewiss:
I'm not sure if I went about writing the tests the correct way. I'm interested in helping out with rollup and thought this would be a good, small way to get started. I used the tests linked in the issue as a reference. Please let me know if I should rename anything or if the tests aren't sufficient. Also, I wasn't sure where to find a reference of unconditionally adding parens (as per Buble) to give that a try instead of the newline, but I'm open to trying that instead.

Cheers

Fixes #1275

### DIFF
--- a/src/ast/nodes/VariableDeclaration.js
+++ b/src/ast/nodes/VariableDeclaration.js
@@ -98,7 +98,7 @@ export default class VariableDeclaration extends Node {
 			const needsSemicolon = !forStatement.test( this.parent.type );
 
 			if ( this.end > c ) {
-				code.overwrite( c, this.end, needsSemicolon ? ';' : '' );
+				code.overwrite( c, this.end, needsSemicolon ? ';' : '\n' );
 			} else if ( needsSemicolon  ) {
 				this.insertSemicolon( code );
 			}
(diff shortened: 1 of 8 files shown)


Rich-Harris:
Thank you! This is great. I made a [small tweak](https://github.com/rollup/rollup/commit/c63782ef6725735600ae83c5ca4dcb4dec787eae) — it preserves the semi-colon if the var declaration is the body of the loop (since that's the only time it could be a child of the for loop node but *not* be in the head, which is what gave us the false positive), rather than adding the newline. The effect is the same

PR #1510, opened 2017-08-02
Descriptive error message for invalid format type.

chyzwar:
I just spent 30 minutes because I misspelled a clj (cljs). This error message would make easier to figure out what option is wrong and what is current vs expected value.

Maybe this message can be a bit better but I do not know wider context that well.


Rich-Harris:
thank you!

PR #1515, opened 2017-08-04
Fix issue : In multiple configurations context, no bundle is created

CodeTroopers:
This modifications fix the issue but need improvement to log the success of bundle generation and fix the UnhandledPromiseRejectionWarning.
Your help would be appreciated.


CodeTroopers:
fix #1515

Rich-Harris:
thanks 👍

PR #1534, opened 2017-08-11
allow passing interop through cli - true/false

### LINKED ISSUE #1524: Allow removing interop from CLI
hi! :wave:

Just thinking can you allow removing the interop from the CLI, like `--no-interop`? It seems it's not possible now. I don't know to use rollup config file, since i have dozen of other dot and config files in each repo, don't won't one more for absolutely nothing so serious, also i just not using any plugins, that's why i want to use just the cli.

I prefer to not removing the interop, if i could prepend `/* istanbul-ignore-next */` to it, but using intro doesn't help because adds new line after it. And really, that's the only thing that i need make two bundles - cjs and es, no plugins and things, my packages require node 6.

Of course i can create a plugin that do that, but i don't want config file. All that is related to what i said before some months in another discussion - we don't need config file for some simple things and the tools should force you to do that. That's one of the reasons why i started building a wrapper around Rollup - [@rolldown](https://github.com/charlike/rolldown).

edit: fixed link, forgot that i deleted the org.

charlike:
fixes #1524 , sorry for the other stuff, didn't disable prettier in time. but seems to fail. Rollup does not support `--no-` prefix?

also, there was 2 failing tests before i start.

### DIFF
--- a/bin/src/help.md
+++ b/bin/src/help.md
@@ -28,6 +28,7 @@ Basic options:
 --outro                  Content to insert at end of bundle (inside wrapper)
 --banner                 Content to insert at top of bundle (outside wrapper)
 --footer                 Content to insert at end of bundle (outside wrapper)
+--interop                Include interop block (true by default)
 
 Examples:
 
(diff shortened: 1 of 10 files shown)


Rich-Harris:
Thanks — the test failure was just some minor whitespace stuff

charlike:
cool, thanks too! :taco:

PR #1558, opened 2017-08-16
use Promise.resolve in mapSequence, to ensure each item is thenable

### LINKED ISSUE #1532: TypeError: Cannot read property 'then' of undefined
```
TypeError: Cannot read property 'then' of undefined
    at next (/var/www/FileBrowser/node_modules/rollup/dist/rollup.js:148:22)
    at /var/www/FileBrowser/node_modules/rollup/dist/rollup.js:152:48
    at <anonymous>
```

Rollup: `0.46.1`

This happens with `watch` option at the first modification.

Rich-Harris:
fixes #1532

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rollup",
-  "version": "0.47.1",
+  "version": "0.47.4",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


jonataswalker:
Many thanks for the fix.

PR #1571, opened 2017-08-21
Make sure Task.run returns a Promise

marijnh:
Earlier in the file, there's a call `mapSequence(this.tasks, task => task.run())`, which was crashing for me (when calling `watch` directly, not going through the command-line tool) because of the short-circuit return in `run` returning undefined. I don't really know what I'm doing here, since I only briefly looked at the code, but returning an empty promise seems like it should address this, and indeed it stops my use from crashing.

### DIFF
--- a/src/watch/index.js
+++ b/src/watch/index.js
@@ -160,7 +160,7 @@ class Task {
 	}
 
 	run() {
-		if (!this.dirty) return;
+		if (!this.dirty) return Promise.resolve();
 		this.dirty = false;
 
 		const options = Object.assign(this.inputOptions, {


Rich-Harris:
That's weird — are you on the latest version? This *looks* like the same bug as #1532, and so it should have been fixed by #1558. I could easily be wrong though

marijnh:
Huh, I think I ran the failing code only against 0.46.3, made the fix in `node_modules`, and only then created a patch against the current code, so yeah it's quite likely that I'm fixing a problem that is already fixed in a different way. Sorry for the noise!

PR #1574, opened 2017-08-23
Do not open alternate buffer during piped --watch

evocateur:
Recently, the output of `rollup --watch` was modified to reset the buffer when changes caused a rebuild. Works brilliantly when invoked in an active terminal, e.g. `rollup -cw`.

However, this really doesn't play nicely with processes that pipe the output from a child process (such as [`lerna exec`](https://github.com/lerna/lerna#exec)), which is effectively a _non_-interactive terminal. Clearing the screen repeatedly causes a bunch of thrashing, as well as unwanted removal of other process logging.

### Tests?!

I was unable to locate examples of testing both CLI _and_ watching, so I did a bunch of local testing with npm-linked rollup in several lerna repos where I was using rollup. Here's an example repo I whipped up that demonstrates the problem and solution:

https://github.com/evocateur/lerna-rollup-example

### DIFF
--- a/bin/src/run/alternateScreen.js
+++ b/bin/src/run/alternateScreen.js
@@ -1,0 +1,33 @@
+import ansiEscape from 'ansi-escapes';
+import { stderr } from '../logging.js';
+
+const SHOW_ALTERNATE_SCREEN = '\u001B[?1049h';
+const HIDE_ALTERNATE_SCREEN = '\u001B[?1049l';
+
+export default function alternateScreen ( enabled ) {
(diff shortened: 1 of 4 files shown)


evocateur:
Holy cow, that AppVeyor build has been running for over 30 minutes on node 4. That's not ideal :/

Rich-Harris:
aaaahhhh, that's much cleaner. Thank you 👍

evocateur:
Cheers!

Issue #1596, opened 2017-08-30
bin/rollup ignores watch options


btakita:
`inputOptions` from `function mergeOptions` in `bin/rollup` does not include the watch member.

On linux platforms, chokidar stops watching after the first change of a file. This is fixed by using the `usePolling: true` chokidar config.

https://github.com/paulmillr/chokidar/issues/556

I will create a PR to fix this issue.

Also related: https://github.com/rollup/rollup-watch/issues/51.

PR #1597, opened 2017-08-30
mergeOptions inputOptions includes config.watch option.

btakita:
Fixes https://github.com/rollup/rollup/issues/1596

The following option now applies.
`
watch: {
  chokidar: {
    usePolling: true
  }
}
`


Rich-Harris:
thank you!

Issue #1600, opened 2017-08-30
New config API must specify output file and ignores -o


bjornharrtell:
Before the new config API I used a config without specifying dest, instead specifying it with the -o flag when running rollup with the config. When trying the same with a newer version that has the new config API it complains with unrelated `Error: You must specify options.format, which can be one of 'amd', 'cjs', 'es', 'iife' or 'umd'`.

shellscape:
Looks like this was fixed by #1606. If this behavior is still observed, please let us know and we'll reopen the issue.

PR #1606, opened 2017-09-01
merge command & config's output options

### LINKED ISSUE #1573: -o from CLI breaks output.format in rollup.config.js with rollup 0.48.2
After upgrading to rollup 0.48.2 I tried to migrate my rollup.config.js following the instructions from the gist (https://gist.github.com/Rich-Harris/d472c50732dab03efeb37472b08a3f32).

Despite following the example from the gist, I'm getting the error `[!] Error: You must specify options.format, which can be one of 'amd', 'cjs', 'es', 'iife' or 'umd'`

Basically, my rollup config looks like this:

```
let config = {
  input: 'src/index.js',
  output: {
    format: 'umd',
    name: 'Test123',
    sourcemap: true
  }
};

export default config;
```

And I'm using a package.json script like this:

```
"scripts": {
  "deploy": "rollup -c -o dist/bundle.js"
}
```

Repro repo here: https://github.com/jharris4/rollup-048-config-bug

sunnylost:
Fix #1573 , #1600 

Both command line and config file can set `output` option, they should be merged but instead of used separately.

### DIFF
--- a/bin/src/run/mergeOptions.js
+++ b/bin/src/run/mergeOptions.js
@@ -73,14 +73,21 @@ export default function mergeOptions ( config, command ) {
 		paths: getOption('paths')
 	};
 
-	const outputOptions = (
-		(command.output || config.output) ?
+	let mergedOutputOptions;
+	if (Array.isArray(config.output)) {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
thank you!

PR #1620, opened 2017-09-09
Print timestamps on "END" events.

jacksonrayhamilton:
Previously, it was not clear when work had been completed.  This sets clearer expectations for someone monitoring the watcher output as to when they can expect their code to have updated.

### DIFF
--- a/bin/src/run/watch.js
+++ b/bin/src/run/watch.js
@@ -3,6 +3,7 @@ import * as rollup from 'rollup';
 import chalk from 'chalk';
 import ms from 'pretty-ms';
 import onExit from 'signal-exit';
+import dateTime from 'date-time';
 import mergeOptions from './mergeOptions.js';
 import batchWarnings from './batchWarnings.js';
 import alternateScreen from './alternateScreen.js';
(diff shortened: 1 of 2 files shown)


bin/src/run/watch.js
```diff
@@ -76,7 +76,10 @@ export default function watch(configFile, configs, command, silent) {
 					break;
 
 				case 'END':
-					if ( !silent && isTTY ) stderr( `\nwaiting for changes...` );
+					if ( !silent && isTTY ) {
+						stderr( `\nfinished at ${new Date()}` );
```
evocateur: The default Date format is rather ...quaint?

```sh
$ node -p "new Date().toString()"
Fri Sep 15 2017 15:13:26 GMT-0700 (PDT)
```

Why does this have to be across multiple lines, anyway? `[YYYY-MM-DD HH:MM:SS] waiting for changes...` seems more than sufficient. [date-time](https://www.npmjs.com/package/date-time) looks like a good library for the formatting.

jacksonrayhamilton: I moved the timestamp to the "waiting" line and reformatted it.

Rich-Harris:
great idea, thank you!

PR #1648, opened 2017-09-24
Remove conditional expressions if they have no effect.

corneliusweig:
This is a small improvement to the logic of evaluating ConditionalExpression.
Before, the conditional expression would always be kept in full, unless the testValue was known. This behavior is changed as follows:
- if the test evaluation produces an effect, keep the whole expression
- if the test evaluation is unkown and has no effect, remove the expression unless some branch has an effect
- if the test evaluation is known, replace the expression with the resulting branch (as before)


corneliusweig:
Has been combined with #1650.

PR #1698, opened 2017-11-05
Remove line offset in public error object

### LINKED ISSUE #1697: error.loc.line in public API does not match with the real file
I'm currently working on a [PR](https://github.com/facebook/react/pull/11456) on facebook/react.
What I realized during this course of work is that the `error.loc.line` in the `error` object thrown is not consistent with the `error.loc.file` itself.
To my understanding, it's mainly due to tracing sourcemaps within rollup. [`props.loc`](https://github.com/rollup/rollup/blob/b949eb08169115ff66648838cbc4833379bf9440/src/Module.js#L295)
However, I think that the `error.loc.line` thrown through the public API should be consistent with the `error.loc.file`

Below is a minimal log from reproduction.
It shows that `error.loc.line` given through the public API is **14** while in the real file it is on line **13**.
```
error.message:
 'forEach' is not exported by packages/react/src/ReactChildren.js
error.loc:
 { file: '/Users/soojae/Documents/react/packages/react/src/React.js',
  line: 14,
  column: 9 }
-- MISSING_EXPORT (undefined) --
'forEach' is not exported by packages/react/src/ReactChildren.js
  11 |
  12 | import {Component, PureComponent, AsyncComponent} from './ReactBaseClasses';
> 13 | import {forEach, map, count, toArray, only} from './ReactChildren';
     |         ^
  14 | import ReactCurrentOwner from './ReactCurrentOwner';
  15 | import {
  16 |   createElement,
error Command failed with exit code 1.
```

misoguy:
Fix #1697

### DIFF
--- a/src/Module.js
+++ b/src/Module.js
@@ -291,8 +291,9 @@ export default class Module {
 			props.pos = pos;
 
 			const { line, column } = locate( this.code, pos, { offsetLine: 1 } ); // TODO trace sourcemaps
-
-			props.loc = { file: this.id, line, column };
+			const lineInRawFile = line - 1;
+			
+			props.loc = { file: this.id, line: lineInRawFile, column };
 			props.frame = getCodeFrame( this.code, line, column );
 		}
 


misoguy:
Seems like changing the props itself breaks many tests. I'm closing this PR to find a better solution.

PR #1745, opened 2017-11-20
Make .intro, .outro, .banner, .footer to return promises

btd:
Reworked version of https://github.com/rollup/rollup/pull/1253 with adding more test cases to cover more combinations.

There is still missing one thing. How to report rejected promises?


btd:
//cc @lukastaegert 
By mistake closed previous.

lukastaegert:
Without taking a look: IMO printing the error message to stderr and aborting the build should be a sensible course of action.

btd:
Added error generation (with `error` function already provided). Message from looks like this:
default
I am not sure how to write test for this.

PR #1796, opened 2017-12-15
Reexport aliasing bug

guybedford:
This tests the remaining bug from https://github.com/rollup/rollup/issues/1794#issuecomment-351728892, not already solved by https://github.com/rollup/rollup/pull/1795.

~~Fix still pending though....~~


guybedford:
Found the fix here and included it in the PR.

PR #1801, opened 2017-12-19
Refactoring - WIP [do not merge]

guybedford:
In working on some of this dynamic import analysis and chunking code, it ends up that a lot of the graph analysis that is already in Rollup itself is being done twice, resulting in extra unnecessary build costs in double parsing etc.

What I've found is having direct access to the Bundle interface can share this analysis quite well.

This is my current refactoring of Rollup to handle this approach, separated from the dynamic import PR, which I've attempted to create as separate logical commits as much as possible with simple conceptual steps.

I'd like to keep this open as a PR so that it is clear what is going on here. Hopefully we can work towards a merge over time.

Currently this performs the following changes:
* Separates Bundle.js into Graph.js and Bundle.js
* Adds graph.link and graph.buildSingle
* Bundle is created only when graph.buildSingle is called
* Graph is exported from the main rollup build to allow deeper hooking in the wrapper project

Hopefully that makes it clearer where this is roughly going. Happy to discuss anytime.

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -151,124 +151,125 @@ export default class Bundle {
 		 .then(addons => addons.filter(Boolean).join(sep));
 	}
 
-	build () {
+	link () {
 		// Phase 1 – discovery. We load the entry module and find which
 		// modules it imports, and import those, until we have all
 		// of the entry module's dependencies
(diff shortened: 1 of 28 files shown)


guybedford:
Moved to https://github.com/rollup/rollup/pull/1811.

PR #1805, opened 2017-12-21
Refactors config validation, deprecation and unification logic.

ankeetmaini:
This was originally raised for #1762. This builds on top of #1769


ankeetmaini:
@lukastaegert I think I've solved it this time. Hahaha!

guybedford:
This work looks great. We've been working on a TypeScript conversion in https://github.com/rollup/rollup/pull/1806, which has seen some inconsistencies in supporting the highly dynamic options interfaces. I wonder if this would simplify that or make it harder...

ankeetmaini:
@guybedford To solve the inconsistencies among the CLI sand node I've created base input/output options in the file mergeOptions.js which is now the central place for all options related logic. I hope this will make it simpler!

Issue #1836, opened 2018-01-05
Add Support for TypeScript for API users


alan-agius4:
This will make it more convenient for typescript users (and suggested by typescript) to provide typings with the repository itself, rather than another repo (such as DefinitelyTyped) which at the moment types how rollup are located..

Since the rollup is written in TypeScript this will be more easier to do. However the blocker is that the TypeScript compilation of rollup is done using `rollup-plugin-typescript` which doesn't support emitting declaration files.

https://github.com/rollup/rollup-plugin-typescript/issues/62

Propose, use `tsc` to build `TypeScript` and than flatten using rollup or try to fix `rollup-plugin-typescript` to emit declarations.

PR #1837, opened 2018-01-05
feat: ship TypeScript declaration files

alan-agius4:
Closes: #1836

Some code changes needed to be done as some interface needs to be exported.

### DIFF
--- a/src/ast/CallOptions.ts
+++ b/src/ast/CallOptions.ts
@@ -4,10 +4,10 @@ import Expression from './nodes/Expression';
 import NewExpression from "./nodes/NewExpression";
 import Property from "./nodes/Property";
 
-interface callOptions {
-	withNew: boolean,
-	args?: Expression[],
-	caller: TaggedTemplateExpression | CallExpression | NewExpression | Property | void
+export interface CallCreateOptions {
+	withNew: boolean;
+	args?: Expression[];
+	caller: TaggedTemplateExpression | CallExpression | NewExpression | Property | void;
 }
 
 export default class CallOptions {
@@ -15,11 +15,11 @@ export default class CallOptions {
 	args: Expression[];
 	caller: TaggedTemplateExpression | CallExpression | NewExpression | Property | void;
 
-	static create (callOptions: callOptions) {
+	static create (callOptions: CallCreateOptions) {
 		return new this(callOptions);
 	}
 
-	constructor ({ withNew = false, args = [], caller = undefined } = {}) {
+	constructor ({ withNew = false, args = [], caller = undefined }: CallCreateOptions = {} as any) {
 		this.withNew = withNew;
 		this.args = args;
 		this.caller = caller;
--- a/src/ast/nodes/AssignmentPattern.ts
+++ b/src/ast/nodes/AssignmentPattern.ts
@@ -2,7 +2,7 @@ import Node from '../Node';
 import ExecutionPathOptions from '../ExecutionPathOptions';
 import Pattern from './Pattern';
 import Expression from './Expression';
-import Scope from '../Scopes/Scope';
+import Scope from '../scopes/Scope';
 import { UnknownAssignment } from '../values';
 import Declaration from './Declaration';
 import { ObjectPath } from '../variables/VariableReassignmentTracker';
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "dist/rollup.js",
   "module": "dist/rollup.es.js",
   "jsnext:main": "dist/rollup.es.js",
+  "typings": "dist/typings/src/node-entry.d.ts",
   "bin": {
     "rollup": "./bin/rollup"
   },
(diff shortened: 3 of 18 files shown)


lukastaegert:
Thanks, this looks like a useful addition! Note however that the Travis-CI failed dramatically. Since appveyor succeeded, I assume this is an OS specific issue we should figure out.

alan-agius4:
Ill try to have a look at it in the coming days.

alan-agius4:
@lukastaegert all green, issue was some import paths has wrong casing.

src/ast/CallOptions.ts
```diff
@@ -4,22 +4,22 @@ import Expression from './nodes/Expression';
 import NewExpression from "./nodes/NewExpression";
 import Property from "./nodes/Property";
 
-interface callOptions {
-	withNew: boolean,
-	args?: Expression[],
-	caller: TaggedTemplateExpression | CallExpression | NewExpression | Property | void
+export interface CallCreateOptions {
```
lukastaegert: I agree this is probably a better name 👍

src/ast/nodes/AssignmentPattern.ts
```diff
@@ -2,7 +2,7 @@ import Node from '../Node';
 import ExecutionPathOptions from '../ExecutionPathOptions';
 import Pattern from './Pattern';
 import Expression from './Expression';
-import Scope from '../Scopes/Scope';
+import Scope from '../scopes/Scope';
```
lukastaegert: 👍

alan-agius4:
Awesome cheers 😁

MattiasBuelens:
As a TypeScript developer: thank you! 😄 

Note that DefinitelyTyped has a manually authored type definition published as `@types/rollup`. When the next version of Rollup is released, this type package [should be removed](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package) in favor for Rollup's own types.

alan-agius4:
You are welcome :). I am aware of that, and I’ll tale care of removing it

guybedford:
This PR seems to be outputting a full individual file compile of Rollup into the dist folder, along with typings for each of the unbuilt files.

Surely we should just be publishing one top-level rollup.d.ts here for the outward API?

I'm not sure I follow the use case for having separate typings per development module too, will TypeScript really be able to match up the types from these individual files to the dist/rollup.js file?

guybedford:
It seems the typings run through "dist/typings/src/node-entry" as the main file in Node and these are working ok here.

The only issue is that it doesn't support the `import rollup from 'rollup'` default form and requires users to write `import * as rollup from 'rollup'` it would be nice if we could get this fixed too.

alan-agius4:
Well at the moment, `rollup` doesn't export a default, this has nothing to do with typings per see. Regarding the flatting of `.d.ts` files this is not yet possible with `TypeScript` yet.

MattiasBuelens:
Indeed, TypeScript doesn't have a way to output a single declaration file yet, see Microsoft/TypeScript#4433.

There are tools that try to do this (such as `dts-bundle` and `dts-generator`), but they come with their own limitations. Basically, these tools just concatenate the individual `.d.ts` files into one file, and fix some naming conflicts along the way. They don't "tree shake" the type definitions, so you still end up with one big file with lots of (internal) modules.

guybedford:
My other worry here is that for example the `InputOptions` in the API here are just a generic configuration object, so a lot of the type benefits aren't actually there.

It could be worth considering manually maintaining the outward .d.ts API file, especially since the Rollup API is so encapsulated, almost all of the types here are internal.

alan-agius4:
Interal types should be marked as such so that they are not outputted.

Personally I don’t agree to having the types not generated by typescript
when you are wrting your code in typescript, why should someone keep on
maintaning the definitions when they are created put of the box?

My recommandation would be to structure the project in a way to emit the
types that are really needed and mark internal stuff as indeed internals so
they are not emitted.

However, I think while most of the types are outputted, these still cannot
be consumed by the API users, these are just internal for typescript and
not for developers, if there is a type that you can access which one
shouldn’t please do let me know.

alan-agius4:
Re ‘InputOptions’ it doesn look like a configuration object and is correctly defined with the proper types. In general i shouldn’t be able to assign an invalid value to the InputOptions.

InputOptions is actual one of those critical types to have.

guybedford:
@alan-agius4 I'm referring to the generic options we take in https://github.com/rollup/rollup/blob/master/src/rollup/index.ts#L208.

I didn't know it was possible to mark types as "internal" in TypeScript... how would we go about doing this?

alan-agius4:
Basicilly for the internal usefully you use it when somethings needs to be
exported you still dont want it exposed.

More info about this can be found here —stripInternal.
https://www.typescriptlang.org/docs/handbook/compiler-options.html


Actually the strandard for that GenericObject that you highlighted is
usually called a Dictionary, and declared like such;
https://github.com/sketch7/ssv-core/blob/develop/src/collection.ts.

However, yeah that might be redundant unless its reused in multiple places.

Really don’t like the name of it now that I am looking at it!

alan-agius4:
I’ll have a look tomorrow probably, Maybe’ll be able to clean up some of
the undeed / unwanted declarations.

lukastaegert:
A dictionary seems to contain uniform information which is not the case here?

Anyway the reason we are using a generic object is because
* we do not want to litter the configuration object type with deprecated options
* we want to check for unexpected keys
* I was not considering others using this type 😜

I think we actuallyhandle this differently in the future:
* Use a type that only contains valid, non-deprecated options in our signature
* Cast this to a generic object when checking for deprecations and unexpected keys

Anyway, I think there is no harm in keeping this as it is for now (though merging might have been a little premature...) and do incremental improvements. But we should not remove the DefinitelyTyped definitions just yet.

PR #1841, opened 2018-01-05
Code Splitting

guybedford:
Finally got the first draft of this feature working (#372). This is feature complete, passing tests, and ready for review. It deals with the edge cases comprehensively.

This is written as a PR against the dynamic import inlining PR, so I'll update this as that gets reviewed in as well (lets get that one going first too).

The API for this is to take `input` as an array of entry points:

```js
  input: ['main1.js', 'main2.js', 'feature.js'],
  experimentalCodeSplitting: true,
  experimentalDynamicImport: true
});
```

Then the bundle returned can be written with a `dir` argument instead of a `file` argument:

```js
bundle.write({
  dir: 'dist',
  format: 'cjs'
});
```

Code splitting supports output formats `cjs`, `es` and `amd` but not iife or umd due to the inability to get good dependency support.

The above would output files: `dist/main1.js`, `dist/main2.js` and `dist/feature1.js` where those files would not contain any dupicated modules between them. Instead, where modules are duplicated, a separate chunk is created with a hash suffix and both modules will load from that. The algorithm to do this is basically a graph colouring algorithm where every entry point is given a unique colour (hash buffer), and we then iterate all reachable modules from that iterate, colouring in the graph. As we iterate the other entry points the colours mix, and at the end, the chunks are the modules that have the same colours.

Detected dynamic imports are treated as "discovered" entry points in this process, thereby sharing the chunking.

There is no registry used in any of this - we simply wire up the import and export bindings, carefully managing execution order consistency and then adapting the finalisers and deshadowing process to deal with the new deshadowing problems between these boundaries.

The wiring gets complex quickly on these cases, so I had to make some architectural changes to the deshadowing and finaliser process, basically:

* `getName ()` is now simply return this.safeName || this.name on all Variable classes.
* Bundle now handles working out the module boundaries, which it then passes to the finalisers through getModuleDeclarations(). This entirely abstracts away finalisers from having to think about things like AST Nodes, variables, externalModules etc, and just follow the bundle data returned (separation of concerns win hopefully, although sorry to hack up the finalisers to much!)
* The deshadowing function now runs on render, as the same variable can have a different deshadowing depending on which bundle it is in. A nice result of this is that unnecessary deshadowing test cases from normal builds could be simplified.
* The deshadowing function is now called `setRenderIdentifiers` and includes the previous getName logic, ie setting an identifier safeName to `exports.exportName` where necssary and everything. We could even consider changing `safeName` to `bundleName`, `outputName` or `renderName` to make this clearer.


guybedford:
I've finished padding out the tests now including full dynamic import chunking cases for CJS and AMD output being supported.

jthoms1:
We would be very interested in using this within StencilJS.  I have been working on a way to accomplish this using a custom tool to identify bundling use cases based on https://github.com/samthor/srcgraph. It appears that this PR will mitigate much of the issues I have had.

TheLarkInn:
This is killer work @guybedford. If you need any help with approaching use cases that we've already covered let me know if we can lend a hand.

Rich-Harris:
I won't pretend to have fully read and understood this, #1816 and #1840, but I've been giving it a little poke tonight and it really is something. Bloody amazing work!

Munter:
Is it possible to add a sort of manifest file to the output? This would let consumers implement preloading in browsers of parsed she converted to http headers or H2 push

guybedford:
@Munter the `chunks` property provides this information:

```js
bundle.chunks: {
  [name: string]: {
    imports: string[],
    exports: string[],
    modules: string[]
  }
}
```

Reading the `imports` through the chunks provides the chunk dependency graph for preloading.

src/rollup/index.ts
```diff
@@ -344,7 +361,149 @@ export default function rollup (rawInputOptions: GenericConfigObject) {
...
+									}
+
+									code += `//# ${SOURCEMAPPING_URL}=${url}\n`;
+								}
+
+								promises.push(writeFile(dir + '/' + chunkName, code));
+								return Promise.all(promises).then(() => {
+									return mapSequence(
+										graph.plugins.filter(plugin => plugin.onwrite), (plugin: Plugin) =>
+											Promise.resolve(plugin.onwrite(assign({ bundle: chunk }, outputOptions), chunk))
```
andrewiggins: Should `onwrite` also receive the `chunkName` as a parameter somewhere? Previously in "single bundle mode", `onwrite` could figure out what files were written using the `outputOptions`, but now with chunks, I think it needs the `chunkName` to be able to re-construct the written file paths.

guybedford: The chunk object above actually does include the `name` as well, so I believe that would be passed to the plugins ok?

src/rollup/index.ts
```diff
@@ -223,8 +227,21 @@ export default function rollup (rawInputOptions: GenericConfigObject) {
...
+				inputOptions.input = [inputOptions.input];
+		}
+		else if (inputOptions.input instanceof Array && !codeSplitting) {
+			error({
+				code: 'INVALID_OPTION',
+				message: 'Multiple inputs only supported when setting the experimentalCodeSplitting flag option.'
+			});
+		}
+
+		if (!codeSplitting) return graph.buildSingle(inputOptions.input)
```
jthoms1: When the codeSplitting flag is true then the TS types returned by generate will be different.  Currently TS believes that the same types are returned whether this is set or not.  I am not exactly sure the right approach to resolve this rather than setting an OR statement in the OutputBundle's generate result.

guybedford: There are basically two overloads of the generic rollup function from a typing perspective:

```js
interface SingleInputOptions extends InputOptions {
  input: string;
}
interface MultipleInputOptions extends InputOptions {
  input: string[];
}
function rollup(options: SingleInputOptions): Promise<BundleSingle>;
function rollup(options: MultipleInputOptions): Promise<BundleChunks>;
```

jthoms1: Cool this would work as well.

src/rollup/index.ts
andrewiggins: Ah okay, must've misread it. Sounds good!

jthoms1:
I am using this branch in a local version of StencilJS to build out Ionic.  In my testing so far I have not ran into any issues.  The chunking logic works exactly as expected.

test/chunking-form/samples/dynamic-import-treeshaking/_expected/es/dep2.js
```diff
@@ -0,0 +1,7 @@
+import { multiplier } from './chunk-36864708.js';
+
+function mult (num) {
+  return num + multiplier;
+}
+
+export { mult };
```
lukastaegert: So I assume the point of this test is to show that the unused imports from `dep2.js` are properly removed. But maybe we can do better by not emitting this chunk at all? But this might also be a possible future enhancement.

For instance the logic could be that a chunk is not emitted when
* There are no synchronous imports from it (otherwise we should probably include it for potential side-effects until we found away to improve the side-effect detection logic to find out if a chunk has side-effects in its initialization code. Actually, this could probably be done by checking if any of the top-level statements are included or if one of the imports has side-effects. But again, maybe a future enhancement)
* None of the dynamic imports are included

Another interesting fact I stumbled upon that we might improve in the future is that when
* a chunk exports a variable and
* this chunk is imported by two other chunks both of which import this variable and
* only one of the chunks actually uses this variable,
the imported variable is retained for both chunks. The reason is that there is only one variable entity and as soon as it is included, all imports of this variable are included. If none of the chunks use the variable, it is properly removed. I think it is more of a cosmetic issue but still.

test/chunking-form/samples/entrypoint-fascade/_config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+	description: 'entry point fascade creation',
+	options: {
+		input: ['main1.js', 'main2.js']
+	}
+};
```
lukastaegert: I guess indeed the proper term should be facade without the "s" in both the description and folder name.

test/chunking-form/samples/dynamic-import-treeshaking/_expected/es/dep2.js
guybedford: You've exactly understood the two limitations of the approach I put together - the chunking algorithm is run on the graph _before_ treeshaking, and the export boundaries between chunks aren't properly pruned. I originally aimed to handle both of these cases (and my first few attempts were in this direction), but hit some road blocks.

When running chunking on the graph after treeshaking, execution order still has to be maintained as an invariant. In addition we need to work out which chunks can be considered pure in order to skip emission like you mention.

If I remember correctly, the main issue I hit when implementing this (after fixing up export tracing etc etc) was that the concept of "marked" or "used" must then be considered uniquely for each specific chunk, which wasn't a simple capability to add. I can't remember what the other issues were, but if you're interested I could try trawling through my notes for future reference.

guybedford: Note that in may of my practical investigations of the algorithm, these assumptions ended up not seeming so bad at all though. The majority of cases work really well without these features. I think as we use this feature we'll get a better idea for how bad "pure" chunk relations and unpruned boundaries might end up looking, and can either consider specific fixes, or taking the more general route then.

guybedford:
Thanks for the initial review. I've made the changes and fixed up the namespace test.

To explain the entrypoint facade concept the point is that when an entry point module is contained in a chunk, where that chunk has other exports as well, those other exports would effectively pollute the entry point module exports if we treated that chunk as the entry point itself.

So the way we handle this is to first imagine all entry points as being "fascades" to their actual exports in whatever chunk their actual code happens to be in. Those export names might then be changed and bundled with other exports, but the facade module allows us to just reexport the right interface.

In the case where a chunk contains an entry point, and no name changes are necessary and no other exports are provided from the chunk we then make the optimization that the entrypoint facade is no longer necessary as the chunk matches the facade. So then we can just treat the chunk as the facade (as happens in most of the tests).

Let me know if there is anything that is still unclear on that.

thelgevold:
I gave the code in this branch a try with a mid size Angular application.
It seems to work really well.

In my setup I used Rollup code splitting to generate lazy modules. Next I implemented lazy loading in Angular using the router, aided by systemjs url mappings.

Here is a summary of my findings: http://www.syntaxsuccess.com/viewarticle/lazy-loading-with-rollup.

Here is the config: https://github.com/thelgevold/angular-2-samples/blob/master/rollup/rollup-lazy-config.js

lukastaegert:
Nice!

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+export interface ReexportSpecifier {
+	reexported: string;
+	imported: string;
+};
+export interface ImportSpecifier {
+	local: string;
+	imported: string;
+};
+
 export default class Bundle {
```
lukastaegert: I am more and more getting the feeling that `Chunk` might be a better name. Or otherwise, we could start referring to chunks as bundles. At least to me, it seems these are the same concepts. If so, they should not have different names in the code depending on context.

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
 		this.externalModules = undefined;
+
+		this.dependencies = undefined;
 		this.entryModule = undefined;
+		this.entryModuleFacade = false;
+		orderedModules.forEach(module => {
+			if (module.isEntryPoint) {
+				if (!this.entryModule) {
+					this.entryModule = module;
+					this.entryModuleFacade = true;
```
lukastaegert: I would prefer `isEntryModuleFacade` here as to my understanding, this flag determines if the bundle itself is posing as the facade or if an additional facade needs to be created. Alternatively, the inverse flag might make even more sense to the casual reader: `needsEntryModuleFacade`, as now this flag signals that additional care needs to be taken?

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+
+		this.exportedVariables = new Map();
+		this.imports = [];
+		this.exports = {};
 		this.externalModules = undefined;
+
+		this.dependencies = undefined;
 		this.entryModule = undefined;
+		this.entryModuleFacade = false;
+		orderedModules.forEach(module => {
```
lukastaegert: Could make sense to put this into a separate method, e.g. `analyseModules`? It is quite possible more code will be added here in the future.

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+		let i = 0;
+		if (variable.exportName) {
+			safeExportName = variable.exportName;
 		} else {
-			// Necessary to properly replace namespace imports
-			this.orderedModules.forEach(module => module.includeAllInBundle());
+			safeExportName = variable.exportName = variable.name;
+		}
+
+		let curExport = this.exports[safeExportName];
```
lukastaegert: `curExport` has a double meaning: First it is used to iterate through existing exports, then it is used to contain the new export. I would suggest the following change:

```ts
while (this.exports[safeExportName]) safeExportName = (variable.exportName || variable.name) + '$' + ++i

const newExport = this.exports[safeExportName] = { module, name: undefined, variable }
```
and then use `newExport` instead of `curExport` from then on.
Also, this produces names like `myShadowedExport$1$2$3` (if I am not mistaken). Would be cooler (though just cosmetically) to have names like `myShadowedExport$3` instead. As conflicting exports are probably not very common, I do not think this is too performance critical as long as the unshadowed version is fast.

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+			this.externalModules = [];
+			this.dependencies = [entryFacade.bundle];
+			return;
+		}
+
+		this.externalModules = [];
+		this.dependencies = [];
+
+		this.orderedModules.forEach(module => {
+			module.dependencies.forEach(dep => {
```
lukastaegert: Why not `dependency`? Would be the proper singular and reads a little nicer further down 😉

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+			}
+			this.exportedVariables.set(variable, exportName);
+		});
 	}
 
-	processExternals () {
-		// prune unused external imports
-		this.externalModules = this.graph.externalModules.filter(module => {
-			return module.used || !this.graph.isPureExternalModule(module.id);
+	generateDependencies (entryFacade?: Module) {
```
lukastaegert: Isn't it more like `collectDependencies`?

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
+					this.dependencies.push(depModule);
+					if (dep.isExternal) {
+						this.externalModules.push(dep);
+					}
+				}
+			});
+		});
+	}
+
+	generateImports () {
```
lukastaegert: This is a really long homogeneous wall of code that I find really hard to read. Maybe we could at least extract the namespace handling to a separate method, that would cut the size in half. Not sure there are other things that could be extracted.

src/Bundle.ts
```diff
@@ -104,54 +385,115 @@ export default class Bundle {
...
+					right: '], resolve, reject) })'
+				}
+			}
 		}
 
-		return resolvedId;
-	}
+		if (this.graph.dynamicImport) {
+			this.orderedModules.forEach(module => {
+				module.dynamicImportResolutions.forEach((replacement, index) => {
```
lukastaegert: Maybe it is possible to extract this into a separate method?

src/Bundle.ts
```diff
@@ -104,54 +385,115 @@ export default class Bundle {
...
+						node.setResolution(replacement, dynamicImportMechanism);
+					}
+				});
+			});
+		}
 
 		// ensure no conflicts with globals
 		keys(this.graph.scope.variables).forEach(name => (used[name] = 1));
 
 		function getSafeName (name: string): string {
```
lukastaegert: If we supply `used`, this function could be extracted out of this class. Also, this function might then be reused to deconflict other things like bundle file names.

src/Bundle.ts
```diff
@@ -104,54 +385,115 @@ export default class Bundle {
...
-				const safeName = getSafeName(name);
+			if (!es || module.exportsNamespace) {
+				const safeName = getSafeName(module.name);
 				toDeshadow.add(safeName);
-				(<ExternalVariable>declaration).setSafeName(safeName);
+				module.name = safeName;
+			}
+		});
+
+		this.imports.forEach(impt => {
```
lukastaegert: Maybe this could be extracted as well, just like the `orderedModules` loop below? My hope is that then, this really long function could read more like a script of high level intents where it is easy for people not familiar with the code to quickly identify where they need to look to change certain things.

src/Bundle.ts
```diff
@@ -162,46 +504,72 @@ export default class Bundle {
...
-		this.orderedModules.forEach(module => {
-			if (this.graph.dynamicImport) {
-				module.dynamicImportResolutions.forEach((replacement, index) => {
-					const node = module.dynamicImports[index];
+		const dependencies: BundleDependencies = [];
 
-					if (!replacement)
-						return;
+		let imports: ImportSpecifier[];
+		this.dependencies.forEach(dep => {
```
lukastaegert: What are you doing with each dependency here? Might be another candidate for extraction (with a good name that answers this question 😉)

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
 
+export type BundleDependencies = {
+	id: string;
+	name: string;
+	isBundle: boolean;
+	reexports?: ReexportSpecifier[];
+	imports?: ImportSpecifier[];
+}[];
+
+export type BundleExports = {
```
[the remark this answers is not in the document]
guybedford: This is because the shape is an array shape, which can't be represented by an interface (interfaces are objects types only).

src/Bundle.ts
guybedford: Ok I've changed the terminology to:
* Chunk as the name of each group of rendered modules
* Bundle as the name of the group of Chunks for a build

src/Bundle.ts
```diff
@@ -12,68 +12,349 @@ import transformBundle from './utils/transformBundle';
...
 	graph: Graph;
 	orderedModules: Module[];
+
+	// this represents the bundle module wrappings
+	// which form the output dependency graph
+
+	// map from variable exported by this bundle to its safe exported name
+	exportedVariables: Map<Variable, string>;
+	imports: {
+		module: Bundle | ExternalModule;
```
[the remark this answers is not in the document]
guybedford: As a dynamic language, there is no harm in having an array representing multiple distinct objects in JavaScript. It is typically function arguments having more than two variations of call shape that is the worry from a deoptimization perspective.

src/Bundle.ts
guybedford: Sure. This shouldn't produce names like that as both the `exportName` and `name` are the original names due to the rendering changes.

src/Bundle.ts
guybedford: Sure, I was struggling to name these functions actually.

src/Bundle.ts
guybedford: `dep` is just a shortname. I usually prefer shorter iterates for readability.

src/Bundle.ts
guybedford: That sounds like a very sensible refactoring to do, I was aiming to just leave the utility functions exactly as-is for this PR. Further PR work can definitely dive into some of these refactorings.

src/Bundle.ts
guybedford: This is effectively a tracing function to populate the final import values. The key thing to follow with a function like this is that it is populating `this.imports` on the chunk object from the module.imports and chunk information. Understand the input and output data structures and the function will make sense.

I've split out a duplicated sub-function here for better readability.

src/Bundle.ts
guybedford: Sure

src/Bundle.ts
guybedford: Sometimes inline code is easier to read as high level intents only make sense when you understand the high level picture. Maintaining low-level readability is important too.

src/Bundle.ts
guybedford: This function is consolidating the "module declaration" interface for the entire chunk. So it is gathering the export list, and the dependency list in a form that provides all the information about the cross-chunk boundaries. The dependencies loop here is exactly gathering the dependencies part of that interface. Again focus on the output data structures first and foremost over the words of it. Much more descriptive level to look at things.

bin/src/run/watch.ts
```diff
@@ -125,16 +125,22 @@ export default function watch (configFile: string, configs: RollupWatchOptions[]
...
 		// removing a non-existent listener is a no-op
 		process.stdin.removeListener('end', close);
 
 		screen.close();
-		watcher.close();
+		if (watcher) watcher.close();
 
 		if (configWatcher) configWatcher.close();
+
+		process.exit(1);
```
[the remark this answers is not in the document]
guybedford: Good catch.

src/Bundle.ts
guybedford: The analysis is all in the separate "collect" / "generate" functions. This is only for the entryModule and entryModuleFacade properties.

src/Bundle.ts
guybedford: 👍

guybedford:
Getting there :) there may well be ways to further refactor some of the interfaces here, will be good to discuss further. I'd suggest trying to load up the code in an IDE that provides the typing, as seeing the full return value interfaces makes it much easier to follow what is going on. The chunk rendering data structures are quite detailed pieces but cover the cases needed. Will also be interesting to see if they solved some of the re-export bugs in the process as well when this lands.

src/Bundle.ts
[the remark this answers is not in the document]
lukastaegert: Ah I see. I guess I missed the array brackets :). Still, it could be
```ts
interface ChunkDependency = {
	id: string;
	name: string;
	isChunk: boolean;
	reexports?: ReexportSpecifier[];
	imports?: ImportSpecifier[];
}
export type ChunkDependencies = ChunkDependency[]
```
or maybe not assign a special type to the array at all.

src/Bundle.ts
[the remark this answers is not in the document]
lukastaegert: Having read a little further I now understand this a little better. It's just the discrepancy in `module: Bundle...` that keeps tripping me of–but of course, bundles (chunks) are basically modules.

src/Bundle.ts
lukastaegert: Ah you are right, misread this.

src/Bundle.ts
lukastaegert: Very good, much better than my suggestion 👍

src/Bundle.ts
lukastaegert: 👍

src/Bundle.ts
lukastaegert: I agree that there is a line where extraction harms readability more than it helps. Nevertheless
for instance in this case, just extracting two methods, e.g. `deconflictImportedVariables` and `deconflictDeclaredVariables` (or similar) would tell me at a glance where I need to look if one of these two concerns needs to be updated. Instead of heaving to start to read each block at least a little to be sure I am  not mistaken about what the block does. It also prevents others from mixing in other concerns carelessly.
From that extraction, I will also learn that it is probably safe to switch these if I am not happy with the deconflicting order (remember there is a ticket to rework just that).
Furthermore if we extract `getSafeName` (does not need to be into a utility file for now, above the class could be enough), the only dependency of these to the function scope are `used` and `es` so these extracted functions are really rather encapsulated and can easily be reasoned about separately.

> high level intents only make sense when you understand the high level picture

I would actually argue the other way around: Telling the reader high level intents by extracting low level functions actually gives them the chance to understand the high level picture as they are not mentally burdened with the need to understand the details of each step at once.

src/Bundle.ts
lukastaegert: But you have to read a lot of code you are probably not interested in but you cannot be sure you are not interested in to get there. Basically having read this function again what it does is
```javascript
const reexportDeclarations = this.getReexportDeclarations()
const chunkDependencies = getChunkDependencies(reexportDeclarations)
const exports = getChunkExports()
return { dependencies, exports }
```

I know, this might look too simple to the naked eye but it is basically splitting this function into three totally separate, about equally sized concerns with little interdependency. Really easy to find out where I have to look if I want to fine-tune any of these things.

But please keep in mind, these are just my personal suggestions. I would be perfectly happy with keeping the code as it is. I just feel that these kinds of transformations will really help us keep agile as it is really easy to drill down into this kind of code for
* people unfamiliar with it
* people familiar with it that have not looked at it for half a year and probably for forgotten the details

lukastaegert:
I might have added some comments to outdated code. If that is a problem, I could just add them to the proper places of the updated code.

lukastaegert:
Damn, turned out a little later today than expected so I will again continue tomorrow (evening probably). On Friday, I can spend most of my time on rollup so I expect to finish the review then. Thanks for your patience and taking the time to refactor the code, looks really nice!

jthoms1:
Will there need to be updates to existing plugins to allow for this branch to be merged?  I know that during testing some errors occurred around multiple entry points with `rollup-plugin-commonjs` and `rollup-plugin-node-resolve`.

lukastaegert:
I am aware of an issue with `rollup-plugin-commonjs` when `experimentalDynamicImport` is enabled and used. To fix this, changes in this plugin (+ possibly rollup, depending on the solution) will be required. See https://github.com/rollup/rollup/pull/1816#issuecomment-356930606 and following comments for a discussion of the problem.

To my knowledge, code splitting is able to interact properly with dynamic imports but does not depend on them. Thus, there should be no issue with the CJS plugin when this is released. However these issue need to be fixed before we can make dynamic imports an always enabled feature.

As for `rollup-plugin-node-resolve`, I am not aware of issues. If there are it would be important to know so that we can have a look.

src/Module.ts
```diff
@@ -564,16 +594,16 @@ export default class Module {
 		return this.declarations['*'];
 	}
 
-	render (es: boolean, legacy: boolean, freeze: boolean): MagicString {
+	render (legacy: boolean, freeze: boolean): MagicString {
```
lukastaegert: Glad we finally got rid of the hardly used `es` flag!

src/ast/nodes/VariableDeclarator.ts
```diff
@@ -21,21 +19,4 @@ export default class VariableDeclarator extends NodeBase {
...
-				if (this.init) {
-					code.overwrite(this.start, this.id.end, variable.getName(es));
-				} else if (this.module.graph.treeshake) {
-					code.remove(this.start, this.end);
-				}
-			}
-		});
-
-		super.render(code, es);
-	}
```
lukastaegert: 👍

src/ast/scopes/Scope.ts
```diff
@@ -81,35 +81,37 @@ export default class Scope {
...
-			let i = 1;
+			if (!names.has(name)) {
+				return;
+			}
 
-			while (names.has(deshadowed)) {
+			name = declaration.name;
+			let deshadowed, i = 1;
+			do {
 				deshadowed = `${name}$$${i++}`;
```
lukastaegert: I think this could actually be changed to `${name}$${i++}` to avoid the extra "$". Might cause too much test noise for this PR, though, and might be better handled when/if we should decide to unify the deconflicting helpers. Just noticed.

lukastaegert:
This was the main part of the review, everything that is missing are the finalisers + some small files here and there. I'm confident that I will conclude this tomorrow and looking forward to releasing this! Great stuff! Note that we definitely need to find a solution for the browser bundle first, though, see my comments.

src/Module.ts
```diff
@@ -32,16 +32,22 @@ import ImportDefaultSpecifier from './ast/nodes/ImportDefaultSpecifier';
...
-import ExternalVariable from './ast/variables/ExternalVariable';
 import { isTemplateLiteral } from './ast/nodes/TemplateLiteral';
 import { isLiteral } from './ast/nodes/Literal';
 import { missingExport } from './utils/defaults';
+import Chunk from './Chunk';
 
 wrapDynamicImportPlugin(acorn);
 
-export interface IdMap {[key: string]: string;}
+export interface VariableTrace {
```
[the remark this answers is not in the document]
guybedford: :+1:

src/Module.ts
```diff
@@ -123,24 +128,31 @@ export default class Module {
...
 	magicString: MagicString;
 	originalCode: string;
-	originalSourcemap: RawSourceMap;
+	originalSourcemap: RawSourceMap | void;
 	reexports: { [name: string]: ReexportDescription };
-	resolvedExternalIds: IdMap;
 	resolvedIds: IdMap;
 	scope: ModuleScope;
 	sourcemapChain: RawSourceMap[];
+	strongDependencies: Module[];
```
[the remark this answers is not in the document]
guybedford: This code was the circular reference hinting code in Graph.ts, which has a comment on it noting that it doesn't work. I wasn't quite sure what the exact behaviour of this was supposed to be either. Due to the changes here, I just made that code to follow the typing ok, but it still doesn't work. We could potentially just remove it entirely for now, and note it in an issue, but the code may well act as a good placeholder too.

src/Module.ts
```diff
@@ -175,6 +186,8 @@ export default class Module {
 			this.dynamicImports = [];
 			this.dynamicImportResolutions = [];
 		}
+		this.isEntryPoint = false;
+		this.execIndex = undefined;
```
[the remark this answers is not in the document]
guybedford: It's usually a good practice in JavaScript to set all object properties upfront to avoid reshaping - in this scenario we are creating a a lot of Module objects, before then later running a batch operation on all which will set this property, which would otherwise result in a bulk instance shape change at that point. Reshaping of a class will usually converge over time so you can often get away without it, but because of the upfront initialization of all Module objects first, the reshaping here can benefit from it (the best example I could find https://jsperf.com/initializing-properties-to-undefined, although it's probably not necessarily indicative)

src/Module.ts
```diff
@@ -197,7 +210,6 @@ export default class Module {
 		this.sources = [];
 		this.dependencies = [];
 		this.resolvedIds = resolvedIds || blank();
-		this.resolvedExternalIds = resolvedExternalIds || blank();
```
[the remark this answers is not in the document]
guybedford: Glad you agree :)

src/Module.ts
```diff
@@ -512,6 +519,29 @@ export default class Module {
 		error(props);
 	}
 
+	getAllExports () {
+		const allExports = assign(blank(), this.exports);
+
+		keys(this.reexports).forEach(name => {
+			allExports[name] = true;
+		});
```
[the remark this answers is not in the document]
guybedford: Makes sense. Yes would be good to bulk replace the object shorthands, was wondering about that. Maybe we can create an issue to track.

src/ast/scopes/Scope.ts
guybedford: This is how it's always been it seems, so may result in large ecosystem churn to consider separately. The rule seems to work out as $$ for locals and $ for module specifiers with this method.

guybedford: (as in I agree!)

src/Graph.ts
```diff
@@ -24,7 +24,11 @@ import {
 import { RawSourceMap } from 'source-map';
 import Program from './ast/nodes/Program';
 import { Node } from './ast/nodes/shared/Node';
-import Bundle from './Bundle';
+import Chunk from './Chunk';
+import path from 'path';
+import GlobalScope from './ast/scopes/GlobalScope';
+import xor from 'buffer-xor';
+import * as crypto from 'crypto';
```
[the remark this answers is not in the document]
guybedford: Sure will see what I can do.

src/Graph.ts
```diff
@@ -227,99 +267,254 @@ export default class Graph {
...
+				chunk.generateImports();
+				chunk.generateEntryExports(entryModule);
+
+				timeEnd('phase 4');
+
+				return chunk;
+			});
+	}
+
+	buildChunks (entryModuleIds: string[]): Promise<{ [name: string]: Chunk }> {
```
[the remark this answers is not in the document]
guybedford: At it's made an improvement on the `buildSingle` function though from where it was before as the main build method.

I'm still weary of over-abstracting though, it can be just as much or even more of a maintenance issue having to jump around a file to work out exactly what a method is doing. Seeing the low-level details in one place is a powerful way to understand something even if it does mean you have to stare at it for longer than 5 minutes :P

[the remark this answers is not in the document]
guybedford: Also the idea is the code comments provide the high-levels... reading through the comments gives an algorithmic overview, better than some clever method names might.

src/Graph.ts
```diff
@@ -227,99 +267,254 @@ export default class Graph {
...
-			const seenEntry = seen[module.id];
-			if (seenEntry) {
-				if (seenEntry === curEntry)
-					hasCycles = true;
+		const visit = (module: Module, seen: { [id: string]: boolean } = {}) => {
+			if (seen[module.id]) {
+				// one more xor for circular entry points to ensure distinction
+				// to ensure one entry point per chunk property
+				if (module.isEntryPoint)
+					module.entryPointsHash = xor(module.entryPointsHash, curEntryHash);
```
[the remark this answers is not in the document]
guybedford: Excellent catch - yes we must avoid xoring with itself!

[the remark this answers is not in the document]
guybedford: Ok changed the algorithm slightly :)

src/Graph.ts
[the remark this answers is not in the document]
guybedford: Ok I've replaced it with native functions here and gone with numbered chunks. Hashing could be useful, but it should at least be proper hashing of the source that can cover all file names in the build as a separate feature.

src/Module.ts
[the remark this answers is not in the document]
lukastaegert: I am all in favour of removing the code.

I think—if the issue has not been solved already e.g. by your algorithm to fix the execution order—we are much better off starting fresh from the current code-base. The code will still be available via git. I think, these kind of reminders in the form of dead code in the code base are a bad pattern that is misleading and distracts from the important code. Issues and their discussions are a much better kind of reminders 😀

src/Module.ts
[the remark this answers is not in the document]
lukastaegert: Agreed. Though I would actually assign `null` instead of `undefined` to make this more explicit, the same way it is done e.g. in the ESTree spec. That also reduces the risk of this being accidentally removed as a seeming no-op (which of course it isn't)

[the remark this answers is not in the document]
guybedford: I usually try to avoid null for the `typeof null === 'object'` catch but sure.

src/rollup/index.ts
```diff
@@ -345,7 +362,149 @@ export default function rollup (rawInputOptions: GenericConfigObject) {
 
 				return result;
 			});
+
+		return graph.buildChunks(inputOptions.input)
```
lukastaegert: My hope is that if we can manage to make code-splitting always on (as outlined above), we could basically remove the first version of `generate`.

src/rollup/index.ts
```diff
@@ -68,6 +68,7 @@ export interface InputOptions {
 	legacy?: boolean;
 	watch?: WatcherOptions;
 	experimentalDynamicImport?: boolean;
+	experimentalCodeSplitting?: boolean;
```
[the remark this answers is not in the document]
guybedford: The difference here is that dynamic imports are inlined into one file for the single-file output, while dynamic imports can generate chunks for the directory output.

So I think it's perhaps clearer to consider `input[], dir` and `input, file` as completely distinct cases. We could consider allowing `input` as a string for the `dir` case possibly.

src/rollup/index.ts
guybedford: Perhaps a middle ground for always-on could be that one input and an output file would output chunks in the same folder as the output file, possibly separating the dynamic import inlining option as `inlineDynamicImports: true` or similar?

lukastaegert: Ah I see, I was forgetting about this. I guess a flag would definitely make sense as otherwise the behaviour would be inconsistent.

lukastaegert: So maybe:
* If we have a single entry point
  * if `output.file` is provided, it is the name of the chunk generated for the entry point
  * if a dynamic import is encountered + `inlineDynamicImports: true`, it is of course inlined
  * if a dynamic import is encountered + `inlineDynamicImports: false`
    * if `output.dir` is provided, additional generated chunks are placed there
    * otherwise, an error is shown
* If we have multiple entry points
  * if `output.dir` is missing, we show an error
  * if `output.file` is provided, we show a warning that it is ignored
?

test/chunking-form/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+								try {
+									expectedCode = normaliseOutput(
+										sander.readFileSync(samples, dir, '_expected', format, file)
+									);
+								} catch (err) {
+									expectedCode = 'missing file';
+								}
+
+								/*try {
```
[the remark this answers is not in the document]
guybedford: Got this going :)

src/Graph.ts
```diff
@@ -27,23 +27,24 @@ import Chunk from './Chunk';
...
+	let uniqueName = name;
 	let uniqueIndex = startAtTwo ? 2 : 1;
-	while (curEntryChunkNames.indexOf(uniqueEntryName) !== -1)
-		uniqueEntryName = entryName + uniqueIndex++;
-	curEntryChunkNames.push(uniqueEntryName);
-	return uniqueEntryName + ext;
+	while (chunkNames[uniqueName])
+		uniqueName = name + uniqueIndex++;
+	chunkNames[uniqueName] = true;
+	return uniqueName + ext;
```
lukastaegert: Very good, also like the stream-lined names!

src/rollup/index.ts
guybedford: I noticed the `experimentalCodeSplitting` flag isn't really behaving like an "experimental" flag here since it's guiding the whole process. I've made the following change for now:

* If `input` is a string, do a single file build, assuming `file` and warning otherwise
* If `input` is an array and the flag is enabled, do a code splitting build, assuming `dir` and warning otherwise

We can potentially look at different arrangements of this before flag removal, but at least the flag itself is not part of the final decision now.

lukastaegert:
I will put this into a release branch first which I will merge into master once we are sure everything is there that should be there :)

nnmrts:
This was a great story to read.

PR #1858, opened 2018-01-09
chore: clean up internal types and install missing `@types`

alan-agius4:
This is one of the first PRs to improve the TypeScript structure and eventually try to reduce the amount of outputted `.d.ts` files


src/Graph.ts
```diff
@@ -569,9 +569,10 @@ export default class Graph {
 				? Promise.resolve(resolvedId)
 				: this.resolveId(source, module.id)
 			).then(resolvedId => {
-				const externalId =
-					resolvedId ||
-					(isRelative(source) ? resolve(module.id, '..', source) : source);
+				// todo: there seems to be some issue here!
```
alan-agius4: @lukastaegert I might found some issue here, though I won't like to address it in the PR.

alan-agius4:
Actually in most cases you don’t need to add the internal keyword unless a
class or method is exposed and shouldn’t. As stripInternal will still
create some files but blank. I think overall we need to improve some
exports etc...

alan-agius4:
Basicilly the the latest changes these are exported
image
image
image

If you see anything that shouldn't be exported let me know.

PR #1863, opened 2018-01-11
import collapse

kellyselden:


src/utils/mergeOptions.ts
```diff
@@ -1,7 +1,6 @@
 import ensureArray from './ensureArray.js';
-import deprecateOptions from './deprecateOptions.js';
```
TrySound: How's js even work here?

KingHenne: I think `rollup-plugin-node-resolve` is the reason.

KingHenne: See also [this snarky comment](https://github.com/rollup/rollup/blob/master/rollup.config.js#L38). 😉

PR #1868, opened 2018-01-12
fix: file 'node_modules/rollup/typings/package.json.d.ts' not found

alan-agius4:


alan-agius4:
Closed in favor of 1871

PR #1872, opened 2018-01-12
docs(README): add `TypeScript` API section

alan-agius4:


Rich-Harris:
Thanks @alan-agius4. I think maybe this would fit better in a section on https://github.com/rollup/rollupjs.org, rather than on the README directly (which is designed to be a quick overview of the project — using the API programmatically is an advanced topic by itself, with TypeScript doubly so). What do you reckon?

alan-agius4:
I totally agree, though I don’t know where can I find the sourcecode of
that :)

alan-agius4:
Ill try to do a PR there :)

Rich-Harris:
cool! This is probably the right place for it — https://github.com/rollup/rollupjs.org/blob/master/guide/en/04-javascript-api.md

alan-agius4:
PR done https://github.com/rollup/rollupjs.org/pull/108 :)

PR #1887, opened 2018-01-18
RFC: add Module.includeAllInBundleRecursive

kellyselden:
Uses #1845 as a base to create meaningful tests. That description also perfectly applies here https://github.com/rollup/rollup/pull/1845#issue-286521730.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -4,7 +4,7 @@ import { blank, keys } from './utils/object';
 import Module, { IdMap, ModuleJSON } from './Module';
 import ExternalModule from './ExternalModule';
 import ensureArray from './utils/ensureArray';
-import { load, makeOnwarn, resolveId } from './utils/defaults';
+import { load, makeOnwarn, resolveId, missingExport } from './utils/defaults';
 import { mapSequence } from './utils/promise';
 import transform from './utils/transform';
 import relativeId from './utils/relativeId';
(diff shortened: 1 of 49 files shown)


guybedford:
I'm not sure I follow the use case here, is there an issue tracking the feature or further description somewhere?

kellyselden:
Expanding upon https://github.com/rollup/rollup/pull/1845#issue-286521730, I essentially want to add the module with the missing export to the entry modules. This PR has to be rewritten for the new system, but the use case still remains. I wish this use case didn't have to exist, but it is a interop concern that will hopefully correct itself once we roll this out to Ember users with a warning. Here are a couple of examples I found with my testing:
https://github.com/DockYard/ember-composable-helpers/pull/287
https://github.com/jmurphyau/ember-truth-helpers/pull/78
https://github.com/samselikoff/ember-cli-mirage/blob/v0.4.1/app/initializers/ember-cli-mirage.js#L4 `testConfig` doesn't exist here https://github.com/samselikoff/ember-cli-mirage/blob/v0.4.1/blueprints/ember-cli-mirage/files/__root__/config.js

guybedford:
Does the `interop` option not make the default export grabbing work through Rollup?

So if I'm hearing you right, this change is necessary to ensure the missing export hook triggers correctly? 

I've very concerned that the missing export hook merged exposes the internal `Module` API. This isn't a public API we should be exposing, and I think we should change that before it gets relied on.

//cc @lukastaegert

guybedford:
Also it would help to hear what the exact fix is that you are implementing through the missing export hook.

kellyselden:
The pseudo code is

```
missingExport(entryPoints, module) {
  // print deprecation
  entryPoints.push(module);
}
```

As you can see, I don't need any Module API after the chunking change, but I still need a way to continue traversing a broken tree.

guybedford:
Is this PR still valid?

PR #1889, opened 2018-01-18
use consistent strings in tests

kellyselden:
This is an issue in #1878, where I use the existing import collapsing code from finalisers/es.js, which rewrites code using single tick ' strings.


kellyselden:
This is not needed after all.

PR #1907, opened 2018-01-23
Fix CLI and dynamic import workflow bugs

guybedford:
In the process of putting together some example material these came up that were missed with some of the changes.

### DIFF
--- a/bin/src/run/build.ts
+++ b/bin/src/run/build.ts
@@ -10,7 +10,7 @@ import { BatchWarnings } from './batchWarnings';
 import { SourceMap } from 'magic-string';
 
 export default function build (inputOptions: InputOptions, outputOptions: OutputOptions[], warnings: BatchWarnings, silent = false) {
-	const useStdout = outputOptions.length === 1 && !outputOptions[0].file && inputOptions.input instanceof Array === false;
+	const useStdout = outputOptions.length === 1 && !outputOptions[0].file && !outputOptions[0].dir;
 
 	const start = Date.now();
 	const files = useStdout
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -59,6 +59,13 @@ export interface ImportSpecifier {
 	imported: string;
 };
 
+export interface DynamicImportMechanism {
+	left: string;
+	right: string;
+	interopLeft?: string;
+	interopRight?: string;
+};
+
 export default class Chunk {
 	id: string;
 	name: string;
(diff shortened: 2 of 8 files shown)


guybedford:
The example workflow I found these on was for https://github.com/rollup/rollupjs.org/pull/110.

test/chunking-form/samples/dynamic-import-chunking/_expected/cjs/main.js
```diff
@@ -11,7 +11,7 @@ function fn (num) {
 }
 
 function dynamic (num) {
-  return Promise.resolve(require("./dep2.js"))
+  return Promise.resolve({ default: require("./dep2.js") })
   .then(dep2 => {
     return dep2.mult(num);
   });
```
lukastaegert: As you can see here with this change, the code will be broken (you would need to write `dep2.default.mult(num)` to make it work).

src/Chunk.ts
```diff
@@ -390,8 +390,8 @@ export default class Chunk {
 		if (!es) {
 			if (format === 'cjs') {
 				dynamicImportMechanism = {
-					left: 'Promise.resolve(require(',
-					right: '))'
+					left: 'Promise.resolve({ default: require(',
+					right: ') })'
```
lukastaegert: What is the scenario you hope to fix here? Test? As you can see with the test below, this makes named exports no longer work (and again shows that we probably need functional tests). My feeling is that the correct approach would be more complicated.

guybedford: Great catch. Ok so the rule is this:

* When the dynamic import is an external module, we need to treat it as `{ default: module }` as that is how interop works when importing CommonJS from ES modules.
* When the dynamic import is a known module from our current bundle operation, we treat it as a full module object, as it's effectively within our private interface for bundling.

I've added the adjustments to make these cases work out now.

test/chunking-form/samples/dynamic-import-chunking/_expected/cjs/main.js
guybedford: This is no longer happening now.

bin/src/run/build.ts
```diff
@@ -10,7 +10,7 @@ import { BatchWarnings } from './batchWarnings';
 import { SourceMap } from 'magic-string';
 
 export default function build (inputOptions: InputOptions, outputOptions: OutputOptions[], warnings: BatchWarnings, silent = false) {
-	const useStdout = outputOptions.length === 1 && !outputOptions[0].file && inputOptions.input instanceof Array === false;
+	const useStdout = outputOptions.length === 1 && !outputOptions[0].file && !outputOptions[0].dir;
```
lukastaegert: Even though I just merged this, there is one minor thing I just noticed: If inputOptions is an array of length one and you specify neither `output.dir` nor `output.file`, generate will return `undefined` for the code and rollup will fail. I will revert this particular change for this release for now as the previous behaviour also better corresponded to this check https://github.com/rollup/rollup/blob/c7981c447677d06a3646cc26299eb5dc357998db/src/rollup/index.ts#L235 and we'll need to find a better solution here

guybedford:
Thanks for spotting this - assuming the example workflow still runs ok
sounds sensible to me.

PR #1922, opened 2018-01-28
add the ability to rewrite modules in-place

kellyselden:
Closes #1878.

This turned out to be remarkably simpler once @guybedford laid the foundational work with dynamic imports and chunking. I essentially tapped into the dynamic imports system and made every import behave like a dynamic import.

To try this out, all you have to do is add the `experimentalPreserveModules` to your existing `experimentalCodeSplitting` setup. Then, the directory tree should be preserved in your output dir.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -28,9 +28,15 @@ import { Node } from './ast/nodes/shared/Node';
 import Chunk from './Chunk';
 import * as path from './utils/path';
 import GlobalScope from './ast/scopes/GlobalScope';
-import { randomUint8Array, Uint8ArrayXor, Uint8ArrayToHexString } from './utils/entryHashing';
+import {
+	randomUint8Array,
+	Uint8ArrayXor,
+	Uint8ArrayToHexString,
+	isZero
+} from './utils/entryHashing';
 import { blank } from './utils/object';
 import firstSync from './utils/first-sync';
+import commondir from './utils/commondir';
 
 export type ResolveDynamicImportHandler = (
 	specifier: string | Node,
@@ -40,9 +46,15 @@ export type ResolveDynamicImportHandler = (
 function generateChunkName(
 	id: string,
 	chunkNames: { [name: string]: boolean },
-	startAtTwo = false
+	startAtTwo = false,
+	inputRelativeDir = ''
 ): string {
-	let name = path.basename(id);
+	let name;
+	if (inputRelativeDir) {
+		name = path.relative(inputRelativeDir, id).replace(/\\/g, '/');
+	} else {
+		name = path.basename(id);
+	}
 	let ext = path.extname(name);
 	name = name.substr(0, name.length - ext.length);
 	if (ext !== '.js' && ext !== '.mjs') {
@@ -313,7 +325,10 @@ export default class Graph {
 		});
 	}
 
-	buildChunks(entryModuleIds: string[]): Promise<{ [name: string]: Chunk }> {
+	buildChunks(
+		entryModuleIds: string[],
+		preserveModules: boolean
+	): Promise<{ [name: string]: Chunk }> {
 		// Phase 1 – discovery. We load the entry module and find which
 		// modules it imports, and import those, until we have all
 		// of the entry module's dependencies
@@ -327,7 +342,12 @@ export default class Graph {
 				timeStart('phase 2');
 
 				this.link();
-				const { orderedModules, dynamicImports } = this.analyseExecution(entryModules);
+
+				const { orderedModules, dynamicImports } = this.analyseExecution(
+					entryModules,
+					preserveModules
+				);
+
 				dynamicImports.forEach(dynamicImportModule => {
 					if (entryModules.indexOf(dynamicImportModule) === -1)
 						entryModules.push(dynamicImportModule);
@@ -382,6 +402,15 @@ export default class Graph {
 				entryModules.forEach(entryModule => {
 					entryModule.chunk.generateEntryExports(entryModule);
 				});
+				if (preserveModules) {
+					// preserve the links between modules
+					// (normally lost when combining modules into chunks)
+					orderedModules.forEach(module => {
+						if (entryModules.indexOf(module) === -1) {
+							module.chunk.generateEntryExports(module, preserveModules);
+						}
+					});
+				}
 				// for each chunk module, set up its imports to other
 				// chunks, if those variables are included after treeshaking
 				chunkList.forEach(chunk => {
@@ -394,13 +423,27 @@ export default class Graph {
 					[name: string]: Chunk;
 				} = {};
 
+				let inputRelativeDir: string;
+				if (preserveModules) {
+					if (orderedModules.length === 1) {
+						inputRelativeDir = path.dirname(orderedModules[0].id);
+					} else {
+						inputRelativeDir = commondir(orderedModules.map(module => module.id));
+					}
+				}
+
 				// name the chunks
 				const chunkNames: { [name: string]: boolean } = blank();
 				chunkNames['chunk'] = true;
 				chunkList.forEach(chunk => {
 					// generate the imports and exports for the output chunk file
 					if (chunk.entryModule) {
-						const entryName = generateChunkName(chunk.entryModule.id, chunkNames, true);
+						const entryName = generateChunkName(
+							chunk.entryModule.id,
+							chunkNames,
+							true,
+							inputRelativeDir
+						);
 
 						// if the chunk exactly exports the entry point exports then
 						// it can replace the entry point
@@ -432,7 +475,7 @@ export default class Graph {
 		);
 	}
 
-	private analyseExecution(entryModules: Module[]) {
+	private analyseExecution(entryModules: Module[], preserveModules: boolean = false) {
 		let curEntry: Module, curEntryHash: Uint8Array;
 		const allSeen: { [id: string]: boolean } = {};
 
@@ -450,7 +493,12 @@ export default class Graph {
 			// entry point and colouring those modules by the hash of its id. Colours are mixed as
 			// hash xors, providing the unique colouring of the graph into unique hash chunks.
 			// This is really all there is to automated chunking, the rest is chunk wiring.
-			Uint8ArrayXor(module.entryPointsHash, curEntryHash);
+			if (!preserveModules) {
+				Uint8ArrayXor(module.entryPointsHash, curEntryHash);
+			} else if (isZero(module.entryPointsHash)) {
+				module.isEntryPoint = true;
+				module.entryPointsHash = randomUint8Array(10);
+			}
 
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -28,12 +28,7 @@ import { Node } from './ast/nodes/shared/Node';
 import Chunk from './Chunk';
 import * as path from './utils/path';
 import GlobalScope from './ast/scopes/GlobalScope';
-import {
-	randomUint8Array,
-	Uint8ArrayXor,
-	Uint8ArrayToHexString,
-	isZero
-} from './utils/entryHashing';
+import { randomUint8Array, Uint8ArrayXor, Uint8ArrayToHexString } from './utils/entryHashing';
 import { blank } from './utils/object';
 import firstSync from './utils/first-sync';
 import commondir from './utils/commondir';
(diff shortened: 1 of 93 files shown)


kellyselden:
Turns out the the folders are not preserved: "app/index.js" => "index.js". I will work on that next.

kellyselden:
Added another option to enable original directory structure.

src/Graph.ts
```diff
@@ -490,7 +521,15 @@ export default class Graph {
...
 		}
 
-		return { orderedModules: ordered, dynamicImports, hasCycles };
+		// new items can be added during this loop
+		for (let i = 0; i < preservedModules.length; i++) {
+			curEntry = preservedModules[i];
+			curEntry.isEntryPoint = true;
+			curEntryHash = randomUint8Array(10);
+			visit(curEntry);
+		}
```
guybedford: It seems like this is done to ensure that no chunking happens right?

I think it can be done more easily by altering the visit function itself to do something like:

```js
if (!preserveModules)
  Uint8ArrayXor(module.entryPointsHash, curEntryHash);
else if (isZero(module.entryPointsHash))
  module.entryPointsHash = randomUint8Array(10);
```

then I think that would be all there is too it without any extra loops or need to track modules.

src/Graph.ts
```diff
@@ -325,12 +339,23 @@ export default class Graph {
...
+
 				dynamicImports.forEach(dynamicImportModule => {
 					if (entryModules.indexOf(dynamicImportModule) === -1)
 						entryModules.push(dynamicImportModule);
 				});
 
+				preservedModules.forEach(module => {
+					if (entryModules.indexOf(module) === -1)
+						entryModules.push(module);
+				});
```
guybedford: This is just "all modules" right? If so it should be possible to generate from Object.values(this.graph) + this.externals?

src/Graph.ts
```diff
@@ -396,7 +421,7 @@ export default class Graph {
 				chunkList.forEach(chunk => {
 					// generate the imports and exports for the output chunk file
 					if (chunk.entryModule) {
-						const entryName = generateChunkName(chunk.entryModule.id, chunkNames, true);
+						const entryName = generateChunkName(chunk.entryModule.id, chunkNames, true, inputRelativeDir);
```
guybedford: Is this a mandatory option then? Is there not a way to deduce the lowest base by default if this is not provided?

kellyselden: I was able to rely on the package [commondir](https://github.com/substack/node-commondir) to calculate given the entry modules. It is used in a subsequent feature request [here](https://github.com/kellyselden/rollup/commit/90ded1adb48521c7b5a07d63a83bcf6d9c2bed5b#diff-ef9931d02c30e05453c44ad0b6281929R8) to make it something other than common base path, but I will move the option into that branch instead.

src/Graph.ts
kellyselden: Perfect! It got a lot simpler with this change.

src/Graph.ts
kellyselden: removed with your other suggested change.

src/Graph.ts
guybedford: Since `commondir` is not widely maintained and a small piece of code, do you think we could just inline the implementation?

guybedford:
It's pretty awesome how small this feature has got. And it definitely preserves the original pathnames ok with directory creation etc?

kellyselden:
From my testing, yes. It does get weird if your entry file imports from '../../../../../x', then your output dir will look pretty strange, but it should still work and be "correct".

src/Graph.ts
kellyselden: I could do that. Are you sure I should just copy paste? I would argue that we depend on it until we need to fix something, then copy. @lukastaegert care to tie-break?

lukastaegert: I must admit I did not read everything here. Looking at commondir, it relies on nodes `path.resolve` which we do not have available in a browser build—even though we have our own polyfill for that (damn, we REALLY need a test for this. You should see a warning about "path" being external when building rollup with this setup—this is always BAD). Thus I fear this needs to go into our own code base somehow.

However I am happy for every bit of functionality we can extract from Graphs.ts and put into other files and for me, this is a clear candidate for the "utils" folder.

kellyselden: ported into the utils folder.

src/utils/commondir.ts
```diff
@@ -0,0 +1,21 @@
+// ported from https://github.com/substack/node-commondir
+export default function commondir (relfiles: string[]) {
+    var files = relfiles;
```
guybedford: Perhaps just name the argument `files`/

src/utils/commondir.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    var res = files.slice(1).reduce(function (ps, file) {
```
guybedford: `const` and arrow function can be used.

src/utils/commondir.ts
```diff
@@ -0,0 +1,21 @@
...
+        if (!file.match(/^([A-Za-z]:)?\/|\\/)) {
+            throw new Error('relative path without a basedir');
+        }
```
guybedford: If we know that our inputs will pass this, it may be possible to remove this check.

src/utils/commondir.ts
```diff
@@ -0,0 +1,21 @@
...
+
+        var xs = file.split(/\/+|\\+/);
```
guybedford: `const`

src/utils/commondir.ts
```diff
@@ -0,0 +1,21 @@
...
+        for (
+            var i = 0;
+            ps[i] === xs[i] && i < Math.min(ps.length, xs.length);
+            i++
```
guybedford: Tailing semicolon.

src/Graph.ts
```diff
@@ -28,14 +28,25 @@ import { Node } from './ast/nodes/shared/Node';
...
+	chunkNames: { [name: string ]: boolean },
+	startAtTwo = false,
+	inputRelativeDir = ''
+): string {
+	let name;
+	if (inputRelativeDir) {
+		name = path.relative(inputRelativeDir, id).replace(/\\/g, '/');
+	} else {
+		name = path.basename(id);
+	}
```
guybedford: Check indentation here.

lukastaegert:
Sure thing. Was a little busy the last days to get my own PR in order. Will probably take a look at this evening or tomorrow. I think this could become on of the core features of 0.57.0!

lukastaegert:
BTW some of the chunking form tests seem to be broken. And this should probably be rebased against the current master soon.

kellyselden:
@lukastaegert I discovered a bug, so I added more test cases. I'm trying to diagnose it now.

src/utils/commondir.ts
kellyselden: fixed

src/utils/commondir.ts
kellyselden: fixed

src/utils/commondir.ts
kellyselden: removed

src/utils/commondir.ts
kellyselden: fixed

src/utils/commondir.ts
kellyselden: compiler didn't like that

src/Graph.ts
kellyselden: fixed

kellyselden:
@lukastaegert @guybedford Turns out it might not be a bug, but an understanding issue. It has to do with what's going on in \test\chunking-form\samples\chunking-reexport. The way it's collapsing reexports there is unexpected to me, and probably not desirable for this new feature. I would love some pointers.

guybedford:
@kellyselden can you link me to the test case that is failing? I just re-ran the CI here and it seems nothing is breaking specifically. If I can follow the exact simplified scenario I may be able to suggest further.

kellyselden:
@guybedford https://github.com/rollup/rollup/pull/1922/files#diff-c7e8e1a25cc0ed7f80f7f0df33f956c7R7
I skipped it because it appears to be working as you intended after comparing with \test\chunking-form\samples\chunking-reexport, I just don't understand why it is working that way, and if I should change it for this new feature. The expected

src/Graph.ts
lukastaegert: One thing I observed that I think we should definitely improve is the fact that if a module imports something from outside the common directory of the entry points (i.e. if the entry point imports something that starts with `../`), new files and folders are created outside the intended target directory. This can definitely leave a mess on an unsuspecting user's system. In my opinion there are two possible solutions
1. throw an error if something like this would happen (probably not ideal), or
2. find the common root directory not among the entry points but among the generated chunk file names.

To achieve the latter, we could do the following:
* use `generateChunkName` only for the non-path-preserving case (and remove the last argument)
* if we preserve paths/modules, we use the completed `chunks` object to build a new object with correct paths (possibly extract a helper method here)

This is just a rough idea, some fine-tuning may be necessary.

src/rollup/index.ts
```diff
@@ -363,7 +364,10 @@ export default function rollup (rawInputOptions: GenericConfigObject) {
 				return result;
 			});
```
lukastaegert: Another issue I observed (the best location for this comment would actually be at line 237) is that implicitly, this feature requires BOTH experimental code-splitting AND input to be an array. That is, if I just provide a single entry point not wrapped in an array, nothing will happen. Ideally I would expect this feature to work the following way:
* `experimentalPreserveModules` implicitly enables `experimentalCodeSplitting`
* If `experimentalPreserveModules` is enabled and `input` is not an array, it is treated as an array of length 1.
This should mostly be possible by changing the check in line 237 and transforming the input if necessary.

src/utils/commondir.ts
```diff
@@ -0,0 +1,16 @@
...
+			i = 0;
+			ps[i] === xs[i] && i < Math.min(ps.length, xs.length);
+			i++
+		);
+		return ps.slice(0, i);
+	}, files[0].split(/\/+|\\+/));
+
+	// Windows correctly handles paths with forward-slashes
+	return res.length > 1 ? res.join('/') : '/';
+};
```
lukastaegert: May be nice to make the variables a little more descriptive now that we have inlined this. `xs` e.g. seems to be something like a `path` or `pathSegements`, `ps` is something like a `commonPath`, and `res` could be a `resultingPath`. Just some ideas to get rid of these completely arbitrary names.

lukastaegert:
Another thing to note that we may want to skip for the first version but should definitely keep in mind:
* Unused imports can lead to empty files being generated and imported. I understand that at the moment, there is no easy way to detect whether an import needs to be included for side-effects. I would leave this behaviour as it is for the first version but as a future improvement, we might consider e.g. adding `.included` flags to `Program` nodes as well (that become `true` once some statements are included) and make sure there is a way for import statements to check the imported module has an included `Program` node. Then
    * the `.included` flag of imports could just reflect this flag (thus tree-shaking would remove them automatically if necessary)
    * non-included modules can be omitted
* Alternatively when empty files are generated, we print the warning "Generated an empty bundle" once for each empty file. This is probably rather irritating to users. If we do not omit these files altogether, we may want to do two things:
    * Consolidate these warnings
    * Print "Generated x empty files" (as we are no longer bundling)

guybedford:
@lukastaegert if there's a way to have `included` on the Program node itself that would be a great feature I think! Already seeing some empty bundles in chunking testing here.

Another thing that gets interesting is to be able to treat "pure code" as a movable entity (bounded by its non-pure usages in terms of ensuring the same TDZ behaviours) between chunks to try and get a better bundling. That will be tricky in current architecture but mentioning it so we can perhaps put some thought to that too as that's another next step in squeezing out these optimizations. At that point the entire "chunk of modules" concept breaks down into "chunk of blocks of pure and side-effect code" where import and export statements are reconstructed between those.

kellyselden:
@lukastaegert Fixed the "imports from outside" issue via this method
> find the common root directory not among the entry points but among the generated chunk file names.

src/rollup/index.ts
kellyselden: fixed

src/utils/commondir.ts
kellyselden: updated names

kellyselden:
@lukastaegert added a little info to the description of how to apply to an existing code splitting setup.

kellyselden:
Ready for another review.

src/Graph.ts
```diff
@@ -372,6 +394,15 @@ export default class Graph {
...
 				});
+				if (preserveModules) {
+					// preserve the links between modules
+					// (normally lost when combining modules into chunks)
+					orderedModules.forEach(module => {
+						if (entryModules.indexOf(module) === -1) {
+							module.chunk.generateEntryExports(module, preserveModules);
+						}
+					});
+				}
```
guybedford: This changes things a little unfortunately. I was under impression that this approach was based on "treating all modules as entry points" which would mean exposing all their exports. But it sounds like we still want tree shaking of exports to apply here.

As such, we're not really treating all modules as entry points, so perhaps my previous code suggestions were a little misleading as that was my impression.

What happens if this entire section is removed along with the `isEntryPoint` being set in the analyseExecution method? That seems like it would get us closer to the consistent conceptual model with this here.

kellyselden: I tried that but the code that checks `if (chunk.entryModule) {` or `if (chunk.isEntryModuleFacade) {` is now misbehaving. I tried fixing those, but then the object diffs were not making any sense to me:
capture

guybedford: You might get that if there were no entry points at all... Perhaps in the case of no entry points all modules should be treated as entry points, but if there are entry points, then nothing else should be treated as an entry point - is that the sort of workflow you're going for here?

kellyselden: No. There is always at least one entry point for this feature.

kellyselden:
I'm a little lost on how I can progress this further. Making the suggested changes made the feature start falling apart in ways that didn't make sense to me. Perhaps this can be merged in a non-ideal but working state, and then iterated upon, or maybe @guybedford can try tweaking the code to get a better outcome than I did. Either that, or I'm open to other options.

guybedford:
@kellyselden I'll have a look this week and see if I can suggest something.

guybedford:
@kellyselden I've pushed a change here to treat entry points separately from other modules. Combined with https://github.com/rollup/rollup/pull/2021 that should get things working!

There is one other critical issue that we should discuss here though which is the following test case:

```js
var p = 5;
export { p, p as q }
```

will currently fail for preserveModules. We may need to create some custom wrapper code at the end to handle this scenario, as code splitting is very much based on one variable exposed per module boundary due to rendering restrictions.

kellyselden:
That scenario would rewrite usages of the alias to the original in all the dependents, right? I can see how that would be a problem if you were treating all your modules' exports as public API. In Ember's case on the other hand, since we are bundling for the app, the fact that that alias is removed wouldn't be a problem as long as it's consistent throughout the graph.

That said, I combined your changes here with #2021, and it worked great in my Ember tests. Excellent work @guybedford! Surprisingly, I also tried importing lodash-es into this, and it worked. So something you did fixed #2015 without the need for #2020. You may already know this, but just wanted to point it out.

guybedford:
Right, it would only be entry points that happened to export the same variable twice that would be a problem. Which I guess is enough of an edge case that it can be specifically resolved if and when it is hit.

It would still be nice to have a test case along those exact lines to codify the existing behaviour whatever that might be, and ensure there isn't a deeper bug exposed.

kellyselden:
@guybedford I added the test for export aliases. Since there are still two commits, I added the test to the first one, and export aliases worked as expected. Then I had to modify the test to remove the aliases once your commit was added on top. You probably already knew this, I just wanted to point it out in case you didn't.

olstenlarck:
This option is still not documented on the site? It seems that i just should try it ;/

guybedford:
I think we should document this now actually. A documentation PR is more than welcome!

PR #1924, opened 2018-01-29
enabled testing directory structures in chunked mode

kellyselden:
Extracted from #1922.

Allows comparing directory trees, instead of just top-level files.


kellyselden:
Any objections to this?

PR #1940, opened 2018-02-04
add chunking failing test

kellyselden:
I'm doing something similar with #1922, but I may have found a bug. This creates `Error: Could not resolve entry (_DataView.js)`, but it should rollup fine to my understanding.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -901,6 +901,12 @@
         }
       }
     },
+    "ensure-posix-path": {
+      "version": "1.0.2",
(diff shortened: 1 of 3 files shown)


guybedford:
@kellyselden thanks for posting this. Ideally we should isolate this without adding a new dependency. Do you think you could try and narrow it down further?

kellyselden:
Yea it wasn't meant to be the final test, just a starting point for whoever takes it on. I'll try to isolate it in the coming days.

kellyselden:
Closing for now, probably isn't an issue.

Issue #1941, opened 2018-02-04
Reinstate cycle detection


adrianheine:
Circular dependencies regularly create issues (rollup/rollup-plugin-commonjs#248, rollup/rollup-plugin-commonjs#284). It would be great to detect them, and apparently that used to be the case, but it haven't been working for quite a while, was marked as broken in the codebase and removed, see 6800aa1844e9c2efb6c067875d9ae0da89f0d459 for the last commit that removed leftovers.

PR #1942, opened 2018-02-04
Add simple warning for circular dependencies

adrianheine:
See #1941. This is very primitive and doesn't warn through rollup's logging system.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -268,10 +268,7 @@ export default class Graph {
 
 				this.link();
 
-				const { orderedModules, dynamicImports, hasCycles } = this.analyseExecution([entryModule]);
-				if (hasCycles) {
-					this.warnCycle(entryModule, orderedModules);
-				}
+				const { orderedModules, dynamicImports } = this.analyseExecution([entryModule]);
 
 				timeEnd('phase 2');
 
@@ -428,20 +425,14 @@ export default class Graph {
 	}
 
 	private analyseExecution (entryModules: Module[]) {
-		let hasCycles = false, curEntry: Module, curEntryHash: Uint8Array;
+		let curEntry: Module, curEntryHash: Uint8Array;
 		const allSeen: { [id: string]: boolean } = {};
 
 		const ordered: Module[] = [];
 
 		const dynamicImports: Module[] = [];
 
-		const visit = (module: Module, seen: { [id: string]: boolean } = {}) => {
-			if (seen[module.id]) {
-				hasCycles = true;
-				return;
-			}
-			seen[module.id] = true;
-
+		const visit = (module: Module, parents: { [id: string]: string | null } = { [module.id]: null }) => {
 			if (module.isEntryPoint && module !== curEntry)
 				return;
 
@@ -453,7 +444,15 @@ export default class Graph {
 
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
-					visit(<Module>depModule, seen);
+					if (depModule.id in parents) {
+						if (!allSeen[depModule.id]) {
+							this.warnCycle(depModule.id, module.id, parents);
+						}
+						return;
+					}
+
+					parents[depModule.id] = module.id;
+					visit(<Module>depModule, parents);
 				}
 			});
 
@@ -490,11 +489,24 @@ export default class Graph {
 			visit(curEntry);
 		}
 
-		return { orderedModules: ordered, dynamicImports, hasCycles };
+		return { orderedModules: ordered, dynamicImports };
 	}
 
-	private warnCycle (_entryModule: Module, _ordered: Module[]) {
-		// TODO: reinstate
+	private warnCycle (id: string, parentId: string, parents: { [id: string]: string | null }) {
+		const path = [relativeId(id)];
+		let curId = parentId;
+		while (curId !== id) {
+			path.push(relativeId(curId));
+			curId = parents[curId];
+			if (!curId) break;
+		}
+		path.push(path[0]);
+		path.reverse();
+		this.warn({
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: path[0],
+			message: `Circular dependency: ${path.join(' -> ')}`
+		});
 	}
 
 	private fetchModule (id: string, importer: string): Promise<Module> {
--- a/test/function/samples/cycles-pathological-2/_config.js
+++ b/test/function/samples/cycles-pathological-2/_config.js
@@ -3,9 +3,21 @@ var assert = require( 'assert' );
 module.exports = {
 	description: 'resolves even more pathological cyclical dependencies gracefully',
 	buble: true,
-	options: {
-		onwarn: function ( message ) {
-			assert.ok( /unable to evaluate without/.test( message ) );
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'main.js',
+			message: 'Circular dependency: main.js -> b.js -> main.js'
+		},
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'b.js',
+			message: 'Circular dependency: b.js -> d.js -> c.js -> b.js'
+		},
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'main.js',
+			message: 'Circular dependency: main.js -> b.js -> d.js -> c.js -> main.js'
 		}
-	}
+	]
 };
--- a/test/function/samples/already-deshadowed-import/_config.js
+++ b/test/function/samples/already-deshadowed-import/_config.js
@@ -1,3 +1,10 @@
 module.exports = {
 	description: 'handle already module import names correctly if they are have already been deshadowed',
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'bob.js',
+			message: 'Circular dependency: bob.js -> alice.js -> bob.js'
+		}
+	]
 };
--- a/test/function/samples/can-import-self-treeshake/_config.js
+++ b/test/function/samples/can-import-self-treeshake/_config.js
@@ -2,6 +2,11 @@ module.exports = {
 	description: 'direct self import',
 	warnings: [
 		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'lib.js',
+			message: 'Circular dependency: lib.js -> lib.js'
+		},
+		{
 			code: 'EMPTY_BUNDLE',
 			message: `Generated an empty bundle`
 		}
--- a/test/function/samples/can-import-self/_config.js
+++ b/test/function/samples/can-import-self/_config.js
@@ -4,5 +4,12 @@ module.exports = {
 	description: 'a module importing its own bindings',
 	exports: function ( exports ) {
 		assert.equal( exports.result, 4 );
-	}
+	},
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'lib.js',
+			message: 'Circular dependency: lib.js -> lib.js'
+		}
+	]
 };
--- a/test/function/samples/cycles-default-anonymous-function-hoisted/_config.js
+++ b/test/function/samples/cycles-default-anonymous-function-hoisted/_config.js
@@ -1,3 +1,10 @@
 module.exports = {
-	description: 'Anonymous function declarations are hoisted'
+	description: 'Anonymous function declarations are hoisted',
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'f.js',
+			message: 'Circular dependency: f.js -> g.js -> f.js'
+		}
+	]
 };
--- a/test/function/samples/cycles-defaults/_config.js
+++ b/test/function/samples/cycles-defaults/_config.js
@@ -1,5 +1,12 @@
 module.exports = {
-	description: 'cycles work with default exports'
+	description: 'cycles work with default exports',
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'a.js',
+			message: 'Circular dependency: a.js -> b.js -> a.js'
+		}
+	]
 };
 
 // test copied from https://github.com/esnext/es6-module-transpiler/tree/master/test/examples/cycles-defaults
--- a/test/function/samples/cycles-immediate/_config.js
+++ b/test/function/samples/cycles-immediate/_config.js
@@ -1,5 +1,12 @@
 module.exports = {
-	description: 'handles cycles where imports are immediately used'
+	description: 'handles cycles where imports are immediately used',
+	warnings: [
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'evens.js',
+			message: 'Circular dependency: evens.js -> odds.js -> evens.js'
+		}
+	]
 };
 
 // Test copied from https://github.com/esnext/es6-module-transpiler/tree/master/test/examples/cycles-immediate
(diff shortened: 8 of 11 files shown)


src/Graph.ts
```diff
@@ -437,7 +434,9 @@ export default class Graph {
 
 		const visit = (module: Module, seen: { [id: string]: boolean } = {}) => {
 			if (seen[module.id]) {
-				hasCycles = true;
+				if (!allSeen[module.id]) {
+					this.warnCycle(module);
```
guybedford: For even better debugging perhaps pass the full cycle stack?

```js
let cycleStack = [];
Object.keys(seen).forEach(name => {
  if (cycleStack) {
    cycleStack.push(name);
  } else if (name === module.id) {
    cycleStack = [];
  }
});
this.warnCycle(module, cycleStack);
```

Then something like `stack: cycleStack.join(' -> ')` in the log.

guybedford: Ideally the above should be relative to the base like we do in other places to be readable.

guybedford: Sorry, being silly here, this isn't a parent-based structure so doesn't capture the path of course.

adrianheine:
Definitely, this was just a POC.

src/Graph.ts
adrianheine: I'll try to get the full cycle. I have an idea how to do that :)

adrianheine: `relativeId` strips `\0000commonjs-proxy:` prefixes :(

src/Graph.ts
```diff
@@ -428,16 +425,21 @@ export default class Graph {
...
 		const allSeen: { [id: string]: boolean } = {};
 
 		const ordered: Module[] = [];
 
 		const dynamicImports: Module[] = [];
 
+		const cycles: {id: string, path: { [id: string]: string | null }}[] = []
+
 		const visit = (module: Module, seen: { [id: string]: boolean } = {}) => {
+			const modId = relativeId(module.id);
```
guybedford: Can we move this into the if statement?

src/Graph.ts
```diff
@@ -428,16 +425,21 @@ export default class Graph {
...
 			if (seen[module.id]) {
-				hasCycles = true;
+				if (!allSeen[module.id]) {
+					cycles.push({id: modId, path: {[modId]: null}});
```
guybedford: Spacing - `cycles.push({ id: modId, path: { [modId]: null } });`

src/Graph.ts
```diff
@@ -454,6 +456,15 @@ export default class Graph {
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
 					visit(<Module>depModule, seen);
+					const depId = relativeId(depModule.id);
```
guybedford: Same here.

src/Graph.ts
```diff
@@ -490,11 +501,25 @@ export default class Graph {
...
 
-		return { orderedModules: ordered, dynamicImports, hasCycles };
+		cycles.forEach(c => this.warnCycle(c))
+
+		return { orderedModules: ordered, dynamicImports };
 	}
 
-	private warnCycle (_entryModule: Module, _ordered: Module[]) {
-		// TODO: reinstate
+	private warnCycle ({id, path}: {id: string, path: { [id: string]: string | null }}) {
```
guybedford: spacing

src/Graph.ts
```diff
@@ -454,6 +456,15 @@ export default class Graph {
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
 					visit(<Module>depModule, seen);
+					const depId = relativeId(depModule.id);
+					cycles.forEach(cycle => {
+						if (cycle.path[depId] === null) {
+							cycle.path[depId] = modId;
```
guybedford: `modId` can just be recomputed.

guybedford: Alternatively we could do the relative pathing in the warning function itself as a map.

src/Graph.ts
```diff
@@ -454,6 +456,15 @@ export default class Graph {
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
 					visit(<Module>depModule, seen);
+					const depId = relativeId(depModule.id);
+					cycles.forEach(cycle => {
+						if (cycle.path[depId] === null) {
```
guybedford: The complexity of this loop does worry me.... cyclical code shouldn't get a performance penalty. Ideally as much of the looping could be moved into the warning function itself, with only markers provided from this main iteration loop (eg parent information etc).

Having these extra loops within the main visit function doesn't seem ideal.

src/Graph.ts
```diff
@@ -454,6 +456,15 @@ export default class Graph {
...
+							cycle.path[depId] = modId;
+							if (cycle.id !== modId) {
+								cycle.path[modId] = null;
+							}
```
guybedford: I will have to think about this further as I'm not sure I understand all the edge cases. As I say I'd be much happier merging if we can move as much of this logic into the warning function itself, with simple non-loop-based analysis passed through rather (eg having the `seen` object provide a parent map or similar information that doesn't involve extra looping to create).

Otherwise I'll want to test this much more thoroughly for edge cases and loop complexity.

src/Graph.ts
adrianheine: It's re-used further down.

src/Graph.ts
adrianheine: What, moving into the loop?

src/Graph.ts
adrianheine: Ok.

src/Graph.ts
adrianheine: This is only looping over the list of cycles, which usually should be 0 or 1.

src/Graph.ts
adrianheine: Ok. I tried doing it in the warning function, but seemed more complex. I can try again, though.

src/Graph.ts
adrianheine: Ok.

src/Graph.ts
adrianheine: Ok, it's actually pretty straightforward.

adrianheine:
Sure, if this is otherwise good to go I can add some tests.

src/Graph.ts
```diff
@@ -454,6 +455,14 @@ export default class Graph {
...
 				if (!depModule.isExternal) {
 					visit(<Module>depModule, seen);
+					cycles.forEach(cycle => {
+						if (cycle.path[depModule.id] === null) {
+							cycle.path[depModule.id] = module.id;
+							if (cycle.id !== module.id) {
+								cycle.path[module.id] = null;
+							}
+						}
+					})
```
guybedford: I'd still like to see the complexity reduced here if possible. We are looping each cycle for each dependency for each module and doing a dictionary lookup. A graph like A -> (B,C,D), B -> A, C -> A, D -> ... where the D branch is deep will result in unnecessary looping over the former A,B,C cycles for each module of the D branch, which seems very unnecessary.

adrianheine: Just to make sure we have the same understanding of this code: Your example will result in three dictionary lookups per module in the `D` branch, right?

adrianheine: I can imagine marking the cycle as closed in L461 (by moving it out of `cycles` into some other array), that would remove unnecessary lookups in your example, but there are probably pathological cases were this wouldn't help at all.

guybedford: Yes. This is why I was thinking of parent-based seen as one could do:

```js
if (seen[id]) {
  let cycleStack = [];
  let parentId;
  while (parentId = seen[id]) {
    cycleStack.push(parentId);
    id = parentId;
  }
}
seen[id] = parentId; // with the parent added to the visitor to do this...
```

Each individual `seen` will only have one parent path per module by the nature of `seen`, guaranteeing the above stack which then only iterates the exact path. Although I may be overlooking some of the details.

adrianheine: That would make sense, since `seen` is only used as a boolean anyway.

guybedford: Note that we may catch cases that aren't actually cycles - eg:

```
A -> B -> C
    -> C
```

C will be in seen, when it is hit again as a parallel dependency not a cyclical one. The way to confirm that it isn't cyclical is that when running up the parent path from the current C (direct dependency of A), we can note that we don't come back around to C. Whether as if we had:

```
C -> A -> B -> C
```

When we come across the second seen "C" and run up the parent stack from the current id, we do find the loop.

adrianheine: Are you sure about this? It looks to me like modules are traversed depth-first.

src/Graph.ts
```diff
@@ -453,7 +446,22 @@ export default class Graph {
...
 				if (!depModule.isExternal) {
-					visit(<Module>depModule, seen);
+					if (depModule.id in parents) {
+						if (!allSeen[depModule.id]) {
+							let path = [depModule.id];
+							let id = module.id;
+							while (id !== depModule.id) {
+								path.push(id);
+								id = parents[id];
+								if (!id) break;
```
adrianheine: This should never happen imho.

guybedford: When the top-level import is hit I think?

adrianheine: Yeah, which means there is no cycle, in which case I think this code is not triggered.

guybedford: Right of course. How's the output looking?

adrianheine: For example like this:

> (!) Circular dependency: node_modules/readable-stream/lib/_stream_readable.js -> node_modules/readable-stream/lib/_stream_duplex.js -> commonjs-proxy:/tmp/rollup-stream-issue-repro/node_modules/readable-stream/lib/_stream_readable.js -> node_modules/readable-stream/lib/_stream_readable.js

guybedford: Nice, if we can just get a test with one of these outputs sounds like we're nearly there?

adrianheine: I just updated the function tests to expect this warning, so I think that's covered.

guybedford: Great I guess we can remove this break then as well?

test/function/samples/cycles-pathological-2/_config.js
```diff
@@ -3,9 +3,21 @@ var assert = require( 'assert' );
...
+		},
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'b.js',
+			message: 'Circular dependency: b.js -> d.js -> c.js -> b.js'
+		},
+		{
+			code: 'CIRCULAR_DEPENDENCY',
+			importer: 'main.js',
+			message: 'Circular dependency: main.js -> b.js -> d.js -> c.js -> main.js'
```
guybedford: Ideally we should avoid warning each module in the cycle I think for readability with these messages?

One option would be to filter by unique cycle here, although that may be quite involved. Another option might be to just permit one cycle warning per module (perhaps keying `cycles` by id and then checking each module against this when populating a cycle path, skipping new cycles that have a match of a module here).

src/Graph.ts
adrianheine: If my assumption is wrong this will end in an infinite loop, so I would rather like to keep it. Instead of breaking we could just return, though, and maybe emit a warning, since hitting this is a clear sign of a mistake in my reasoning.

test/function/samples/cycles-pathological-2/_config.js
adrianheine: I think these are three different cycles. The last two share the `b -> d -> c` part of the path, but otherwise they are different.

src/Graph.ts
guybedford: I'm still wondering about the case where a imports b and c, and b imports c. As our visit function itself is written as a pre-order visitor, we will call:

* visit(a) for the top-level
* visit(b) for a
* visit(c) for b
* then visit(c) for a again

So I still seem to think that we'd have `c` populated in parents by the time the `visit(c)` for `a` happens. Which then would trigger this case?

test/function/samples/cycles-pathological-2/_config.js
adrianheine: I agree, though, that in real-world code cycles seem to overlap a lot or share a root cause and deduplication would be nice.

src/Graph.ts
adrianheine: No, `visit(c)` for `b` would have finished by then, so that `c` would be in `allSeen`.

test/function/samples/cycles-pathological-2/_config.js
guybedford: Ok I see they are fully distinct already so perhaps this won't be necessary. We can always see how it works out in real world cases and if it gets too noisy can then consider such filtering further.

src/Graph.ts
guybedford: Ahh right. Interesting how allSeen gives us some great higher level properties, which was completely unintended :P

src/Graph.ts
```diff
@@ -453,7 +446,22 @@ export default class Graph {
...
+							}
+							cycles.push({ id: depModule.id, path });
```
guybedford: I think we can now move all the logic into the cycle warning function as mentioned before.

Say - `this.warnCycles(id, parents)`, with the logic in there?

Would just be good to keep conceptual boundaries in place.

guybedford:
Just tested this out and it's working perfectly.

In my example building Babel I get 18 cyclical dependency warnings, whereas with the filtering this would be reduced to around 4 or 5 I think.

src/Graph.ts
adrianheine: `parents` have to be frozen at the time the cycle is detected since there can be a lot of parents per module. I moved the path logic to `warnCycle`, it's not as nice as it was here imho, but probably worth it.

src/Graph.ts
```diff
@@ -453,7 +446,16 @@ export default class Graph {
 
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
-					visit(<Module>depModule, seen);
+					if (depModule.id in parents) {
+						if (!allSeen[depModule.id]) {
+							parents[depModule.id] = module.id;
+							cycles.push({ id: depModule.id, parents: { ...parents } });
```
guybedford: Could we not just output the warning directly here instead of collecting the cycles list?

src/Graph.ts
adrianheine: The last patch doesn't correctly work. Sometimes the cycle doesn't end where it started.

src/Graph.ts
adrianheine: Sure, I just did that.

src/Graph.ts
```diff
@@ -453,7 +444,15 @@ export default class Graph {
 
 			module.dependencies.forEach(depModule => {
 				if (!depModule.isExternal) {
-					visit(<Module>depModule, seen);
+					if (depModule.id in parents) {
+						if (!allSeen[depModule.id]) {
+							this.warnCycle(depModule.id, { ...parents, [depModule.id]: module.id });
```
guybedford: The parents expansion shouldn't be necessary now that this is inline I believe?

adrianheine: We may not alter `parents` here because that would break for other cycles involving the same modules. We could do `Object.assign(Object.create(parents}, { [depModule.id]: module.id })`, that's a lot simpler to create but takes more time on lookup.

guybedford: But parents isn't changed while `warnCycle` runs synchronously right?

adrianheine: No, but we may not set `parents[depModule.id]` either. We could pass `module.id` separately, though.

guybedford: Yes that sounds sensible.

adrianheine: Did that. Looks better :)

guybedford:
Thanks for your work on this! If there's too much noise, perhaps we could provide an option to turn off these warnings, but lets see how it goes first.

test/function/samples/cycles-pathological-2/_config.js
diegocr: Just rechecking https://github.com/rollup/rollup-plugin-commonjs/issues/248 found about this PR which i did overlooked earlier. Great work here, looking forward to full circular references support, though :-)

Issue #1976, opened 2018-02-16
Code splitting erroneously implementing renamed exports


jthoms1:
Reproducible in the following repo: https://github.com/jthoms1/rollup-issues

If you have multiple entry points with `experimentalCodeSplitting` flag and an entry point file contains an `export { A as ItemA} ` this carried into the common chunk file and is also kept in the original entry point. This causes an error at runtime.

I would expect the `export as` naming to only exist in the entry point file and not carried into the common chunk file.

Apologies for not explaining this well but the repo is very straight forward the the issue is easily reproducible.

guybedford:
This should be fixed in 0.56.3.

PR #1977, opened 2018-02-17
Export naming fix for #1976

guybedford:
This provides a fix for #1976 where the re-export handling needs to be adjusted to deal with internally renamed exports instead of exporting the original variable name.


guybedford:
I've updated this PR to a fix now.

PR #1987, opened 2018-02-19
remove private api from MissingExport hook

kellyselden:
Closes #1973.

I don't think calling `Module` functions is going to be necessary after all.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -14,7 +14,6 @@ import { isAbsolute, isRelative, normalize, relative, resolve } from './utils/pa
 import {
 	InputOptions,
 	IsExternalHook,
-	MissingExportHook,
 	Plugin,
 	ResolveIdHook,
 	RollupWarning,
@@ -69,7 +68,12 @@ export default class Graph {
 	isPureExternalModule: (id: string) => boolean;
 	legacy: boolean;
 	load: (id: string) => Promise<SourceDescription | string | void>;
-	handleMissingExport: MissingExportHook;
+	handleMissingExport: (
+		exportName: string,
+		importingModule: Module,
+		importedModule: string,
+		importerStart?: number
+	) => void;
 	moduleById: Map<string, Module | ExternalModule>;
 	modules: Module[];
 	onwarn: WarningHandler;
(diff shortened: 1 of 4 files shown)


src/Graph.ts
```diff
@@ -63,7 +62,7 @@ export default class Graph {
 	isPureExternalModule: (id: string) => boolean;
 	legacy: boolean;
 	load: (id: string) => Promise<SourceDescription | string | void>;
-	handleMissingExport: MissingExportHook;
+	handleMissingExport: (module: Module, name: string, otherModule: string, start?: number) => void;
```
guybedford: I take it `module` is the importer here, and start is the index in the importer? How about `name, importingModule, importedModule, importerStart`?

kellyselden:
https://github.com/jmurphyau/ember-truth-helpers/blob/v2.0.0/app/helpers/not-eq.js#L1
reexports `notEq`
https://github.com/jmurphyau/ember-truth-helpers/blob/v2.0.0/addon/helpers/not-equal.js#L3
`notEq` doesn't exist, only `notEqualHelper` is exported.
It was probably a refactor gone wrong. But when compiled down to AMD modules for the browser, the code doesn't throw because everything is essentially properties on a "global" export. If you were to try to call the broken code, it would throw, but if you don't use it, the app works fine.

```js
define('ember-truth-helpers/helpers/not-equal', ['exports'], function (exports) {
  'use strict';

  Object.defineProperty(exports, "__esModule", {
    value: true
  });
  exports.notEqualHelper = notEqualHelper;
  function notEqualHelper(params) {
    return params[0] !== params[1];
  }

  exports.default = Ember.Helper.helper(notEqualHelper);
});
```

```js
define('my-app/helpers/not-eq', ['exports', 'ember-truth-helpers/helpers/not-equal'], function (exports, _notEqual) {
  'use strict';

  Object.defineProperty(exports, "__esModule", {
    value: true
  });
  Object.defineProperty(exports, 'default', {
    enumerable: true,
    get: function () {
      return _notEqual.default;
    }
  });
  Object.defineProperty(exports, 'notEq', {
    enumerable: true,
    get: function () {
      return _notEqual.notEq;
    }
  });
});
```

This is why I need a missing export hook, and a way to continue the graph traversal for backwards compatibility. The hook allows me to return true so rollup can continue without stopping, and a way to print a deprecation message to Ember users. Does that make sense?

src/Graph.ts
kellyselden: updated the wording and order.

guybedford:
Completely, thanks for explaining it to me here! The thing to note is that this won't ever support the `es` module output and will likely throw an internal error or have other undefined behaviour if this early error is skipped.

So ideally we should actually remove the missing export from the analysis entirely so that tracing can go on without it I think.

guybedford:
Actually thinking through the cases here I believe this will output correctly even for es modules with code splitting, as we only create the import/exports where tracing succeeds.

So this looks good to merge here to me.

PR #1990, opened 2018-02-19
revert chunking form test change

kellyselden:
Extracted from #1922. Reverted from #1975.

`fixturify` comparisons where brought in because of the preserve modules work, where there can be a directory tree. The change to this file doesn't work because it only iterates top-level entries.

### DIFF
--- a/test/chunking-form/index.js
+++ b/test/chunking-form/index.js
@@ -62,11 +62,7 @@ describe('chunking form', () => {
 								expectedFiles = [];
 							}
 
-							assert.deepEqual(Object.keys(actualFiles), Object.keys(expectedFiles));
-
-							Object.keys(actualFiles).forEach(fileName => {
-								assert.strictEqual(actualFiles[fileName], expectedFiles[fileName], 'Unexpected output for ' + fileName);
-							});
+							assert.deepEqual(actualFiles, expectedFiles);
 						});
 					});
 				});
--- a/test/chunking-form/index.js
+++ b/test/chunking-form/index.js
@@ -62,7 +62,16 @@ describe('chunking form', () => {
 								expectedFiles = [];
 							}
 
-							assert.deepEqual(actualFiles, expectedFiles);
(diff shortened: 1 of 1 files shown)


guybedford:
The reason I made this change was because when file sources don't match we were getting an object diff instead of a string diff.

What is the data structure for nested files that it won't work with this approach? Surely paths should be flat in the object?

Otherwise, I guess the string diff approach can be adapted for the nested object structure please.

kellyselden:
The object diff looks pretty readable to me:

screen shot 2018-02-20 at 10 00 44 am

This shows all the errors, instead of failing on the first one.

What don't you like about this?

kellyselden:
Added a recursive string comparison:

screen shot 2018-02-20 at 10 31 38 am

kellyselden:
Now it only fails on the first file mismatch, instead of on all file mismatches like before:

capture

Just want to make you aware.

test/chunking-form/index.js
```diff
@@ -62,11 +62,16 @@ describe('chunking form', () => {
...
+							(function recurse(actualFiles, expectedFiles, dirs) {
+								const fileNames = Array.from(new Set(Object.keys(actualFiles).concat(Object.keys(expectedFiles))));
+								fileNames.forEach(fileName => {
+									const sections = dirs.concat(fileName);
+									if (typeof actualFiles[fileName] === 'object' && typeof expectedFiles[fileName] === 'object') {
+										return recurse(actualFiles[fileName], expectedFiles[fileName], sections);
+									}
+									assert.strictEqual(actualFiles[fileName], expectedFiles[fileName], 'Unexpected output for ' + sections.join('/'));
+								});
+							})(actualFiles, expectedFiles, []);
```
lukastaegert: Any reason you are doing this as an IIFE and not just define a function variable and call it? Then you could even extract it as a helper out of this test and make the test more readable (provided you give it a better name than `recurse` 😉)

PR #1991, opened 2018-02-20
failing test for default export renames

kellyselden:
_expected is _actual in this case, that's why tests are passing. This is so you can see the failed rename in main.js easily.


guybedford:
Thanks, failing test cases are a huge help!

I'm pretty sure this should be covered by https://github.com/rollup/rollup/pull/2001 now, with the test case there.

kellyselden:
Yep #2001 fixes this.

Issue #1817, opened 2017-12-26
Error: “version” is a required argument while building


besrabasant:
I am using **rollup** to bundle my project.

At the start, everything was working fine, but I do not know what I changed in my config, I started getting this error on running `rollup -c`.
```

[!] Error: "version" is a required argument.
Error: "version" is a required argument.
    at Object.getArg (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:14625:11)
    at SourceMapConsumer$1.BasicSourceMapConsumer (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:15763:22)
    at new SourceMapConsumer$1 (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:15491:7)
    at Module.getOriginalLocation (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:16925:16)
    at Module.error (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:16942:26)
    at CallExpression.bindNode (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:12326:17)
    at CallExpression.bind (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:11415:8)
    at eachChild.child (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:11433:34)
    at keys.forEach.key (/home/programmersedge/.nvm/versions/node/v6.11.1/lib/node_modules/rollup/dist/rollup.js:11450:5)
    at Array.forEach (native)
```
Here is my `rollup.config.js`
```

import resolve      from 'rollup-plugin-node-resolve'
import babel        from 'rollup-plugin-babel'
import filesize     from 'rollup-plugin-filesize'
import typescript   from 'rollup-plugin-typescript2'
import commonjs     from 'rollup-plugin-commonjs'
import postcss      from 'rollup-plugin-postcss-modules'
import autoprefixer from 'autoprefixer'
import sass         from "node-sass"

const preprocessor = (content, id) => new Promise((resolve, reject) => {
    sass.render({
        file: id,
        sourceMap: "string",
        sourceComments: true,
[...]

PR #2012, opened 2018-02-25
Catch errors when following source maps in error reporting

adrianheine:
See #1817. Apparently this is a common problem.


TomasHubelbauer:
I can't review this but I can confirm it solved my issue with #1817.

PR #2035, opened 2018-03-05
Enable watcher in code-splitting mode

### LINKED ISSUE #1950: Watcher for experimentalCodeSplitting with several inputs and output dir
Hi, 
I'm very exited with new code splitting feature and want use in my project but one thing is stoping me - absents of support for watching code with several inputs and output to dir. 

Do you have any plans to implement this feature?

anandthakker:
Closes #1950 

- [x] Update typing for `rollup.rollup()` to reflect code-splitting output type.
- [x] Update watcher to handle code-splitting output.
- [x] Add tests for watcher in code-splitting mode.
- [x] Lint

### DIFF
--- a/src/watch/index.ts
+++ b/src/watch/index.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import { EventEmitter } from 'events';
 import createFilter from 'rollup-pluginutils/src/createFilter.js';
-import rollup, { CachedChunkSet, InputOptions, OutputOptions, OutputChunk, OutputChunkSet } from '../rollup/index';
+import rollup, { CachedChunk, CachedChunkSet, InputOptions, OutputOptions, OutputChunk, OutputChunkSet } from '../rollup/index';
 import ensureArray from '../utils/ensureArray';
 import { mapSequence } from '../utils/promise';
 import { addTask, deleteTask } from './fileWatchers';
@@ -100,7 +100,7 @@ export class Task {
 	closed: boolean;
 	watched: Set<string>;
 	inputOptions: InputOptions;
-	cache: CachedChunkSet;
+	cache: CachedChunk | CachedChunkSet;
 
 	chokidarOptions: WatchOptions;
 	chokidarOptionsHash: string;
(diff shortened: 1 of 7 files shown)


anandthakker:
@guybedford @lukastaegert I believe this should be ready for a 👀

src/watch/index.ts
```diff
@@ -197,25 +192,35 @@ export class Task {
...
-				this.watched = watched;
+				if ((<OutputChunkSet>result).chunks) {
+					const chunks = (<OutputChunkSet>result).chunks;
+					this.cache = { chunks };
+					for (const chunkName in chunks) {
+						watchChunk(chunks[chunkName]);
+					}
+				} else {
+					const chunk = (<OutputChunk>result);
+					this.cache = { chunks: { [this.inputOptions.entry]: chunk } };
```
guybedford: If the cache consists of `chunks` in both cases, do we still need the `this.cache.modules` check in https://github.com/rollup/rollup/pull/2035/files#diff-7c1e6304262c178e7fe268b62060e26bR85 ?

anandthakker: Ah, good question.  I should have explained this in the code.

`Task#chunk` is typed as `{chunks: { [name]: { modules } } }` even in the single-bundle case just to keep [this code](https://github.com/rollup/rollup/pull/2035/files#diff-093a6241a0003efb710688c0b3332f53R237) slightly simpler.  

However, for the type of [`InputOptions.cache`](https://github.com/rollup/rollup/pull/2035/files#diff-c5fd8e65630bd641bf07130346ee8778R66), which is part of the public JS API, I went with `{ modules } | {chunks: { [name]: { modules } } }` so that this wouldn't be a breaking change (and because I think the existing `cache` option, wherein a developer can simply pass the previous bundle result back in as `cache:`, seems to make sense).

anandthakker: > Task#cache is typed as {chunks: { [name]: { modules } } } even in the single-bundle case just to keep this code slightly simpler.

On second thought, I think it's probably clearer and easier to understand in the future if we just keep `Task#cache` typed identically to `InputOptions#cache`.  I'll push that change.

anandthakker:
@guybedford ah, whoops I didn't realize I'd gotten behind master.  I'll rebase & resolve now.

anandthakker:
Updated `Task#cache` typing and rebased onto master

anandthakker:
@guybedford @lukastaegert anything else you need here before this can be merged?

PR #2080, opened 2018-03-21
Generate an interop default block for external modules on iife/umd

rdarder:
Currently, when options.interop == true (default) and an external
module is imported, cjs generates an interop block that looks
like this:

```js
  function _interopDefault (ex) { return (ex && (typeof ex === 'object')
&& 'default' in ex) ? ex['default'] : ex; }
```

for iife and umd, there's a simliar block generated by `getInteropBlock()`
that looks like this:

```js
  var foo__default = 'default' in foo ? foo['default'] : foo;
```

but for external modules that export as a namespace, it's generated as:

```js
  var foo__default = ${name}['default'];
```
(I don't know why this is more appropriate for namespace exports, but
it's not the issue I'm facing)

Now, when an external module is traced (ExternalModule.traceExports)
to have both namespace and named imports, the namespace case takes
priority.

Changed to generate the latter expression whenever named imports are
found, regardless of namespace imports.

### DIFF
--- a/src/finalisers/shared/getInteropBlock.ts
+++ b/src/finalisers/shared/getInteropBlock.ts
@@ -10,10 +10,10 @@ export default function getInteropBlock(
 		.map(({ name, exportsNamespace, exportsNames, exportsDefault }) => {
 			if (!exportsDefault || options.interop === false) return null;
 
-			if (exportsNamespace) return `${varOrConst} ${name}__default = ${name}['default'];`;
-
 			if (exportsNames)
 				return `${varOrConst} ${name}__default = 'default' in ${name} ? ${name}['default'] : ${name};`;
+
+			if (exportsNamespace) return `${varOrConst} ${name}__default = ${name}['default'];`;
 
 			return `${name} = ${name} && ${name}.hasOwnProperty('default') ? ${name}['default'] : ${name};`;
 		})
--- a/test/form/samples/import-external-namespace-and-default/_expected/amd.js
+++ b/test/form/samples/import-external-namespace-and-default/_expected/amd.js
@@ -1,6 +1,6 @@
 define(['foo'], function (foo) { 'use strict';
 
-	var foo__default = foo['default'];
+	var foo__default = 'default' in foo ? foo['default'] : foo;
 
 	console.log( foo.bar );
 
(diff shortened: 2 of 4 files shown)


guybedford:
I can definitely appreciate the expected convenience here.

You say in the example that the case is exporting the default and named exports, but the code change you describe refers to the namespace export case? Or does it just happen that in your own code you use the namespace form to trigger this path?

If so, it sounds like the proposed convenience is a bit inconsistent then if it only applies when there is a namespace exported? Perhaps it would be a little more consistent to remove the `exportsNamespace` branch entirely here?

rdarder:
In the test case, ExternalModule.exportsNames *and* ExternalModule.exportsNamespace are both true. I got puzzled by that, but it seems that the trackExport gets called with 'bar' when it parses `other.js` accessing foo.bar (sorry for the tangled explanation, I'm not that familiar with rollup).
It may be the case that that tracking is not working properly, I don't know the exact semantics of exportsNames (although it seems it should be false in this case, right?)
This is the log I get by running that test and putting console.log(name) in ExternalModule.trackExports(name)
```
      import-external-namespace-and-default
*
default
*
bar  <=========
*
default
        ✓ generates amd
        ✓ generates cjs
        ✓ generates system
        ✓ generates es
        ✓ generates iife
        ✓ generates umd

```

Other than that, I completely agree that we should remove the namespace branch altogether, as it seems there's no gain in removing the fallback just for that case.
Moreover, perhaps we should aim to emit the same js code that the cjs finalizer emits ? (checks that the module is an object before checking if `default in mod? ...`
Let me know your thoughts and I will update the PR accordingly. Thanks!

guybedford:
Let's see what @lukastaegert and @Rich-Harris think about removing the namespace branch here of `var mod__default = mod['default']` for the more flexible `var mod_default = 'default' in mod ? mod['default'] : mod` case. We need to make sure we're on the right side of the lines with flexibility and consistency here, but I'm under the impression that Rollup strives to do the right thing as much as possible to make these scenarios easy where it can so this seems to align with that as far as I can tell.

> Moreover, perhaps we should aim to emit the same js code that the cjs finalizer emits ? (checks that the module is an object before checking if default in mod? ...

This is handled in the final fallthrough, with the optimization of these branches being that we know we don't need to do this check if things like `mod.bar` are assumed to exist.

guybedford:
I created #2136 as a follow-up PR here. Let's continue discussion there.

PR #2089, opened 2018-03-26
Allow customizing the folder/file layout of experimentalPreserveModules

kellyselden:
Now that you can keep the folder layout of your project with `experimentalPreserveModules`, I believe some additional configuration is necessary. When using the node-resolve plugin, you can leave your project root with the `opts.basedir ` property. When the common ancestor is found to determine the new folder layout, it can get pretty gnarly.

Given `process.cwd() = '.'`.

Source:

```
- path
  - to
    - node_modules
      - dep
        - index.js
- path
  - to
    - project
      - index.js
```

Desired:

```
- node_modules
  - dep
    - index.js
- index.js
```

with exposing the existing option `inputRelativeDir = 'path/to/project` and a `moduleDest` hook that lets you override the node_modules location.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -17,6 +17,7 @@ import {
 	CachedChunkSet,
 	InputOptions,
 	IsExternalHook,
+	ModuleDestHook,
 	Plugin,
 	ResolveIdHook,
 	RollupWarning,
@@ -61,7 +62,11 @@ function generateChunkName(
 	let uniqueIndex = startAtTwo ? 2 : 1;
 	while (chunkNames[uniqueName]) uniqueName = name + uniqueIndex++;
 	chunkNames[uniqueName] = true;
-	return uniqueName + ext;
+	name = uniqueName + ext;
+	if (!name.startsWith('.')) {
+		name = './' + name;
+	}
+	return name;
 }
 
 export default class Graph {
@@ -83,6 +88,7 @@ export default class Graph {
 		importerStart?: number
 	) => void;
 	moduleById: Map<string, Module | ExternalModule>;
+	moduleDest: ModuleDestHook;
 	modules: Module[];
 	onwarn: WarningHandler;
 	plugins: Plugin[];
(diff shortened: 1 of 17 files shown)


guybedford:
Does it need to be a hook here over a string-based configuration option? Otherwise would something like `preserveModulesBase` work? I don't think we need the input base for non-preserve modules cases, so a specific property seems to make sense.

kellyselden:
I might be able to get by with an object lookup:

```
moduleDest: {
  '../../../path/to/node_modules1': './node_modules1',
  '../../../path/to/node_modules2': './node_modules2'
}
```

`moduleDest` is like the `paths` hook, except file location instead of import string. `preserveModulesBase` boolean wouldn't be enough info. I don't think you can correctly guess the right base folder the user wants to use.

Both of these properties only apply to `experimentalPreserveModules`, so they could be renamed as such. They could also be submitted independently if you prefer, as I think they are useful on their own.

It may also help to see an example:
https://github.com/kellyselden/ember-cli/blob/d64d9660aeca063af6694611be93a981253e148a/lib/broccoli/compile.js#L347-L353
https://github.com/kellyselden/ember-cli/blob/d64d9660aeca063af6694611be93a981253e148a/lib/broccoli/compile.js#L364

guybedford:
@kellyselden is this an issue because of symlinks? If so does `preserveSymlinks` help here at all?

kellyselden:
> Another option might be to treat anything outside of the preserve modules base folder as external and ignored by default?

These node_modules are being included in the assets sent to browser, so treating as external won't work.

guybedford:
Another way to achieve this might be to have a custom load and resolveId function, where the resolveId function applies the moduleDest after resolution, and the load function loads the file with the moduleDest reverse applied again.

I think that would provide a possible route for supporting this within a third-party plugin, let me know if that might work for you?

shellscape:
@kellyselden hey there. Thanks for all of your work on this PR so far. It looks like it needs to be rebased to the current master to keep it going. Would you be up for that, or would you prefer to close this one?

PR #2109, opened 2018-04-02
fix: inaccurate CommonJS env sniffing

zheeeng:
Inspired by the [underscore.js UMD sniffing](https://github.com/jashkenas/underscore/blob/master/underscore.js#L53).
Consider in browser env, there are two DOM nodes which have the id 'module' and 'exports', the runtime env will be detected as commonjs.


src/finalisers/umd.ts
```diff
@@ -96,7 +96,7 @@ export default function umd(
 	}
 
 	const wrapperIntro = `(function (global, factory) {
-			typeof exports === 'object' && typeof module !== 'undefined' ? ${cjsExport}factory(${cjsDeps.join(
+			typeof exports === 'object' && typeof module !== 'undefined' && !module.nodeType ? ${cjsExport}factory(${cjsDeps.join(
```
guybedford: An alternative here might just be `typeof module !== 'undefined' && typeof module.exports === 'object'`... better detection of a feature is always preferable to ruling out negative cases.

zheeeng: Your suggestion is awesome. After searching a while, I found that UMD proposal has provided us a shorter [choice](https://github.com/umdjs/umd/blob/master/templates/returnExports.js#L21):

```
typeof module === 'object' && module.exports
```

zheeeng:
@guybedford May I ask is there any problem with the present IIFE wrapping way?

shellscape:
@zheeeng hey there. Thanks for all of your work on this PR so far. It looks like it needs to be rebased to the current master to keep it going. Would you be up for that, or would you prefer to close this one?

PR #2118, opened 2018-04-11
handle missing exports in chunked mode

kellyselden:
Revived from #1986

From https://github.com/rollup/rollup/pull/1986#issuecomment-372232259

@guybedford Turns out I only needed to do # 1, and I got the desired outcome.

> Perhaps we could support the handleMissingExport hook returning true or false to determine whether to create a variable or to throw?

I added that in a previous feature, which is partly why this feature is so small!


src/Chunk.ts
```diff
@@ -594,6 +594,10 @@ export default class Chunk {
 			} else {
 				const variable = this.exportNames[exportName];
 				const module = this.exports.get(variable);
+
+				// skip shimmed exports
+				if (!module) continue;
```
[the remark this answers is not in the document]
kellyselden: Ok I'll take another stab at it.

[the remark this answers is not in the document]
guybedford: Thanks, will be a great help to ensure the assumptions are minimal.

test/chunking-form/samples/missing-export/_config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+	description: 'missing export',
+	options: {
+		input: ['main.js', 'dep2.js'],
+		plugins: [
+			{
+				missingExport() {
+					return true;
+				}
```
guybedford: Rather than have the missingExport hook provide this functionality, how about making this a flag option - `shimMissingExports: true`?

kellyselden: That is already the way to stop the default behavior of hard erroring (in non-code-splitting mode). Are you suggesting checking [here](https://github.com/rollup/rollup/blob/v0.57.1/src/utils/defaults.ts#L76) for `shimMissingExports` and conditionally not erroring?

guybedford: Yes exactly, so that this can be done without a plugin.

kellyselden: done

kellyselden:
Ready for review again. It turned out pretty clean IMO. While it is more encapsulated, I had to add a little bit more compatibility code to accommodate some bad assumptions, like `NamespaceVariable.originals` and `traceExport.isExportAllSearch`.

src/Module.ts
```diff
@@ -708,4 +725,13 @@ export default class Module {
 		warning.id = this.id;
 		this.graph.warn(warning);
 	}
+
+	shimMissingExport(name: string) {
+		const shim = new ShimVariable(name === 'default' ? this.basename() : name);
+		this.exports[name] = {
+			localName: name,
+			shim
+		};
```
guybedford: Would it be possible for `ShimVariable` to inject itself into `originals` here on the module itself so that variable lookups play out correctly? Or does that introduce too many cross-cutting issues?

src/ExternalModule.ts
```diff
@@ -99,7 +99,12 @@ export default class ExternalModule {
 		});
 	}
 
-	traceExport(name: string): Variable {
+	traceExport(
+		name: string,
+		_options?: {
+			isExportAllSearch?: boolean;
+		}
```
guybedford: Could we make this a default flag over an object rather?

src/finalisers/esm.ts
```diff
@@ -83,7 +83,13 @@ export default function esm(
...
 	const exportDeclaration: string[] = [];
 	exports.forEach(specifier => {
-		if (specifier.exported === 'default') {
+		if (specifier.shim) {
+			if (specifier.exported === 'default') {
+				exportBlock.push(`export ${specifier.exported} null;`);
+			} else {
+				exportBlock.push(`export const ${specifier.exported} = null;`);
+			}
+		} else if (specifier.exported === 'default') {
```
guybedford: I wonder if we could shift these types of logic out of the finalisers by rendering a line like:

```js
export const $$shim = null;
```

as part of the core rendering process, then setting the `specifier.local` value to `$$shim`.

guybedford:
Looking good! Thanks for understanding we just want to keep assumptions in unrelated places to a minimum in the codebase.

src/ExternalModule.ts
kellyselden: I figured `module.traceExport(name, { isExportAllSearch: true });` would be more clear to the reader than, `module.traceExport(name, true);`, but I can change it if you want.

src/Module.ts
kellyselden: I thought about it, but I couldn't figure out how to get access to the `NamespaceVariable.originals` from earlier in the traceExport call stack at this point of shimming.

src/finalisers/esm.ts
kellyselden: I'm not sure I follow. Wouldn't it get really complicated if there were multiple missing exports for a given modules? Like:

```js
import { missing1, missing2 } from './foo';

missing1('bar');
missing2('baz');
```

guybedford: Sorry I should have clarified rendering just a `${varOrConst} $$shim = null;`, so that in `foo` itself the output would be:

```js
const $$shim = null;

export { $$shim as missing1, $$shim as missing 2 }
```

Let me know if that makes sense?

src/ExternalModule.ts
guybedford: That would be great just to avoid the object allocation, for what it's worth. It doesn't seem like the flags will be extending anytime soon.

kellyselden: done

src/finalisers/esm.ts
kellyselden: It does now. But aren't you worried about adding a var after the deconflicting step? Any name we choose, like $$shim, would possibly conflict with an existing $$shim var.

kellyselden: I made the change, but the conflict issue still worries me.

src/finalisers/esm.ts
```diff
@@ -82,6 +82,11 @@ export default function esm(
 
 	const exportBlock: string[] = [];
 	const exportDeclaration: string[] = [];
+	exports.forEach(specifier => {
+		if (specifier.shim) {
+			exportBlock.push(`const ${specifier.local} = null;`);
```
guybedford: This can be done in the `preRender` of `Chunk.ts`.

guybedford: Also the `const` should be a `${varOrConst}` reference.

src/finalisers/esm.ts
guybedford: `$$shim` can be added to the list of names to deshadow in the `setIdentifierResolutions` function of `Chunk` to ensure proper handling of this.

src/Module.ts
guybedford: Let me know if you have further questions on this one.

src/finalisers/esm.ts
guybedford: Just like at https://github.com/rollup/rollup/blob/79ad616596f8c5cebafa62883f4ddd63d922bbb5/src/Chunk.ts#L502, when needing a shim variable you can add it to the deshadowing list this way.

preRender renders the main body source, while render renders the finalizers. I think preRender would be a good place. Note that there is a `hoistedSource` for bindings that must be available early this could be added to at https://github.com/rollup/rollup/blob/79ad616596f8c5cebafa62883f4ddd63d922bbb5/src/Chunk.ts#L778. Just writing directly into that string is fine (if (shimExports) hoistedSource += `${varOrConst} $$shim = null`;). Note that finalizers will apply wrappers so that there are no global leaks to worry about.

src/finalisers/esm.ts
kellyselden: I'm now running into the problem where my output is

```js
const $$shim$1 = null;

var $$shim = null;
export { $$shim$1 as $$shim, $$shim };
```

This is a dependency and an entry point. My existing unused `$$shim` was renamed to `$$shim$1` because of the deshadowing, but is still exported as `$$shim` because of entry point. Now this conflicts with my shimmed export `$$shim`. It's almost like we need to reverse the renaming by leaving existing `$$shim`s as is, and instead do something like `chunk.generateUniqueVarAndExportName()` for the shimmed export name. Am I on the right track?

kellyselden: I got something working.

src/finalisers/esm.ts
kellyselden: If I do this, it works for everything except systemjs, where the output is

```js
System.register([], function (exports, module) {
	'use strict';
	return {
		execute: function () {

			exports('$$shim$2', $$shim$2);
			var $$shim$2 = null;

			const $$shim = exports('$$shim$1', null);

		}
	};
});
```

I may have the change the order of `functionExports` in systemjs to get it to work this way.

kellyselden: made the change.

src/Module.ts
kellyselden: I was able to remove the NamespaceVariable.originals workaround completely. Something I changed must have made it obsolete.

kellyselden:
All comments should be addressed.

kellyselden:
Now passing after #2185

kellyselden:
@guybedford Would you mind re-reviewing this?

guybedford:
@kellyselden this is still on my list - just finding it hard to prioritise right now. Thanks for your patience, and let me know if you have any important milestones / schedule you're hoping to meet with this?

kellyselden:
I can stay forked for the time being. I hope you don't mind reviewing with merge conflicts, as this gets out of date with master often.

src/finalisers/system.ts
```diff
@@ -118,12 +123,18 @@ ${t}${t}execute: function () {
 
 ${functionExports.length ? `${t}${t}${t}` + functionExports.join(`\n${t}${t}${t}`) + '\n' : ''}`;
 
+	const wrapperEnd = `
+${shimExports.length ? `${t}${t}${t}` + shimExports.join(`\n${t}${t}${t}`) + '\n' : ''}
+${t}${t}}
+${t}};
+});`;
+
```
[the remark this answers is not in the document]
kellyselden: done

src/Chunk.ts
```diff
@@ -47,6 +48,7 @@ export type ChunkExports = {
 	local: string;
 	exported: string;
 	hoisted: boolean;
+	shim?: boolean;
```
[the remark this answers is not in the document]
kellyselden: done

src/Chunk.ts
```diff
@@ -534,6 +543,33 @@ export default class Chunk {
...
+					if (this.graph.shimMissingExports) {
+						// so a shimmed export doesn't collide with a `foo as $$shim` export
+						used[exportName] = 1;
+					}
+
+					newExportName = exportName;
+				}
+				exportVariable.exportName = newExportName;
+			}
+
```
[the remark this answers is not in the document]
kellyselden: Sounds good. I will submit as a separate PR so this PR doesn't get conflicted. #2255

guybedford:
@kellyselden seeing what I can do on this this evening - do you mind if I push some updates to this branch?

kellyselden:
I don't mind.

guybedford:
Ok sure, I've moved the shim variable to being treated as a global variable, with locals then deduped if they collide with that.

I've also added some deprecation for the missingExportHook, if you think this solves the use case and that might be possible now.

There is some odd duplication of the variables going on in the output of missing-export now, but it still seems to be functionally correct in that it has the right variable reference but not necessarily by consistent names:

```js
_shimmedExport();
_shimmedExport();
_shimmedExport$1();
_shimmedExport();
```

Perhaps try this out in your integrations there and let me know if this has any scoping issues you come across further?

guybedford:
I can confirm with the fix at https://github.com/rollup/rollup/pull/2258, the System case is now handled properly as well, without any further changes here.

guybedford:
@kellyselden I'd like to aim to get this in for the 1.0 release, please do let me know if the changes here work for you.

guybedford:
In that case @lukastaegert it would be great if you could offer a final review here.

lukastaegert:
Sure thing. It's a little late today for a full review today but I added it to the top of the review queue of the [1.0.0 project](https://github.com/rollup/rollup/projects/1) and will get to it first thing tomorrow

src/Module.ts
```diff
@@ -753,4 +753,13 @@ export default class Module {
...
 		this.graph.warn(warning);
 	}
+
+	shimMissingExport(name: string) {
+		// could have already been generated
+		if (!this.exports[name])
+			this.exports[name] = {
+				localName: '_shimmedExport'
+			};
+		return this.graph.scope.findVariable('_shimmedExport');
```
lukastaegert: Instead of searching we might just use `this.graph.exportShimVariable` here which would also be a little clearer.

kellyselden:
I can update the tests to be more clear. As for the rest of the changes, maybe @guybedford has some ideas?

guybedford:
@kellyselden that would be great if you can make the test adjustments. Will aim to look into the output fixes over the weekend.

guybedford:
@lukastaegert let me know if you're still working on this, otherwise I can look into it then as well.

lukastaegert:
Not at the moment, got a little side-tracked by the release issues. My current impression so far is that we should *not* rely on name-matching to detect if an export is shimmed but maybe via a flag added to the export description. Furthermore, we should not use `findVariable` to get a reference of the shimmed variable instance (which may be shadowed) but directly retrieve it from the graph. Also if we want to rely on the shim variable having a specific name in the end, we need to make sure that no variable on module level can have this name.

But I do not have any results yet and will not find any time to look into this further before tomorrow so please go ahead if you have time today.

guybedford:
@lukastaegert it appears you're right there's no clean way to do this at the variable level due to the fact that we dedupe based on name, not variable name (two identifiers of the same name in the same scope cant reference different underlying variables in the deshadowing logic)

Given this, the approach @kellyselden originally took would work here.

My major concern for this feature from the start was to find an approach that would use the existing tracing logic and not pollute it with conditional exceptions. This is non-spec behaviour, and I'm not sure how useful it is to the wider user base to make an already complex part of the codebase more complex.

Rather I'd prefer if we can use a unique identifier name that would fail deshadowing if attempted to be attacked like you have done, as opposed to doing it the "right" way that has all the branch pollution.

To be perfectly honest, given the choice between polluted trace branches and supporting this feature, I'd rather not implement this feature at all if that is the case.

I've updated the variable name to `_missingExportShim`, and integrated the test updates and feedback for now.

src/Module.ts
```diff
@@ -757,4 +758,19 @@ export default class Module {
 		warning.id = this.id;
 		this.graph.warn(warning);
 	}
+
+	shimMissingExport(name: string) {
+		// could have already been generated
+		if (!this.exports[name])
+			this.graph.warn({
+				message: `Export "${name}" has been shimmed in module ${relativeId(this.id)}.`,
```
guybedford: Perhaps clarify `Missing export` here.

src/Module.ts
```diff
@@ -761,9 +762,15 @@ export default class Module {
 	shimMissingExport(name: string) {
 		// could have already been generated
 		if (!this.exports[name])
-			this.exports[name] = {
-				localName: '_missingExportShim'
-			};
+			this.graph.warn({
+				message: `Export "${name}" has been shimmed in module ${relativeId(this.id)}.`,
+				code: 'SHIMMED_EXPORT',
+				missing: name,
```
guybedford: `missingExportName` or `exportName` perhaps?

kellyselden:
This is performing as intended in my test, so all good on my end. I'd like to reiterate that this would be a temporary, stop-gap solution just to get the ecosystem working. I'm all for huge console warnings and deprecating for Rollup v2.

src/Module.ts
lukastaegert: done

src/Module.ts
lukastaegert: done

kellyselden:
woohoo!

PR #2147, opened 2018-04-21
unify MISSING_EXPORT calls

kellyselden:
Should help the logic mentioned in https://github.com/rollup/rollup/pull/2118#discussion_r181550594


src/Module.ts
```diff
@@ -247,6 +254,26 @@ export default class Module {
 
 		timeStart('analyse ast', 3);
 
+		const module = this;
+
+		function handleMissingExport(
+			this: AstContext,
```
guybedford: This exposes a huge surface area - it would help to know exactly what functionality you're looking to get access to here? Also the PR at https://github.com/rollup/rollup/pull/2126 is aiming to unify this plugin context between plugin hooks on a single context API, instead of having a different context for each hook, so this would clash with that.

kellyselden: I changed it to no longer depend on the whole AstContext. It now takes in references to the warn and error functions. Hopefully that helps a little bit.

PR #2148, opened 2018-04-21
fix missingExport call order

kellyselden:


guybedford:
Out of interest - if we had the `shimMissingExports` option would you still need this hook? Or could it just be a normal warning then?

kellyselden:
Potentially. I'm using it to tweak the message to the console (also to prevent the throw). If I no longer needed to prevent the throw, I could see myself using the default warning message and no longer need the hook.

guybedford:
@kellyselden since it isn't yet a documented hook I'd really prefer that route if we can take it.

PR #2162, opened 2018-04-27
Show errors for every build in multiconfig

TrySound:
Currently if one build fails then the next one is swallowed.

I'm working on size-snapshot plugin which fails if size snapshot is
not matched. The problem is that with umd, esm and cjs configs we will
see only umd error which may confuse why there is not error for another
builds.

My solution is to lift `process.exit(1)`.

### DIFF
--- a/bin/src/run/index.ts
+++ b/bin/src/run/index.ts
@@ -97,7 +97,7 @@ export default function runRollup(command: any) {
 			.then(configs => execute(configFile, configs, command))
 			.catch(handleError);
 	} else {
-		return execute(configFile, <any>[{ input: null }], command);
+		execute(configFile, <any>[{ input: null }], command);
 	}
 }
 
@@ -126,6 +126,10 @@ function execute(configFile: string, configs: InputOptions[], command: any) {
 			if (optionError) inputOptions.onwarn({ code: 'UNKNOWN_OPTION', message: optionError });
 
 			return build(inputOptions, outputOptions, warnings, command.silent);
+		}).then(results => {
+			if (!results.every(Boolean)) {
+				process.exit(1);
+			}
 		});
 	}
 }
--- a/bin/src/run/build.ts
+++ b/bin/src/run/build.ts
@@ -82,6 +82,10 @@ export default function build(
 			if (bundle && bundle.getTimings) {
 				printTimings(bundle.getTimings());
 			}
+			return true;
 		})
-		.catch(handleError);
+		.catch((error: any) => {
+			handleError(error, true);
+			return false;
+		});
 }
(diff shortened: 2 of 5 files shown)


guybedford:
You don't think we should just fail on the first error?

guybedford:
(as in stop the build entirely)

guybedford:
Btw I've also proposed `onbuildstart` and `onbuildend` plugin hooks in https://github.com/rollup/rollup/pull/2126 which may be related to your use case.

TrySound:
This test have similar meaning.
https://github.com/rollup/rollup/blob/master/test/cli/samples/multiple-configs/_config.js

I think it's important to have parallel error reporting across bundles for better DX. It's the same like linting: if you could see only one error at once you would spend to much time to fix them.

Same is actually fits to the current issue. Without this change you need to run build a couple of times to see all errors which can detect rollup plugins in different configs.

guybedford:
A task or execution stops on first error so that's why I'm just wondering if it might make more sense not to continue. Linting is a little different in each check being independent, but here the output of one bundle might affect another.

TrySound:
@guybedford Can you explain how builds are related to each other? I didn't find anything reused across rollup calls.

guybedford:
I just mean that the output file of one build could be used as an input in a subsequent build. Although I doubt it's done very commonly, but still from a correctness point of view, just wondering what the benefits are of not stopping - processes that fail earlier can be fixed faster possibly.

TrySound:
I'm not sure I understand what you mean. Builds in multiconfig are completely separated
```js
export default [
  {
    input: './index1.js',
    plugins: [
      eslint()
    ]
  },
  {
    input: './index2.js',
    plugins: [
      eslint()
    ]
  }
]
```

I want to get all eslint errors for both builds and fix them at once without fixing one and then running build and then fixing again which may take a lot of time. Bundling can take a lot of seconds or even minutes and iteration should be as small as possible.

guybedford:
I'm just saying that they are tasks, and tasks can implicitly depend on eachother. Through the file system or through memory.

Generally in programming, executable tasks fail on the first one, with a single error. So I'm just wondering why we're doing things differently here.

For example, a plugin error would repeat three times.

Predictability is as important as fast iteration. At current pace of field with assumptions that things like hot reloading should just work it's easy to forget predictability.

TrySound:
Okay. I'm convinced.

bin/src/run/index.ts
```diff
@@ -126,6 +126,10 @@ function execute(configFile: string, configs: InputOptions[], command: any) {
 			if (optionError) inputOptions.onwarn({ code: 'UNKNOWN_OPTION', message: optionError });
 
 			return build(inputOptions, outputOptions, warnings, command.silent);
+		}).then(results => {
+			if (!results.every(Boolean)) {
+				process.exit(1);
```
guybedford: This still seems useful?

guybedford: Ah, it seems like this was just needed because errors were no longer propagated.

TrySound: I don't think so. It's error batching.

PR #2185, opened 2018-05-12
use npm 5 in lower node versions

kellyselden:
master has regressed in lower node versions https://travis-ci.org/rollup/rollup/jobs/378107881

```
Error: 'crypto' is imported by node_modules/math-random/node.js, but could not be resolved – treating it as an external dependency
```

This is because the lower envs don't use npm 5 and version drift occurred.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,5 @@ env:
     - BUILD_TIMEOUT=10000
 install: npm install --ignore-scripts
 before_install:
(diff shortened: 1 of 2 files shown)


phated:
This is actually still an issue if someone building the library has forced package-lock=false (like me).  It seems that randomatic did a breaking patch update - I'm trying to get them to fix it right now.

In the meantime, you might want to add a `.npmrc` with `package-lock=true` to the project.

PR #2236, opened 2018-06-01
better test case for preserve-modules-reaching-outside

kellyselden:
it will be a better test if commondir is different than `process.cwd()`.


guybedford:
Good idea.

PR #2258, opened 2018-06-10
System uninitialized bindings cases

guybedford:
This fixes https://github.com/rollup/rollup/issues/2257, providing exports both for uninitialized variables and exported globals.

When there is just a single export, an `exports('name', binding)` is emitted, and when there are multiple, the object form of export is used instead.


guybedford:
I've updated this PR with the fix using the `UNDEFINED_EXPRESSION` handling in variables.

Issue #1203, opened 2016-12-30
Plugin API for declaring dependencies programmatically


Rich-Harris:
Via https://github.com/rollup/rollup-watch/issues/23 (cc @droooney, @tivac). Currently, the only way for a file to declare a dependency on another file is for it to be convert to JavaScript with an `import` statement. That means that the only way to handle things like `@import` statements in CSS or compile-to-CSS languages is to convert them into JavaScript somehow, which sounds a bit awkward and haphazard.

Normally this isn't a problem since, if you create a CSS file by following `@import` statements and gluing multiple CSS files together, the build works as expected. But if you're using rollup-watch, it means that changes to the imported files don't have any effect.

If there was a way for plugins to declare a file's dependencies programmatically – in other words, if this...

```css
/* styles/main.css */
@import './layout.css';
@import './typography.css';
```

```js
// app.js
import './styles/main.css';
```

...caused `styles/layout.css` and `styles/typography.css` to be registered as dependencies of `styles/main.css`, as well as `styles/main.css` being registered as a dependency of `app.js` due to the `import` declaration – that problem would be solved.

Proposed API:

```js
export default function myPlugin ( options = {} ) {
  return {
    name: 'my-plugin',

    transform ( code, id ) {
      // some code happens...

      return {
        code: 'SOME_CODE',
        map: { version: 3, ... },
        dependencies: [
          // paths subject to usual resolution logic, so can be
          // relative or absolute, or you can supply your own
          // plugin resolve hook if needs be
          './typography.css',
          '/path/to/project/styles/layout.css'
        ]
      };
    }
  };
}
```

Note that those dependencies would be in *addition* to any discovered by JavaScript `import` declarations, and would *not* be loaded by Rollup – they would just be added to the dependency graph metadata at the end of the bundling process so that rollup-watch would be able to add file watchers.

tivac:
This is fine, but I think documentation on returning `import` statements would be fine too. It took me a bit to get there but either it's not a big deal or Stockholm syndrome kicked in and I don't mind the current paradigm at all. It does exactly what you'd want and leaves no trace after treeshaking.

nathancahill:
I'd love to see this.

aMarCruz:
@Rich-Harris  any progress on this?
rollup-plugin-pug inserts the "import" for included files, but this has no effect in dependency detection.

Example:
```pug
//- main.pug
include ./inner.pug
p Hello
```

compiles to:
```js
import pug from "pug-runtime";
[...]

PR #2259, opened 2018-06-11
Implement transform dependencies

### LINKED ISSUE #1634: Can a plugin trigger a new build when watch is enabled?
When rollup has been started with the `watch` option enabled, is it possible for a plugin to - after the initial build - arbitrarily decide that it would like to trigger / emit an `update` event that would cause rollup to regenerate the bundle?

My use case:

I've created a plugin - [rollup-plugin-import-globs](https://github.com/tkambler/rollup-plugin-import-globs) - that allows me to do this:

```
import './components/*';
```

This works great. There's only one hiccup - if I create a *new* file within this folder or delete an existing one, the watch doesn't seem to want to notice it. I had thought that it would be nice to be able to put my *own* watch in place within my plugin (based on the glob pattern passed) and then trigger my own "change" event later on.

Is this doable?

guybedford:
This implements transform dependencies, as discussed in #1203, allowing plugin transforms to express dependencies on other files which will act to invalidate the whole transform in the cache.

Also resolves #1634, by setting a transform dependency to a directory.


tivac:
Tried this out in [`tivac/modular-css#transform-dependencies`](https://github.com/tivac/modular-css/tree/transform-dependencies).

They're being watched for changes, which is great!

Unfortunately it looks like the `dependencies` returned by the `transform` hook aren't persisted into the `bundle` param of `generateBundle`, so I run into the problem I had before where `generateBundle()` doesn't know about the totality of the files in a bundle. This leads to missing CSS output 😞

tivac:
Also I ran into a separate issue w/ the changes to `emitAsset()` and `getAssetFileName()` where I've now got a race condition:

I need to know the name of the file to generate the source, but I can't get the name of the file until it's got source.

😖

guybedford:
@tivac thanks for trying it out!

> Unfortunately it looks like the dependencies returned by the transform hook aren't persisted into the bundle param of generateBundle, so I run into the problem I had before where generateBundle() doesn't know about the totality of the files in a bundle. This leads to missing CSS output 😞

These dependencies are only for the purposes of the watcher, and hold no further meaning. You'll need to include them in the CSS output manually yourself. Let me know if that makes sense?

> I need to know the name of the file to generate the source, but I can't get the name of the file until it's got source.

The suggestion here is to generate the sources up the CSS dependency tree in order. Circular references should work out given that there is still an underlying order so that they can be removed from the tree first.

tivac:
> These dependencies are only for the purposes of the watcher, and hold no further meaning. You'll need to include them in the CSS output manually yourself.

Oh, that's a bummer but ok. I have the deposit graph of all CSS files and can do the work to track shared deps across bundles.

> The suggestion here is to generate the sources up the CSS dependency tree in order. Circular references should work out given that there is still an underlying order so that they can be removed from the tree first.

That doesn't seem like it'll help this issue, but I'll move that piece into a separate issue if I can't solve it myself.

tivac:
Ok got `modular-css-rollup` fully working with this. Seems to work as advertised, had to do some extra work on my side to shift the paradigm from injecting `import` statements to using the `dependencies` array and the resulting changes in how the plugin is informed of updates & has to output content. Nothing major and the output stuff needed to happen regardless.

guybedford:
@tivac great to hear!

tivac:
To ensure that source-maps are correct I really need to be able to know the name of an asset before I can generate its source. Is that a hard requirement with this API or can that be adjusted somehow?

Currently this tosses an error for me:

```js
const css = this.emitAsset("foo.css");

console.log(
    this.getAssetFileName(css)
);

// Throws:  Plugin error - Unable to get asset file name for asset 86f175e2. Ensure that the source is set and that generate is called first.
```

src/watch/fileWatchers.ts
```diff
@@ -49,14 +53,17 @@ export default class FileWatcher {
...
 	fsWatcher: FSWatcher | fs.FSWatcher;
 	tasks: Set<Task>;
+	transformDependencyTasks: Set<Task>;
 
 	constructor(id: string, chokidarOptions: WatchOptions, dispose: () => void) {
 		this.tasks = new Set();
+		this.transformDependencyTasks = new Set();
 
-		let data: string;
+		let mtime = -1;
```
lukastaegert: I know this corresponds to the key of the fs.statSync response but I think calling this `modificationTime` would make the code below a little more readable to people not that familiar with this function.

src/watch/fileWatchers.ts
```diff
@@ -72,15 +79,24 @@ export default class FileWatcher {
...
+						if (mtime !== -1) {
+							mtime = -1;
+							this.trigger(id);
+						}
+						return;
+					}
+					throw err;
 				}
+				// debounce
+				if (+stats.mtime - mtime > 50) this.trigger(id);
```
lukastaegert: Could you explain in  a few words the idea here? So we ignore changes within 50ms of the previous change? Is this for the duplicate events mentioned in the replaced code or this is a certain scenario we try to solve here?

src/watch/fileWatchers.ts
```diff
@@ -28,17 +30,19 @@ export function addTask(
 		}
 	}
 
-	group.get(id).tasks.add(task);
+	if (isTransformDependency) watcher.transformDependencyTasks.add(task);
+	else watcher.tasks.add(task);
```
lukastaegert: Wouldn't it be nicer to not reach into the internals of the watcher but add a method to the watcher `addTask(task, isTransformDependency)` and let the watcher decide how the type of dependency is handled itself and make `tasks`private?

src/watch/fileWatchers.ts
lukastaegert: In any case I must admit the new logic looks definitely more reliable 👍

src/watch/fileWatchers.ts
lukastaegert: Also it looks like if a dependency is both an transform and a normal dependency, it will be watched twice.

src/rollup/types.d.ts
```diff
@@ -56,6 +56,7 @@ export interface SourceDescription {
 export interface ModuleJSON {
 	id: string;
 	dependencies: string[];
+	transformDependencies: string[];
```
lukastaegert: Even though these dependencies currently originate only from `transform` hooks, this is not really important to any later handling. Most notably, I could imagine the `load` hook declare dependencies as well (i.e. we load a .less file and the loader declares a bunch of dependencies right away).

Maybe we could call it something like `manualDependencies` or `additionalDependencies` or `nonModuleDependencies` or something like that.

Also, as stated, it might make sense to add this to the load hook as well (unless I miss something here).

test/watch/index.js
```diff
@@ -263,6 +263,91 @@ describe('rollup.watch', () => {
 				});
 		});
 
+		it('watches and rebuilds transform dependencies', () => {
```
lukastaegert: Nice test! I would love to see a few more edge cases handled in this or a second test:
* It works correctly if a dependency is both a real and a manually declared dependency
* I also works if we add a dependency on the second run but not on the first
* what happens if a dependency is missing or deleted? If it is still required, I guess we should throw an error.
* if after a change, the hook no longer requires a dependency and this dependency is deleted at the same time, no error should occur

src/watch/fileWatchers.ts
guybedford: So when watchers emit file events at the same time (say save happens twice, or gets triggered as two events somehow), then you don't want to retrigger a whole rebuild. The debouncing approach just ignores events happening within 50ms of eachother, instead of having to read the whole source to determine. This then will also work with folder events too.

src/rollup/types.d.ts
guybedford: It is important for later handling due to the fact that they invalidate the module cache for their parent transform ids.

I think it's a simpler conceptual model to think in terms of transform dependencies invalidating that module. If there are dependencies of the build that don't invalidate any modules they are probably better emitted as assets anyway.

The load hook could support a dependencies array definitely, but it feels less clear whether that should invalidate the transform cache or not?

Another option might be to have `this.emitDependency` where the transform hook has its own context already. But my worry with that is that it isn't necessarily clear to trace what is invalidating what. The nice thing about the transform dependencies is that it is a very explicit invalidation chain.

test/watch/index.js
guybedford: Done :)

guybedford:
> To ensure that source-maps are correct I really need to be able to know the name of an asset before I can generate its source. Is that a hard requirement with this API or can that be adjusted somehow?

@tivac can you clarify why you need the assets own name here? This is a hard constraint due to using hashes in file names. If you need to know the name of a dependency of a file that should work out fine with this API though. Hashes in file names really does tie things up (it was a pain in chunking too), but for mostly good reasons.

tivac:
Sorry, for some reason I thought the `emitAsset` bits were in this PR but in retrospect that was silly.

I'm gonna try and not derail this thread further but I'll answer your question @guybedford and shut up until I have more details.

I'm using `postcss` to power `modular-css` there's a param it expects: [`to`](http://api.postcss.org/global.html#processOptions).

> the path where you’ll put the output CSS file. You should always set to to generate correct source maps.

It seems like I'm getting weird file path behavior in source maps & w/ plugins that use the `to` value when I set it to a directory, so it seems like it might actually need to be the actual output file path directly. Currently w/ the asset API I can't do that, so I've been fighting that a bit over the past couple of days [in a branch](https://github.com/tivac/modular-css/commits/rollup-60-4).

guybedford:
Yes you likely do need the proper ‘to’. In theory the source maps could be
renormalized after generation. That is, all the information is available to
do things in the right order. But yes having the output directory can be
useful information to make these workflows easier.

When designing the emission API I did consider this scenario initially of
getting the directory before the source is known, but the issue with that
is that it is theoretically possible to output hash directories:

assetFileNames: ‘assets/[hash]/[name][extname]’

If we could specifically ban this case then it might be possible to provide
a getAssetDirname function.

If you’d like to pursue such a direction please just create an issue to
track further.

src/watch/fileWatchers.ts
lukastaegert: Ok, but another situation might be that the file is saved first but then a moment later again because some automated process did some modification to it. In this case, the watcher might rerun with an outdated copy of the file (or not because there is still a setTimeout in the watcher, but there is no coded relation between the two).

Also you are doing this on a per-file basis. Wouldn't it make more sense to do the debouncing globally i.e. in `Watcher._makeDirty`? At the moment, the operation triggered by the watcher is a "set dirty" which is rather cheap to perform.

So how about this approach: We do not debounce here. Instead once `Watcher._makeDirty` is triggered, we start a timeout that is reset and restarted if `_makeDirty` is called again within this time. Only if the timeout completes without being interrupted do we trigger a new rollup run. That way, we can await arbitrarily long save operations including many files as long as there are no long pauses between the saves. I would assume that other tools handle this similarly and we no longer have the magic number.

lukastaegert: Also, now that we have TypeScript, it would be nice to replace the underscores in `Watcher` with proper `private` modifiers.

src/watch/fileWatchers.ts
```diff
@@ -72,15 +67,24 @@ export default class FileWatcher {
...
+						if (modifiedTime !== -1) {
+							modifiedTime = -1;
+							this.trigger(id);
+						}
+						return;
+					}
+					throw err;
 				}
+				// debounce
+				if (+stats.mtime - modifiedTime > 50) this.trigger(id);
```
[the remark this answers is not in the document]
guybedford: The goal of a debounce is to remove noisy duplicate events from an event stream, that hold no significant meaning.

An automated process that changes a file for good reason is not an example of a debounce because typically that processing time should be longer than the debounce threshold.

Since the debounce needs to take into account the file modified state, it seems sensible to keep it with the individual file watcher instance as that class is associated with the file's state.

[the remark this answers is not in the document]
guybedford: I've lowered the debounce thresold and renamed the private methods.

lukastaegert:
Add #2280 and #2281 to track.

guybedford:
@lukastaegert thanks for noting those issues specifically, I've included them both here.

src/watch/index.ts
```diff
@@ -21,46 +21,44 @@ import { addTask, deleteTask } from './fileWatchers';
 const DELAY = 100;
 
 export class Watcher extends EventEmitter {
-	dirty: boolean;
-	running: boolean;
-	tasks: Task[];
-	succeeded: boolean;
+	private buildTimeout: NodeJS.Timer;
+	private running: boolean = false;
```
lukastaegert: Any reason why you are initializing to false here but than switch to true in the constructor which would override this?

src/watch/index.ts
```diff
@@ -21,46 +21,44 @@ import { addTask, deleteTask } from './fileWatchers';
...
+			this.rerun = true;
+			return;
 		}
+
+		if (this.buildTimeout) clearTimeout(this.buildTimeout);
+
+		this.buildTimeout = setTimeout(() => {
+			this.buildTimeout = undefined;
+			this.run();
+		}, DELAY);
```
lukastaegert: That's how I thought it should work 👍

src/watch/index.ts
```diff
@@ -83,36 +81,35 @@ export class Watcher extends EventEmitter {
 			.then(() => {
 				this.running = false;
 
-				if (this.dirty) {
+				if (this.rerun) {
+					this.rerun = false;
 					this.run();
```
lukastaegert: Maybe we should still add the usual pre-build timeout here before triggering another run in case there is ongoing file system activity. I.e. `invalidate` instead of `run`.

src/watch/index.ts
```diff
@@ -83,36 +81,35 @@ export class Watcher extends EventEmitter {
...
 export class Task {
-	watcher: Watcher;
-	dirty: boolean;
-	closed: boolean;
-	watched: Set<string>;
-	inputOptions: InputOptions;
+	private watcher: Watcher;
+	private closed: boolean;
+	private watched: Set<string>;
+	private inputOptions: InputOptions;
```
lukastaegert: Nice adding the modifiers 👍

src/watch/index.ts
```diff
@@ -156,7 +153,7 @@ export class Task {
 		});
 	}
 
-	makeDirty(id: string, isTransformDependency: boolean) {
+	invalidate(id: string, isTransformDependency: boolean) {
```
lukastaegert: Invalidate is indeed a much nicer name

src/watch/index.ts
guybedford: Good catch, I've ensured builds never run faster than in intervals of DELAY.

guybedford:
I've included the fixes here. And in the name of better changelog management have squashed and updated the commit message.

lukastaegert:
Thanks, looks good!

> And in the name of better changelog management have squashed and updated the commit message.

Which actually made it a little harder to find out what had changed since I last reviewed 😉 Maybe we should leave this until the merge is actually performed.

src/watch/index.ts
```diff
@@ -21,46 +21,46 @@ import { addTask, deleteTask } from './fileWatchers';
...
-
-		this.dirty = true;
-		this.running = false;
 		this.tasks = ensureArray(configs).map(config => new Task(this, config));
-		this.succeeded = false;
-
-		process.nextTick(() => {
-			this._run();
-		});
+		this.running = true;
```
evocateur: Shouldn't this be `this.running = false;`?

guybedford:
@eight04 this is only for the `transform` hook - `load` is supposed to be idempotent as the cache system will assume it is cacheable between builds.

eight04:
If I want to return a (virtual) module that depends on multiple files, should I return an empty module in `load` then write the actual code and return dependencies in `transform`?

BTW, do we cache virtual modules? (whose id starting with `\0`)

src/rollup/types.d.ts
aMarCruz: @guybedford you are right, `dependencies` solves [this issue](https://github.com/rollup/rollup/issues/1203#issuecomment-303883913) of rollup-plugin-pug

PR #2269, opened 2018-06-14
Use primitive string type in TransformHook

dmitrage:
I beleive it's a typo and it causes issue:
```
Argument of type 'String' is not assignable to parameter of type 'string'. 'string' is a primitive, but 'String' is a wrapper object. Prefer using 'string' when possible.
```


zenmumbler:
Oh and since you’ve enabled linting this class of errors can be caught by e.g. TSLint’s ‘ban’ collection of checks. Worth enabling!

PR #2274, opened 2018-06-14
Add compatibility with umdjs and es6 module import

doug:
Add change similar to that merged into umdjs here: https://github.com/umdjs/umd/pull/125/commits/47e725735b854dfd93c1ca6fcebf1e018bd52606, to allow umd based modules to be imported as valid es6 modules when strict mode does not upgrade this to global window. This allows for importing a umd module with native browser import and keeps compatibility with https://github.com/umdjs/umd.


guybedford:
Thanks for the PR here!

There was already a PR along these lines at https://github.com/rollup/rollup/pull/1964 that got stalled. Would you be interested in picking up that approach here and fleshing out the tests?

PR #2288, opened 2018-06-20
imports of the same external can be independently renamed

kellyselden:
This will be more apparent once #2277 lands, but I thought I would submit on it's own to start the review.

When preserving modules, a use case you might have is a second build step after rollup that "fills in" remaining modules. The first resolved `paths` call is cached, and doesn't allow you to change it based on relative location. I will add inline comments to further clarify.

### DIFF
--- a/test/chunking-form/samples/external-independently-renamed/_config.js
+++ b/test/chunking-form/samples/external-independently-renamed/_config.js
@@ -1,0 +1,19 @@
+'use strict';
+
+let i = 0;
+
+module.exports = {
+	description: 'imports of the same external can be independently renamed',
+	options: {
+		input: ['main1.js', 'lib/main2.js'],
+		external(id) {
+			return id.endsWith('placeholder');
+		},
+		output: {
+			paths() {
+				return i++ === 0 ? './internal' : '../internal';
+			}
+		},
+		experimentalPreserveModules: true
+	}
+};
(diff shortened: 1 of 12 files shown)


test/chunking-form/samples/external-independently-renamed/_expected/es/lib/main2.js
```diff
@@ -0,0 +1 @@
+export * from '../internal';
```
kellyselden: Without the code change, this would be `./internal` because of caching.

test/chunking-form/samples/external-independently-renamed/_config.js
```diff
@@ -0,0 +1,17 @@
...
+let i = 0;
+
+module.exports = {
+	description: 'imports of the same external can be independently renamed',
+	options: {
+		input: ['main1.js', 'lib/main2.js'],
+		external(id) {
+			return id.endsWith('placeholder');
+		},
+		paths() {
```
kellyselden: Currently this only receives `id`, so I use an int to fake knowing the difference between the relative files. Once #2277 lands, this will have `parent` as well as `id`, so you can determine the correct relative import.

guybedford:
@kellyselden I'd be happy to combine this with #2277 if you can please. Just because there may be some better refactoring we can do there around how getPaths is called for the combination.

guybedford:
@kellyselden let me know if you're still interested in combing these?

kellyselden:
This is no longer an issue once that other PR lands. The only thing that is useful is the regression test. I'll add that to the other PR.

guybedford:
Are you saying this one can be closed? Or that they will land cleanly? If so, which should be merged first?

kellyselden:
I added the test from here to #2277. This can be closed if that one lands.

shellscape:

PR #2299, opened 2018-06-24
Fix type definition of Plugin

kimamula:
Fix type definitions of properties of `Plugin` according to the [wiki](https://github.com/rollup/rollup/wiki/Plugins#creating-plugins), which says `options`, `resolveId`, `load`, and `transformChunk` can return `null`.
It is likely that some other properties such as `resolveDynamicImport` can also return `null`, though I only fixed the properties which are explicitly described to possibly return `null`.

### DIFF
--- a/src/rollup/types.d.ts
+++ b/src/rollup/types.d.ts
@@ -81,7 +81,7 @@ export type ResolveIdHook = (
 	this: PluginContext,
 	id: string,
 	parent: string
-) => Promise<string | boolean | void> | string | boolean | void;
+) => Promise<string | boolean | void | null> | string | boolean | void | null;
 
 export type IsExternal = (
 	id: string,
@@ -92,7 +92,7 @@ export type IsExternal = (
 export type LoadHook = (
 	this: PluginContext,
 	id: string
-) => Promise<SourceDescription | string | void> | SourceDescription | string | void;
+) => Promise<SourceDescription | string | void | null> | SourceDescription | string | void | null;
 
 export type TransformHook = (
 	this: PluginContext,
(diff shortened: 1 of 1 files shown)


src/rollup/types.d.ts
```diff
@@ -25,15 +25,17 @@ export interface RollupError {
...
-	sourcesContent?: string[];
-	mappings: string;
-	file: string;
-}
+export type RawSourceMap =
+	| { mappings: '' }
+	| {
+			version: string;
+			sources: string[];
+			names: string[];
```
guybedford: If we instead make the above all optional that might work just as well.

src/rollup/types.d.ts
```diff
@@ -73,15 +75,15 @@ export interface PluginContext {
...
+	warn(warning: RollupWarning | string, pos?: { line: number; column: number }): void;
+	error(err: RollupError | string, pos?: { line: number; column: number }): void;
 }
 
 export type ResolveIdHook = (
 	this: PluginContext,
 	id: string,
 	parent: string
-) => Promise<string | boolean | void> | string | boolean | void;
+) => Promise<string | boolean | void | null> | string | boolean | void | null;
```
guybedford: I honestly thought `null` was included in `void`?

kimamula: Unfortunately not.
Open the following URL and check the `strictNullChecks` option.

https://www.typescriptlang.org/play/#src=function%20foo()%3A%20void%20%7B%0A%20%20%20%20return%20null%3B%0A%7D

src/rollup/types.d.ts
kimamula: I'm afraid not.

If either `source` or `names` is `undefined` when `mappings` is not an empty string, an error is raised [here](https://github.com/rollup/rollup/blob/8ba819f23d46a5a25015b2225aef39be67d0c06a/src/utils/getOriginalLocation.ts#L21-L22).

Issue #2350, opened 2018-07-25
Plugin Development: Logic error in resolver - does not match docs


swernerx:
The docs are mentioning that:

> resolveId – a custom ( importee, importer ) => id resolver (useful for e.g. locating third-party dependencies); you can also return a Promise. Returning null or undefined defers to other resolveId functions (and eventually the default resolution behavior); **returning false signals that importee should be treated as an external module and not included in the bundle.**

(emphasis mine)

Unfortunately it seems like the logic handles all falsy values with the same logic and does not differentiate between `null`/`undefined` and an actual `false`. This means that it ignores a return value of `false` given inside a resolver of a plugin and throws an file-not-found-error instead.

Let's have a look at the responsible code:

```ts
if (!resolvedId && !isExternal) {
	if (isRelative(source)) {
		error({
			code: 'UNRESOLVED_IMPORT',
			message: `Could not resolve '${source}' from ${relativeId(module.id)}`
		});
	}
```

[My plugin](https://www.npmjs.com/package/rollup-plugin-rebase) should be returning relative imports, but these should not be bundled with rollup and kept external (it's about assets like fonts and images).

Looking at the code the issue seems pretty trivial:

```diff
diff --git a/src/Graph.ts b/src/Graph.ts
index 0a5cc5cb..156cd2b6 100644
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -801,7 +801,7 @@ Try defining "${chunkName}" first in the manualChunks definitions of the Rollup
                                                const externalId =
                                                        <string>resolvedId ||
                                                        (isRelative(source) ? resolve(module.id, '..', source) : source);
-                                               let isExternal = this.isExternal.call(this.pluginContext, externalId, module.id, true);
+                                               let isExternal = resolvedId === false || this.isExternal.call(this.pluginContext, externalId, module.id, true);

                                                if (!resolvedId && !isExternal) {
                                                        if (isRelative(source)) {
```

Even the test are fine afterwards as it seems.

swernerx:
Made a PR for this:
https://github.com/rollup/rollup/pull/2351

PR #2351, opened 2018-07-25
Fixed faulty logic related to handling externals when resolver returns false

swernerx:
Fix for https://github.com/rollup/rollup/issues/2350 - handle `resolvedId === false` like `isExternal === true`.


guybedford:
This logic has actually been on my mind the past few days (in particular how we might clean it up and make it more consistent)... will take a look soon.

guybedford:
Another thing that would help this move forward quicker is having a test (perhaps in api?), will add one myself here otherwise.

Issue #2358, opened 2018-07-28
Rollup no longer warns when no name was provided for global module


alan-agius4:
In the previous versions of rollup when no name was provided for a `global` one would get a warning stated something along;
 ```
ionic-angular (guessing 'ionicAngular')
```

However this is no longer the case when you use `globals` is specified. If it's specified and the value returned `undefined` or `null` no warning will be emitted and in the final output even if the moduleName won't be resolved.

```js
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? factory(require('ionic-angular')) :
	typeof define === 'function' && define.amd ? define(['ionic-angular'], factory) :
	(factory(null));
}(this, (function (ionicAngular) { 'use strict';
})));
```

I would expected that there would be a warning when I specify the `globals` option however I dont set it for a specific module.

rollup.config.js
```
export default {
    input: 'index.js',
    output: {
        file: 'bundle.js',
        format: 'umd',
        globals: []
    }
};
```
index.js
```
import * as foo from 'ionic-angular';
```

Terminal
```
Alans-iMac:rollup-test alan$ node_modules/.bin/rollup -c rollup.config.js

index.js → bundle.js...
created bundle.js in 10ms
```

Relates to: https://github.com/dherges/ng-packagr/pull/1026

guybedford:
Apologies for the regression, my fault :)

PR #2359, opened 2018-07-28
fix: no longer warns when no name was provided for global module

alan-agius4:
When the dictionary lookup or function look up return undefined or null, a warning should be emitted and the module name should be guessed.

If the user, doesn't receive a warning that the global name was not found, they won't be able to fix a possible issue. If the global name is incorrect or later set to null.

Closes: #2358

Some more context
https://github.com/dherges/ng-packagr/pull/1026#issuecomment-408605437

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -68,8 +68,17 @@ function getGlobalName(
 	graph: Graph,
 	hasExports: boolean
 ) {
-	if (typeof globals === 'function') return globals(module.id);
-	if (globals) return globals[module.id];
+	let globalName: string | undefined;
+	if (typeof globals === 'function') {
+		globalName = globals(module.id);
+	} else if (globals) {
+		globalName = globals[module.id];
+	}
+
+	if (globalName) {
+		return globalName;
+	}
+
 	if (hasExports) {
 		graph.warn({
 			code: 'MISSING_GLOBAL_NAME',
--- a/test/misc/index.js
+++ b/test/misc/index.js
@@ -543,4 +543,36 @@ describe('misc', () => {
 				);
 			});
 	});
+
+	it('warns when globals option is specified and a global module name is guessed in a UMD bundle (#2358)', () => {
+		const warnings = [];
+
+		return rollup
+			.rollup({
+				input: 'input',
+				plugins: [
+					loader({
+						input: `import * as _ from 'lodash'`
+					})
+				],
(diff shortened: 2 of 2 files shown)


src/Chunk.ts
```diff
@@ -68,8 +68,17 @@ function getGlobalName(
...
+	let globalName: string | undefined;
+	if (typeof globals === 'function') {
+		globalName = globals(module.id);
+	} else if (globals) {
+		globalName = globals[module.id];
+	}
+
+	if (globalName) {
+		return globalName;
+	}
```
guybedford: I think it's ok to not warn when the user has not specified any configuration, as in many cases this is the required behaviour that the package name be the request package name (eg this is the default on npm that a request package name and output package name can match).

The issue maybe arises in environments in which you are defining a global function or map that is trying to be authoritative and then has these invisible misses.

Instead, perhaps let's only warn for the case where `globals` is provided but has misses, by adding that as another case above?

alan-agius4: I am not 100% sure about this, as for UMD, I noticed that a lot of guessed name would be incorrect. If for instance having a `.` in the name. Or when the umd bundle name is totally different, and without a warning, it's hard to rollup users to know that something might be wrong.

However I can do what you suggested.

alan-agius4: @guybedford pushed an update.

src/Chunk.ts
```diff
@@ -68,9 +68,18 @@ function getGlobalName(
...
+
+	if (hasExports || (hasExports ** globals && !globalName)) {
```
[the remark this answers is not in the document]
alan-agius4: yeah looks like the IDE didn't save :P

src/Chunk.ts
```diff
@@ -68,9 +68,18 @@ function getGlobalName(
...
+		globalName = globals(module.id);
+	} else if (globals) {
+		globalName = globals[module.id];
+	}
+
+	if (globalName) {
+		return globalName;
+	}
+
+	if (hasExports || (hasExports && globals && !globalName)) {
```
guybedford: `if (hasExports || globals && !globalName)` would be simpler here I think.

guybedford:
I've been thinking about this some more and maybe you are right we should always warn actually...

alan-agius4:
hahah, removed the last commit :-)

guybedford:
LGTM

PR #2360, opened 2018-07-29
missing export + preserve modules bug

kellyselden:
I tried to narrow it down to the minimal failure. I'm not sure why it is failing yet.

```
TypeError: Cannot read property 'variable' of undefined
      at Chunk.populateEntryExports (src/Chunk.ts:232:5)
      at /Users/kselden/code/rollup/src/Graph.ts:496:7
```

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -353,7 +353,9 @@ export default class Chunk {
 		}
 
 		if (name === 'default') {
-			return;
+			// we found a missing default export, so
+			// let the module create it's own shim.
+			return { variable: module.traceExport(name), module };
 		}
 
 		// external star exports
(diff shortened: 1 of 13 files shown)


kellyselden:
I applied a fix. Let me know what you think.

src/Chunk.ts
```diff
@@ -344,7 +344,9 @@ export default class Chunk {
 		}
 
 		if (name === 'default') {
-			return;
+			// we found a missing default export, so
+			// let the module create it's own shim.
+			return { variable: module.traceExport(name), module };
```
guybedford: This is only possible under shimMissingExports right? Because otherwise there would have been an earlier throw?

Is it not possible to ensure this is populated earlier in the Module trace process? Is the issue that shimMissingExports is not populating module.reexports?

shellscape:

PR #2366, opened 2018-08-01
Add caching for resolve load transform hooks

keithamus:
Refs https://github.com/rollup/rollup/issues/2301

### What?

This adds an initial implementation for caching the `resolveId`, `load`, and `transform` hooks.

### Why?

Adding caching for these hooks should greatly speed up the time it takes rollup to compute a bundle. At GitHub, we've been using Rollup in production since January - building our bundles hundreds of times a day: we've seen significant impact in CI reduction by hacking in a caching layer into Rollup. With our caching layer we see build times taking about 5% the time compared to cold cache. This PR implements most of what we do with our build - with this PR and a primed cache we should see numbers around 10% that of uncached builds.

<details><summary>Some more info about how we get those numbers</summary>

In real world numbers: GitHub's JS code (on my machine) builds in about ~80s with no caching, by caching just the transform hooks we see that number go down to ~30s. Caching transform as well as load/resolve reduces that to ~6s. We also cache the entire rollup end-to-end (not featured in this PR) which gets us down to ~3s.

Testing with the GitHub JS code, we can say `resolveId` takes about ~15% of the time that rollup does its work, `load` takes another ~15%, and `transform` (specifically for us - Babel) takes ~60% of the time (the eagle-eyed here might see this gets us to 90% - the remaining 10% is small bits here and there - e.g. initialization, Rollup's finalizers, etc).

`resolveId` (specifically `rollup-plugin-node-resolve` for us, which I imagine is what most folks are also using) is a lot of tree-traversal/stat calls. Stat calls which are pretty cheap (on an SSD maybe a few milliseconds) - but it does _a lot of them_. For our bundles we see stat calls in the multiple thousands, and readFiles (for `package.json`) in the hundreds. So mostly I/O here.

`load` pretty much does what it says on the tin - it usually just implements `readFile`. So most of the time spent here is just regular I/O.

`transform` for us (and I imagine many other Rollup users) is predominantly Babel. Babel is surprisingly slow - there is very little I/O here, but instead Babel is very specifically CPU intensive and accounts for the lions share of computation.

Luckily each of these can have a pretty similar caching mechanism and so we can implement all 3 at once and chip away the 90% figure, getting build times to about 10% of what they are without cache.
</details>

### How?

A plugin that exposes these hooks and wants them to be cached must also expose a `cacheKey` property which can be a string of their choosing. Conventionally they should roll this key for each version of the plugin or its dependencies (e.g. rollup-plugin-babel might derive its `cacheKey` from the combined versions of all plugins it is given).

Each hook has its own caching key which is derived from multiple parts: the hook name, the plugin name and version, and the arguments given to it. `load` also adds the `mtime` & `size` of the file expected to be loaded...

<details><summary>Some more details about why `load` adds those extras here</summary>

The `load` cache key is designed kind of like a really low-rent Bloom filter - As discussed in #2301, if the file changes mtime or size it's very likely the call should skip the cache and call `load`. "Very likely" is good enough! If a file changes these attributes it's probably because the file itself has changed. `stat` is a super cheap way to check if a file has _likely_ changed, without doing expensive computation of reading a file and calculating hash digests. There _is_ a possibility of the file changing without the mtime or size changing (e.g. changing one character won't change the size, and then `touch -A` can be used to fix the mtime as if the file was never modified) - this is a fringe case but one we might want to look at.
</details>

### What next?

Without any plugins opting into this mechanism, this PR is effectively a no-op. This PR requires opt-in from both plugins (by exposing a `cacheKey`) and by users (by saving the cache somewhere). As plugins start exposing a `cacheKey` (which I'll have follow-up PRs for) we may start discovering bugs, certainly edge cases.

One big outstanding issue is the issue of `PluginContext`. The discussion has started over in https://github.com/rollup/rollup/issues/2301#issuecomment-407474437. We need to solve this before we really merge, I think.

### DIFF
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -1,0 +1,12 @@
+import { HookCache, Plugin } from './rollup/types';
+
+export function cacheResolveIdHook(cache: HookCache, plugin: Plugin) {
+	return function cachedResolveId(id: string) {
+		const key = `resolve|${plugin.name || '(anonymous plugin)'}|${plugin.cacheKey}|${id}`;
+		return key in cache
+			? Promise.resolve(cache[key])
+			: Promise.resolve()
+					.then(() => plugin.resolveId.call(this, id))
+					.then(value => (cache[key] = value));
+	};
+}
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -32,3 +32,14 @@ export function cacheLoadHook(cache: HookCache, plugin: Plugin) {
 			});
 	};
 }
+
+export function cacheTransformHook(cache: HookCache, plugin: Plugin) {
+	return function cachedTransform(code: string, id: string) {
+		const key = `transform|${plugin.name || '(anonymous plugin)'}|${plugin.cacheKey}|${id}|${code}`;
+		return key in cache
+			? Promise.resolve(cache[key])
+			: Promise.resolve()
+					.then(() => plugin.transform.call(this, code, id))
+					.then(value => (cache[key] = value));
+	};
+}
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -1,3 +1,4 @@
+import { stat } from './fs';
 import { HookCache, Plugin } from './rollup/types';
 
 export function cacheResolveIdHook(cache: HookCache, plugin: Plugin) {
@@ -8,5 +9,26 @@ export function cacheResolveIdHook(cache: HookCache, plugin: Plugin) {
 			: Promise.resolve()
 					.then(() => plugin.resolveId.call(this, id))
 					.then(value => (cache[key] = value));
+	};
+}
+
+export function cacheLoadHook(cache: HookCache, plugin: Plugin) {
+	return function cachedLoad(id: string) {
+		// An inherently side-effectful part of `.load` is the filesystem.
+		// We want to invalidate the cache if the file has changed.
+		// A good cheap metric for this is mtime and size. We combine these into the
+		// cache key so that files with different mtime/size invalidate the cache
+		return stat(id)
+			.then(({ mtimeMs, size }) => [mtimeMs, size], () => [0, 0])
+			.then(([mtimeMs, size]) => {
+				const key = `load|${plugin.name || '(anonymous plugin)'}|${
+					plugin.cacheKey
+				}|${id}|${mtimeMs}|${size}`;
+				return key in cache
+					? Promise.resolve(cache[key])
+					: Promise.resolve()
+							.then(() => plugin.load.call(this, id))
+							.then(value => (cache[key] = value));
+			});
 	};
 }
--- a/src/utils/transform.ts
+++ b/src/utils/transform.ts
@@ -14,6 +14,7 @@ import {
 	TransformSourceDescription
 } from '../rollup/types';
 import { EmitAsset } from './assetHooks';
+import { cacheTransformHook } from './cache';
 import error from './error';
 import getCodeFrame from './getCodeFrame';
 import { dirname, resolve } from './path';
@@ -117,15 +118,14 @@ export default function transform(
 	let promise = Promise.resolve(source.code);
 	let transformDependencies: string[];
 
-	plugins.forEach(plugin => {
-		if (!plugin.transform) return;
-
+	plugins.filter(plugin => 'transform' in plugin).forEach(plugin => {
+		const transform = cacheTransformHook(graph.cachedHooks, plugin);
 		promise = promise.then(previous => {
 			const { assets, emitAsset } = createTransformEmitAsset();
 			return Promise.resolve()
 				.then(() => {
 					const context = createPluginTransformContext(graph, plugin, id, previous, emitAsset);
-					return plugin.transform.call(context, previous, id);
+					return transform.call(context, previous, id);
 				})
 				.then(result => {
 					// assets emitted by transform are transformDependencies
--- a/test/cache/index.js
+++ b/test/cache/index.js
@@ -5,6 +5,7 @@ describe('cache', () => {
 	let genPlugin;
 	let resolveIdCalls;
 	let loadCalls;
+	let transformCalls;
 	const expectedCode = `(function () {
 	'use strict';
 
@@ -15,6 +16,7 @@ describe('cache', () => {
 	beforeEach(() => {
 		resolveIdCalls = [];
 		loadCalls = [];
+		transformCalls = [];
 		genPlugin = (modules, i = 1) => ({
 			name: 'test-plugin',
 			cacheKey: `v${i}`,
@@ -25,11 +27,15 @@ describe('cache', () => {
 			load(id) {
 				loadCalls.push(id);
 				return modules[id];
+			},
+			transform(code) {
+				transformCalls.push(code);
+				return code;
 			}
 		});
 	});
 
-	it('caches calls to resolve, load on repeat loads', () => {
+	it('caches calls to resolve, load, transform on repeat loads', () => {
 		const plugins = [genPlugin({ x: `console.log( 42 );` })];
 		let cache = {};
 		return rollup
@@ -47,6 +53,7 @@ describe('cache', () => {
 			.then(({ code }) => {
 				assert.deepEqual(resolveIdCalls, ['x']);
 				assert.deepEqual(loadCalls, ['x']);
+				assert.deepEqual(transformCalls, ['console.log( 42 );']);
 				assert.deepEqual(code, expectedCode);
 			});
 	});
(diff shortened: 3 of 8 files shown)


src/utils/cache.ts
```diff
@@ -0,0 +1,45 @@
+import { stat } from './fs';
+import { HookCache, Plugin } from './rollup/types';
+
+export function cacheResolveIdHook(cache: HookCache, plugin: Plugin) {
+	return function cachedResolveId(id: string) {
+		const key = `resolve|${plugin.name || '(anonymous plugin)'}|${plugin.cacheKey}|${id}`;
```
guybedford: My concern with resolver caching is that we aren't tracking invalidation on the stat paths of the resolver itself, but instead relying on any resolve plugin to change its cacheKey somehow (likely a time interval?) based on that information, which might not be obvious.

One of the goals for the cache is to be able to by-default persist to disk out-of-the-box and I think that we should try and aim for this in the caching work. As a result we should be more cautious than that I think.

So my first thought on this would be to say we shouldn't cache the resolver for now, and can come back to this in due course (and since you say it is only 15% of the problem...), but let me know if you have further thoughts here.

src/utils/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+	};
+}
+
+export function cacheLoadHook(cache: HookCache, plugin: Plugin) {
+	return function cachedLoad(id: string) {
+		// An inherently side-effectful part of `.load` is the filesystem.
+		// We want to invalidate the cache if the file has changed.
+		// A good cheap metric for this is mtime and size. We combine these into the
+		// cache key so that files with different mtime/size invalidate the cache
+		return stat(id)
```
guybedford: We have to be weary of tying the implementation to the file system. Rollup runs equally in a browser environment, with the default load hook only applying if no other plugin is present. Also module Ids in Rollup don't strictly have to be file paths, so we need to be aware of this as well.

For this reason, my initial proposal was, rather than specifying an exact cache key implementation like you have done here, leave this exact part to the plugins themselves, offering a lower level primitive that doesn't tie to the file system.

src/utils/transform.ts
```diff
@@ -117,15 +118,14 @@ export default function transform(
...
-
+	plugins.filter(plugin => 'transform' in plugin).forEach(plugin => {
+		const transform = cacheTransformHook(graph.cachedHooks, plugin);
 		promise = promise.then(previous => {
 			const { assets, emitAsset } = createTransformEmitAsset();
 			return Promise.resolve()
 				.then(() => {
 					const context = createPluginTransformContext(graph, plugin, id, previous, emitAsset);
-					return plugin.transform.call(context, previous, id);
+					return transform.call(context, previous, id);
```
guybedford: I'd be interested to know if you found this makes any performance difference, as we actually already do cache the transform hook (this is the only thing we cache currently). There is a check [here](https://github.com/rollup/rollup/blob/master/src/Graph.ts#L690) that reuses the transformed result when the input source is matched. Ideally this should be cacheKey based though as transform options changing would currently not invalidate this.

The current cache also deals with the `emitAsset` problem by noting which assets are emitted by the transform hook and then automatically re-emitting them when using the cached transform. I'm hoping to get a PR together tomorrow which can do a similar thing for the load function to hopefully unblock this caching direction.

test/cache/index.js
```diff
@@ -0,0 +1,122 @@
...
+	});
+
+	it('will not call transform if given correctly primed cache', () => {
+		const plugins = [genPlugin({ x: `console.log( 42 );` })];
+		return rollup
+			.rollup({
+				input: 'x',
+				plugins,
+				cache: {
+					hooks: { 'transform|test-plugin|v1|x|console.log( 42 );': 'console.log( 42 )' }
```
guybedford: As mentioned, this should already be captured I think.

test/test.js
```diff
@@ -16,5 +16,6 @@ describe('rollup', function() {
 	require('./incremental/index.js');
 	require('./hooks/index.js');
 	require('./cli/index.js');
+	require('./cache/index.js');
```
guybedford: +1 to having dedicated cache tests.

src/utils/cache.ts
keithamus: I agree entirely with these constraints. Here's some ideas:

Giving keys a TTL is easy enough:

```js
const key = `resolve|${plugin.name || '(anonymous plugin)'}|${plugin.cacheKey}|${id}|${~~(Date.now()/60000)}`
```
Of course a TTL isn't really the best for this - because the cache could be invalidated before the TTL is up, or could still be valid past the cache's timeout.

Another thing we could do, is read from the cache and compare it to what was known about the file at the time - e.g. `stat()` calling. This way we capture the external state (the file). Reading ahead I see there is a constraint on the browser - I'll address that below.

```js
const key = `resolve|${plugin.name || '(anonymous plugin)'}|${plugin.cacheKey}|${id}`
if (cache[key] && cache[key].path) {
  // Now we have a cache key we can double check to ensure the mtime/size are the same as what is on disk
  // and only return if they match
  const { mtimeMs, size } = await stat(cache[key].path).catch(() => ({ mtimeMs: 0, size: 0 }))
  if (mtimeMs === cache[key].mtimeMs && size === cache[key].size) {
    return Promise.resolve(cache[key])
  }
}
const path = await plugin.resolveId.call(this, id)
// Now we have the path we can stat it to get mtime and size for it
const { mtimeMs, size } = await stat(cache[key].path).catch(() => ({ mtimeMs: 0, size: 0 }))
cache[key] = { path, mtime, size }
return path
```

src/utils/cache.ts
keithamus: The simple answer here is that `stat` is caught, so any failure to read from disk (including the error thrown from the stubbed version) should be caught and the `[0, 0]` tuple returned.

The issue with allowing plugins to generate their own keys is it leaves a lot of responsibility on them to get it right. This - to me - could be problematic as plugins are often not as well tested as core, and missing pieces of state like the filesystem can surface.

If we were to allow plugins to generate their own keys I think we'd need to define a way for plugins to hit the disk with an object we can introspect or otherwise control, so that the load plugin doesn't just return file contents, but returns a file object (think something like [vynil fs](https://github.com/gulpjs/vinyl-fs)). We can then use the file object to decide how it can be cached (e.g. if the file object represents a file on disk, then stat it, if it is something else then apply other rules). If the only way `load` can operate is to return a `File` object then that changes the dynamic and allows us to do alot (e.g. utilising a `File.hasChanged() === true?`)

src/utils/transform.ts
keithamus: I do remember seeing this - and attempting to get it persisting to disk, but for some reason couldn't. I can't remember the exact reasons why. I'm happy to move this from this PR if I can figure out how to get the module cache working 😃

keithamus:
@guybedford thanks for your initial review. Perhaps what wasn't clear from the OP was that this was an initial sketch. Happy to rework any/all of this. Good to get the discussion going.

I think, as stated above in the review comments - one barrier we'll have to moving forward with caching is that _we need to look at the disk_ if the disk is where plugins get their return values from. Essentially if we want to keep flexibility of `load` and `resolveId` we need them to somehow let the cache system know where they got their sources.

src/utils/cache.ts
guybedford: The issue with statting the file directly, is edge cases like having a package.json where there is no main, and then a main gets added. Statting the index.js file won't tell you that the package.json file was changed to add a main. And if you stat the package.json file, and the next edge case and the next we're doing all the work of the resolver when it comes down to it.

So I tend to think the best cache for a resolver is a TTL-based one, and that the implementation for this should probably still be further downstream from Rollup itself.

Would you consider a PR to the rollup-plugin-node-resolve project here to add a resolver with this kind of functionality? We would need a way to add the plugin state to the Rollup cache object, which perhaps may be something we can open up from the Rollup side with something like `this.setCache('key', value)` and `this.getCache('key', value)` for plugins or similar. Let me know what you think along those lines?

src/utils/cache.ts
guybedford: So here's how I'd imagine a plugin implementing the load hook with a `cacheKey`:

```js
async load (id, cacheKey) {
  const [mtime, size] = cacheKey.split('|');
  const stats = await new Promise((resolve, reject) => 
    fs.stat(id, (stats, err) => err ? reject(err) : resolve(stats)
  );
  if (mtime === stats.mtime && size === stats.size)
    return true;
  // ...usual load routine
}
```

if we document this well, it should be clear to plugins how to use this, and most will likely just copy-paste the snippet anyway.

The benefit here is that plugins then have the flexibility to extend and modify this behaviour based on their exact load method, and I think that will be clear given the above snippet given to them.

We do have the plugin lists, so we could even PR these ourselves once we can agree on the API format.

I'm completely open to API alternatives here as well too.

src/utils/transform.ts
guybedford: Would be very interesting to know the stumbling block as you likely wouldn't be the first on that either :)

guybedford:
@keithamus thanks for taking the time to be able to work on this and take feedback. It's rare we get that sort of contribution, so I'm keen to make the most of it as well and will do my best to review and contribute.

I guess I'm advocating primitives that push the actual disk access to the plugins as stated in my examples on the latest comments. Look forward to discussing further.

keithamus:
Closing this in favour of smaller more focussed PRs.

PR #2369, opened 2018-08-02
Asset fixes and refactoring for caching

guybedford:
This fixes up a couple of failing asset tests, as well as improving validations and refactoring with the new caching changes in mind to ensure the asset system can be compatible with a long-running cache system.

#### Transform hook caching

* Adds a test for and fixes automatic asset re-emission for that module when the transform hook call is cached for a module
* setAssetSource is no longer supported in transform and will throw. This way, we only need to cache the emitAsset calls in transform.
* Refactors emitAsset wrapping for transform, in preparation for having this apply equally to load.
* ~~Extend asset.type to be a string, `"transform"` or `null`, in preparation for distinction between load and transform assets (`"load"`)~~ reverted as current caching direction will avoid this.

#### Validations and bug fixes

* Fixes the test for failure when emitting an asset with an asset dependency at generate time (watcher applies after build not generate)
* Adds support for watch dependencies to setAssetSource
* Ensures dependencies are not supported for emitAsset without a source
* Throws when setAssetSource is called without a source


guybedford:
@lukastaegert ideally I would like all hooks to still run with caching, but just to short-path internally with cached data. This way we don't need to worry about the side effects problem, and can just focus on ensuring the "predictable repeatability" property exists.

guybedford:
Thanks for the review, I've made the typing correction.

lukastaegert:
> This way we don't need to worry about the side effects problem

So then we could unify the plugin contexts again? Sounds perfect to me!

guybedford:
Yes that is the goal. Although due to having unique cache info the current
plan is to have per-plugin-instance contexts as in the latest PR. Looking
forward to getting this stuff in!

PR #2370, opened 2018-08-03
Add memory perf metric along with time metrics

TrySound:
image

Values looks a bit weird, but I think it's something like GC diagram


TrySound:
Seems like metrics are bad parallelized for both time and memory. Will try to fix it in the next PR.

TrySound:
image

PR #2373, opened 2018-08-03
missing export + preserve modules + import default expression bug

kellyselden:
AMD and CJS are generating some `__chunk_1.undefined` safe names.

I got lost in the implementation, but I got to the point of seeing a disconnect between dep2 chunk's `exports` and `exportNames`. `exports` had the shimmed variable but `exportNames` didn't, which was tripping up `chunk.getVariableExportName`.


shellscape:

PR #2375, opened 2018-08-05
Don't test entry points for external

TrySound:
Ref https://github.com/rollup/rollup/issues/1343

I just found that external is almost the same as `resolveId`. Currently
for the stuff like ignoring all `module` paths we should consider entry
point which may look similar and do not start with '.' or '/'. This is
annoying since external entry points are not allowed.

In this diff I suggest to not test entry points for external at all and
skip this check in user configs.

### DIFF
--- a/src/Graph.ts
+++ b/src/Graph.ts
@@ -147,7 +147,7 @@ export default class Graph {
 		this.resolveId = first(
 			[
 				((id: string, parentId: string) =>
-					this.isExternal(id, parentId, false) ? false : null) as ResolveIdHook
+					parentId != null && this.isExternal(id, parentId, false) ? false : null) as ResolveIdHook
 			]
 				.concat(
 					this.plugins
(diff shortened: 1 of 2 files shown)


src/Graph.ts
```diff
@@ -147,7 +147,7 @@ export default class Graph {
 		this.resolveId = first(
 			[
 				((id: string, parentId: string) =>
-					this.isExternal(id, parentId, false) ? false : null) as ResolveIdHook
+					parentId != null && this.isExternal(id, parentId, false) ? false : null) as ResolveIdHook
```
shellscape: is the loose `null` check intentional?

shellscape:
This would technically be a breaking change, since we'd no longer be creating an error. @guybedford will probably want to get this merged into his 1.0 branch.

src/Graph.ts
TrySound: I usually do not distinct `null` and `undefined` and even prefer to not use `undefined` value explicitly. Same true for rollup hooks. `resolveId`, `load`, etc may return both `null` and `undefined`.

guybedford:
I actually just added this feature in https://github.com/rollup/rollup/pull/2382, so we can let that supersede this one.

Issue #2363, opened 2018-07-30
Consider supporting transformDependencies for load


guybedford:
From discussion at https://github.com/rollup/rollup/pull/2259#issuecomment-406855965.

If you look at even standard examples like https://github.com/rollup/rollup/pull/2259/files#diff-db4f35ec21ccb04b7e2a26b7654c4397R327 they are use cases that can be handled via `load`.

Perhaps we need to think a little more generally about this.

//cc @eight04 @keithamus

eight04:
I think when some code is generated by the plugin, the plugin should be able to define the dependency at that point. This includes `load`, `transform`, and maybe `transformChunk` (inline license header, footer, etc).

guybedford:
One thing that would make this simpler is actually just changing it from `return { dependencies }` to a contextual function - `this.addWatchDependency(name)`.

guybedford:
This could also replace the asset dependencies argument currently in place as well.

guybedford:
I've looked into this quite deeply, and I think the complexity added to the whole system outweighs the benefit over just keeping dependencies to the `transform` function.

So I'd suggest rather sticking to using the `transform` hook for doing this after all. Happy to discuss further.

guybedford:
(note the complexities are specifically around caching here)

guybedford:
I ended up implementing a `this.addWatchFile` in https://github.com/rollup/rollup/pull/2389.

PR #2389, opened 2018-08-08
Plugin cache

guybedford:
This provides a plugin cache on plugin contexts with the API:

```js
class PluginCache {
  get (key: string): any;
  set (key: string, value: any): void;
  has (key: string): boolean;
  delete (key: string): boolean;
}
```

The cache is stored in `bundle.cache.plugins[cacheKey]` as a plain object, which is not directly accessible.

Each value in the cache is stored as `[number, value]`, where the first number indicates the access count. This is zero if it was accessed in the last build, and is incremented each build. When it reaches 10 (10 builds without any access through get, set, or has) we automatically evict the cache item.

The cache key for a plugin is `plugin.cacheKey || plugin.name`. When using `plugin.name` this will throw if on duplicate or missing plugin names as soon as cache access is attempted (a get / set / has / delete call).

Two instances of the same plugin can support sharing a cache instance in the same build provided they use a `plugin.cacheKey` and not a `plugin.name`.

There are some complexities around supporting the transform caching and transform dependencies stuff. We basically detect `this.cache` usage in transform and then fully opt-out of the automatic transform caching - no transform dependencies invalidations, no asset reemissions etc.

The hope is to eventually deprecate the automatic transform caching along with its invalidations and asset reemissions. To prepare for this `watchFiles` has been refactored in the build as well, with the goal to just always rerunning all hooks and relying on the plugin internal cache.

Edit - New additions:

* `cacheExpires` option: allows setting an integer representing how many builds should run before unused keys are cleared from the plugin cache
* `this.addWatchFile` in plugins - allows a more generic feature for plugin watcher dependencies, as previously tracked in https://github.com/rollup/rollup/issues/2363. Again this aligns with moving away from special transform handling and deprecating both transform dependencies and auto caching.


src/utils/pluginDriver.ts
```diff
@@ -39,22 +41,36 @@ export type HookContext = (context: PluginContext, plugin?: Plugin) => PluginCon
...
 	const plugins = [...(options.plugins || []), getRollupDefaultPlugin(options)];
 	const { emitAsset, getAssetFileName, setAssetSource } = createAssetPluginHooks(graph.assetsById);
+	const existingPluginKeys: string[] = [];
 
 	let hasLoadersOrTransforms = false;
 
 	const pluginContexts = plugins.map(plugin => {
+		let cacheable = true;
+		if (typeof plugin.name === 'string') {
+			if (existingPluginKeys.indexOf(plugin.name) === -1) cacheable = false;
```
keithamus: Is this correct? Do you mean `!== -1` here? I'm a little confused at this block of code.

guybedford: Yup, did not mean to imply this was bug-free just yet :)

src/utils/pluginDriver.ts
```diff
@@ -39,22 +41,36 @@ export type HookContext = (context: PluginContext, plugin?: Plugin) => PluginCon
...
+	const existingPluginKeys: string[] = [];
 
 	let hasLoadersOrTransforms = false;
 
 	const pluginContexts = plugins.map(plugin => {
+		let cacheable = true;
+		if (typeof plugin.name === 'string') {
+			if (existingPluginKeys.indexOf(plugin.name) !== -1) cacheable = false;
+			existingPluginKeys.push(plugin.name);
+		}
```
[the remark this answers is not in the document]
guybedford: Yes, this was half-baked and missed the logic I intended wrt cacheKey allowing shared instance caches between multiple instances of the same plugin in the build.

guybedford:
Ok the tests are up, ready for review.

src/utils/pluginDriver.ts
```diff
@@ -39,22 +41,49 @@ export type HookContext = (context: PluginContext, plugin?: Plugin) => PluginCon
...
+		graph.assetsById,
+		graph.watchFiles
+	);
+	const existingPluginKeys: string[] = [];
 
 	let hasLoadersOrTransforms = false;
 
 	const pluginContexts = plugins.map(plugin => {
+		let cacheable = true;
+		if (typeof plugin.cacheKey === 'string') {
```
keithamus: This is an empty condition. Is it necessary?

guybedford: It's a ruling-out case for the others to avoid a repeated check. Could be done better certainly, but that was the logic that made sense to me.

guybedford: I've "conventionalized" it :P

guybedford:
I've updated the description in the PR now as well so it is clear what is being provided here.

src/Graph.ts
```diff
@@ -34,6 +36,9 @@ import relativeId, { getAliasName } from './utils/relativeId';
 import { timeEnd, timeStart } from './utils/timers';
 import transform from './utils/transform';
 
+// clear plugin cache items after 10 builds unnaccessed
+const PLUGIN_CACHE_EVICTION_EXPIRY = 10;
```
keithamus: It would be good to see this number configurable; CI machines might build dozens of branches per day may benefit from larger numbers so they aren't evicting cache quite as much.

shellscape: @keithamus would that be better as an environment variable, or configuration option?

keithamus: I guess a configuration option - considering pretty much everything else is. With a rollup.config.js it's relatively easy to expose it as a environment variable.

guybedford: I've added a `cacheExpiry` configuration option.

src/rollup/types.d.ts
```diff
@@ -353,9 +364,14 @@ export interface OutputChunk {
 	map?: SourceMap;
 }
 
+export interface SerialisablePluginCache {
```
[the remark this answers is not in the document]
guybedford: Have to double down on that stuff :P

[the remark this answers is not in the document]
guybedford: Ah `serialize` is British English - will change.

src/utils/assetHooks.ts
```diff
@@ -96,6 +97,11 @@ export function createAssetPluginHooks(
 			const asset: Asset = { name, source, fileName: undefined, dependencies, transform: null };
 			if (outputBundle && source !== undefined) finaliseAsset(asset, outputBundle, assetFileNames);
 			assetsById.set(assetId, asset);
+			if (!asset.transform && asset.dependencies && asset.dependencies.length) {
+				for (const depId of asset.dependencies) {
+					if (watchFiles.indexOf(depId) === -1) watchFiles.push(depId);
```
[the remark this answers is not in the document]
guybedford: Agreed, added.

src/utils/pluginDriver.ts
```diff
@@ -192,3 +226,95 @@ export function createPluginDriver(
...
+			const item = cache[id];
+			if (!item) return undefined;
+			item[0] = 0;
+			return item[1];
+		},
+		set(id: string, value: any) {
+			cache[id] = [0, value];
+		},
+		delete(id: string) {
+			return delete cache[id];
```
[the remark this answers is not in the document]
guybedford: Objects with lots of key changes are treated as "dictionary objects" with hash table keys (which is exactly what we need here). In that case the delete cost is the delete cost of the hash table which shouldn't be a problem. Also we do want to evict unused keys to avoid cache bloat.

guybedford:
> If we want to prevent invalid caches by resetting the cache after n runs, why would there be no invalid cache after n-1 runs? Or is the idea that if the watcher does not trigger, people should just keep making changes until they hit the expiry limit?

The cache expiry is used to avoid memory bloat, not to avoid false positives. That is, when we persist the cache to disk, then we need to ensure that the size of the cache is not going to be monotonic. So every time an entry is touched we bump its counter, and if it hasn't been touched for 10 builds, we evict it to handle this case.

Definitely worth making sure this is properly documented though, given the ease of confusion here.

guybedford:
Included feedback, and rebased to plugin-driver.

lukastaegert:
> Definitely worth making sure this is properly documented though, given the ease of confusion here

Yes indeed. In that case, I would strongly suggest to make this an experimental option for now to signal we might still change how this is configured. Does not feel like something people should play around with usually and in case there are reasons to change it, maybe there is a better way to configure it than with a number but we do not know yet

tivac:
How will using `this.addWatchFile` work in practice? Currently with transform dependencies if a dependency changes the build will restart and transform will be called for the module that registered that dependency. If that approach is now deprecated I'd like to better understand the replacement since I'll need to port over to it and the previous behavior, while not ideal, eventually worked for my needs.

lukastaegert:
Sorry, forgot to change the base of this branch before merge.

guybedford:
Just rebasing, will merge shortly.

guybedford:
@tivac 

> How will using this.addWatchFile work in practice? Currently with transform dependencies if a dependency changes the build will restart and transform will be called for the module that registered that dependency. If that approach is now deprecated I'd like to better understand the replacement since I'll need to port over to it and the previous behavior, while not ideal, eventually worked for my needs.

So the idea is to deprecate both the current transform dependencies and transform cache, eventually, but the deprecation path hasn't even started yet so no need to worry here things will continue to work fine even in 1.0 most likely for a 2.0 removal.

The new approach is:

* addWatchFile adds a file for the watcher to invalidate the build on. This invalidation does not affect the build caching at all though, unlike currently where the transform dependencies invalidate the transform cache.
* Instead of the transform hook by default being cached, all hooks will always be re-run with cache information at this.cache. The plugin has its own responsibility to emit assets and check `this.cache` to handle cached running.
* The full separation between addWatchFile and transform cache will provide more flexible fine-grained cache behaviours.

PR #2397, opened 2018-08-11
WIP: disk cache

guybedford:
This is a work in progress towards https://github.com/rollup/rollup/issues/2182.

Note that it ONLY persists the new plugin cache, not the transform cache, so will be useless until the plugin cache approach is released and in use.

Tests are currently not passing yet, but builds are working making this suitable for testing.


frank-dspeed:
This is also importent for bigger projects i was just using rollup maybe as the first user ever with 9tb it was not working so i needed to workaround i runned it only on a partial set always via a external queue

this don't only needs disk cache it also needs maybe support for storing the ast in a query able database backend :D

Maruda1986:
.

melissamforbs (approved):

lukastaegert:
Closing as this is thoroughly outdated at this point, and caching approaches will likely go into a different direction.

PR #2404, opened 2018-08-13
pr template changes

guybedford:
I'd like to suggest some minor adjustments to the PR template.

* Making the height of the template slightly shorter to get to business asap.
* Encouraging breaking changes to be described in the main section
* Reduce category options by having an "other"
* Have a checklist for tests (yes is the only answer...)
* Include a section at the top to list relevant issues

I'm also wondering about the "lint" one - do we need this with the commit hook? @lukastaegert


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,20 +13,20 @@
...
-- [ ] refactor
-- [ ] tests
+- [ ] refactoring
 - [ ] documentation
-- [ ] metadata
-
-### Breaking Changes?
+- [ ] other
 
+Are tests included?
```
shellscape: There are valid reasons not to include tests. CI updates, config updates, README updates, etc. I understand the intent here, but it's not accurate.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,20 +13,20 @@
 -->
 
 This PR contains:
-
 - [ ] bugfix
 - [ ] feature
-- [ ] refactor
-- [ ] tests
+- [ ] refactoring
```
shellscape: I believe "refactor" is correct here, as the verb describes the state. The past form is "refactored," while the present participle "refactoring" describes something currently in process, the action of undertaking a refactor. While a PR could be considered in process, it's supposed to be a finished work ready for review and merging. Don't be to be pedantic about it, I just see these things and they stand out to me :smile:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,20 +13,20 @@
...
-- [ ] metadata
-
-### Breaking Changes?
+- [ ] other
 
+Are tests included?
 - [ ] yes
+
+Breaking Changes?
+- [ ] yes (please describe clearly below)
```
shellscape: since we're removing the heading, let's make `(please describe clearly below)` stand out some more by bolding it. emphasis here is a good thing.

.github/PULL_REQUEST_TEMPLATE.md
shellscape: _Only_ mentioning because English is fun (seriously, I get a kick out of this stuff). bugfix, feature, and refactor are all part of a class of words called "gerunds," or verbs that are used as nouns.

guybedford:
@lukastaegert one of the reasons I created this was actually exactly to describe what you are asking - to make it clear breaking changes should not get their own section but be included in the main description.  (which was my impression on first read of the original) I've altered the wording slightly here.

Let me know if you have any suggestions for clarification further.

guybedford:
I've also altered the PR description title - otherwise it seems a bit odd I think having all PRs reading "Please describe your changes" as their heading. Other suggestions welcome too.

PR #2409, opened 2018-08-14
1.0: deprecations

guybedford:
This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [x] tests
- [ ] documentation
- [ ] metadata

### Breaking Changes?

- [x] yes
- [ ] no

This is another PR for the 1.0 release only (DO NOT MERGE into master).

This removes old deprecations, and creates new deprecations in Rollup for some of the recent API changes as discussed.

It's currently based to the combination of both #2405 and #2406 to ensure both of those deprecation paths are included.

Removed deprecated features:
* options.load
* options.resolveId
* options.entry
* options.transform
* options.pureExternalModules
* options.moduleId
* Output options as intput options
* outputOptions.dest
* outputOptions.moduleId

New deprecations added:
* transformBundle
* transformChunk
* onwrite
* ongenerate
* transform dependencies
* this.watcher in plugins

All of the new deprecations have deprecation warnings and tests.


TrySound:
Transform is repeated twice

lukastaegert:
Not 100% sure about "output options as input options".
Pro: It is clear where everything belongs and what it has effects on
Con: Using output options as input options could become a nice way of creating defaults for all outputs. 

In any case, the current have-baked approach needed to be removed 👍The question is if we want to reintroduce it properly working for all options? Also maybe in a separate PR?

src/rollup/types.d.ts
```diff
@@ -89,8 +89,8 @@ export interface PluginCache {
 }
 
 export interface PluginContext {
-	// TODO deprecate:
-	watcher: Watcher;
+	/** @deprecated */
```
lukastaegert: 👍 Very good, works nicely in my IDE:
bildschirmfoto 2018-08-15 um 09 04 38

src/rollup/types.d.ts
```diff
@@ -241,17 +243,8 @@ export interface InputOptions {
...
-	preferConst?: boolean;
-	perf?: boolean;
-
 	/** @deprecated */
-	entry?: string;
-	transform?: TransformHook;
-	load?: LoadHook;
-	resolveId?: ResolveIdHook;
-	resolveExternal?: any;
+	preferConst?: boolean;
```
lukastaegert: I know this one is not documented but I am not sure it should be deprecated. Maybe it should be documented instead? `const` seems to have a slight performance benefit in certain environments while `var` has the universal compatibility.

src/utils/pluginDriver.ts
```diff
@@ -102,14 +126,14 @@ export function createPluginDriver(
 				if (typeof warning === 'string') warning = { message: warning };
 				if (warning.code) warning.pluginCode = warning.code;
 				warning.code = 'PLUGIN_WARNING';
-				warning.plugin = plugin.name || '(anonymous plugin)';
+				warning.plugin = plugin.name || `Plugin at position ${pidx + 1}`;
```
lukastaegert: 👍

src/utils/pluginDriver.ts
```diff
@@ -361,3 +395,11 @@ const uncacheablePlugin: (pluginName: string) => PluginCache = pluginName => ({
...
 	}
 });
+
+function hookDeprecationWarning(name: string, newName: string, plugin: Plugin, pidx: number) {
+	return {
+		code: name.toUpperCase() + '_HOOK_DEPRECATED',
+		message: `The ${name} hook used by plugin ${plugin.name ||
+			`at position ${pidx + 1}`} is deprecated. The ${newName} hook should be used instead.`
+	};
+}
```
lukastaegert: Might be nice to add a `.url` with a link to the upcoming plugin API section on the website once it is available.

lukastaegert: Similarly, we could add a link to the big list of options when an unknown option is passed: https://rollupjs.org/guide/en#big-list-of-options

src/rollup/types.d.ts
guybedford: Yes agreed.

guybedford:
> The question is if we want to reintroduce it properly working for all options? Also maybe in a separate PR?

It's good to consider this but note that we have been providing all users with deprecation warnings if not doing this, so that the moment those deprecation paths were created that decision was made to deprecate here as far as I'm concerned. That said, we can certainly reconsider this. Personally I think it's good to be explicit and stick with the model we've set.

src/utils/pluginDriver.ts
lukastaegert: 👍

Conduitry:
The changelog mentions that returning `dependencies` from the `transform` hook has been deprecated in favor of the `watchChange` hook, which doesn't seem accurate.

The actual deprecation message says to use `this.addWatchFile()`, which sounds like it would be better - but this function is not currently documented in the guide.

lukastaegert:
Thanks! I think things got a little messed up with the `watchChange` hook and the `addWatchFile` context function. Looking through these things now.

PR #2410, opened 2018-08-15
feat: skip stdin resuming in watch mode

kvet:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] tests
- [ ] documentation
- [ ] metadata

### Breaking Changes?

- [ ] yes
- [x] no

Solving the following problem: #1919

### DIFF
--- a/bin/src/run/watch.ts
+++ b/bin/src/run/watch.ts
@@ -148,7 +148,7 @@ export default function watch(
 					break;
 
 				case 'END':
-					if (!silent && isTTY) {
+					if (!silent) {
 						stderr(`\n[${dateTime()}] waiting for changes...`);
 					}
 			}
@@ -158,12 +158,6 @@ export default function watch(
 	// catch ctrl+c, kill, and uncaught errors
 	const removeOnExit = onExit(close);
 	process.on('uncaughtException', close);
-
-	// only listen to stdin if it is a pipe
-	if (!process.stdin.isTTY) {
-		process.stdin.on('end', close); // in case we ever support stdin!
-		process.stdin.resume();
-	}
 
 	function close(err: Error) {
 		removeOnExit();
--- a/bin/src/run/watch.ts
+++ b/bin/src/run/watch.ts
@@ -148,7 +148,7 @@ export default function watch(
 					break;
 
 				case 'END':
-					if (!silent) {
+					if (!silent && isTTY) {
 						stderr(`\n[${dateTime()}] waiting for changes...`);
 					}
 			}
@@ -158,6 +158,11 @@ export default function watch(
 	// catch ctrl+c, kill, and uncaught errors
 	const removeOnExit = onExit(close);
 	process.on('uncaughtException', close);
+
+	// only listen to stdin if it is a pipe
+	if (!process.stdin.isTTY) {
+		process.stdin.on('end', close); // in case we ever support stdin!
+	}
 
 	function close(err: Error) {
 		removeOnExit();


shellscape:
I think there are a few questions that need answering before this is merged:

- Do we really want to remove this otherwise valuable check for all environments?
- Would it be wiser path be to provide an override option to pass that check?
- The declaration for `isTTY` is `const isTTY = Boolean(process.stderr.isTTY);`. Can that instead be overridden by the environments needing it to be?

It would appear that it could be overridden:

```console
→ node
> process.stderr.isTTY
true
> process.stderr.isTTY = false
false
> process.stderr.isTTY
false
```

kvet:
It sounds strange. Why do I need to override something if I already run rollup in watch mode?

I think that the issue may be treated as a bug. Watch mode is not running correctly in different environments. So, I think a user should use normal mode when it is needed instead of relying on this check.

shellscape:
> I think that the issue may be treated as a bug. Watch mode is not running correctly in different environments.

That makes the assumption that all environments are equal. But that is not the case, and not a correct assumption.

> So, I think a user should use normal mode when it is needed instead of relying on this check.

There are differences in what the Node API will support between environments that are TTY and non-TTY. It's an important distinction when dealing with console output. This commit https://github.com/rollup/rollup/commit/0295c5b01695b978b4c8c16d51e8817eb4c78a08 implemented the check you are now seeking to revert, and from the code it looks like there was specific intent to that addition.

The fact of the matter here is that executing this in a non-TTY environment can have unpredictable consequences. There's no reason that every executable out there should test their environment for support API when the `isTTY` property is there for that purpose. As such, when running an executable in an environment that is not TTY, it's not reasonable to expect an executable to make that consideration. Rather, the environment should be altered, if it can support the same interfaces.

In my opinion (other maintainers may feel differently) the end goal of this PR would be better suited as a contribution to the Lerna project, to simulate TTY support. Or the suggested workarounds in local setups. I don't believe this is a good change that will benefit users as a whole; conversely, only a select niche of users, and will likely introduce a regression for other users.

bin/src/run/watch.ts
```diff
@@ -148,7 +148,7 @@ export default function watch(
 					break;
 
 				case 'END':
-					if (!silent && isTTY) {
+					if (!silent) {
```
guybedford: It seems this check was initially added exactly for TTY support - https://github.com/rollup/rollup/pull/1574.

Can you perhaps explain why you think this should be the other way around?

bin/src/run/watch.ts
```diff
@@ -159,12 +159,6 @@ export default function watch(
 	const removeOnExit = onExit(close);
 	process.on('uncaughtException', close);
 
-	// only listen to stdin if it is a pipe
-	if (!process.stdin.isTTY) {
-		process.stdin.on('end', close); // in case we ever support stdin!
-		process.stdin.resume();
-	}
```
guybedford: Is this a bug fix, independent of the above? Or are both changes needed for the lerna support?

guybedford:
Lerna should work out of the box definitely, would be great to follow what is needed here further.

bin/src/run/watch.ts
guybedford: Sorry, I mean this line was initially added exactly for Lerna support :)

guybedford: Exact link - https://github.com/rollup/rollup/pull/1574/files#diff-4a84b75586b4af29e839a9837aa7838eR79

kvet:
I found the original PR that broke my project: https://github.com/rollup/rollup/pull/1774. This causes the 'end' event to be triggered in lerna. So rollup stops watching changes.

In webpack a such check is in condition that becomes true when specific option is set: https://github.com/webpack/webpack-cli/blob/master/bin/cli.js#L516. But I didn't find this option in the webpack docs. However, the default value of this option is false.

guybedford:
@kvet thanks for tracking this down, this seems good to me.

As for the use case of https://github.com/rollup/rollup-watch/issues/30, surely that is just `persistent: false`? Or is there more at play?

Regardless I think we should merge this then come back to that.

Issue #2454, opened 2018-09-12
Optional catch binding syntax


bathos:
Support for [optional catch binding](https://github.com/tc39/proposal-optional-catch-binding) syntax can be enabled in Rollup by setting `acorn.ecmaVersion` to 2019, but Rollup itself seems to choke on the resulting AST.

I’m not sure whether I should be using the bug or feature request template given the nature of the issue; it depends on what Rollup aims to support, and I couldn’t turn up a statement one way or another. Optional catch binding hit stage 4 in May and is implemented in all major engines except Chakra (a PR is pending), and Acorn and other parsers support it, so it seemed like 'bug' might be reasonable, but please recategorize if it isn’t.

- Rollup Version: 0.65.2
- Operating System (or Browser): OSX 10.13.5, but n/a
- Node Version: 10.10.0

### How Do We Reproduce?

With `ecmaVersion` set to 2019, run rollup with an input file whose content is:

```js
try {} catch {}
```

### Expected Behavior

No error — or perhaps an error message like "optional catch binding is not supported".

### Actual Behavior

```
TypeError: Cannot read property 'declare' of null
    at CatchClause.initialise (.../rollup/dist/rollup.es.js:10350:20)
    at CatchClause.NodeBase (.../rollup/dist/rollup.es.js:9132:14)
    at new CatchClause (.../rollup/dist/rollup.es.js:10343:42)
    at NodeBase.parseNode (.../rollup/dist/rollup.es.js:9259:29)
    at new NodeBase (.../rollup/dist/rollup.es.js:9131:14)
    at Program.NodeBase.parseNode (.../rollup/dist/rollup.es.js:9254:27)
    at Program.NodeBase (.../rollup/dist/rollup.es.js:9131:14)
    at new Program (.../rollup/dist/rollup.es.js:11839:42)
    at Module.setSource (.../rollup/dist/rollup.es.js:13108:20)
    at .../rollup/dist/rollup.es.js:21732:20
```

PR #2455, opened 2018-09-12
Support optional catch binding (stage 4 / ES2019)

bathos:
As described in issue #2454, acorn supports optional catch bindings but Rollup unconditionally expected the parameter to exist.

This may be a terrible pull request — I’m not very familiar with TypeScript and I haven’t contributed before, so I may not have approached this correctly.

However build/test seems to be working correctly.

### DIFF
--- a/src/ast/nodes/CatchClause.ts
+++ b/src/ast/nodes/CatchClause.ts
@@ -8,7 +8,7 @@ import { PatternNode } from './shared/Pattern';
 
 export default class CatchClause extends NodeBase {
 	type: NodeType.tCatchClause;
-	param: PatternNode;
+	param: PatternNode | null;
 	body: BlockStatement;
 
 	scope: CatchScope;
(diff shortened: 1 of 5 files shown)


src/ast/nodes/CatchClause.ts
```diff
@@ -20,14 +20,15 @@ export default class CatchClause extends NodeBase {
...
+		}
 	}
 
 	parseNode(esTreeNode: GenericEsTreeNode) {
-		this.body = <BlockStatement>new this.context.nodeConstructors.BlockStatement(
-			esTreeNode.body,
-			this,
-			this.scope
+		this.body = <BlockStatement>(
+			new this.context.nodeConstructors.BlockStatement(esTreeNode.body, this, this.scope)
```
bathos: formatting of this line changed due to running `npm run lint`

bathos:
Thanks @lukastaegert! — updated accordingly.

PR #2470, opened 2018-09-21
Fix template string

styfle:
This PR contains:
- [x] bugfix

Are tests included?

Breaking Changes?
- [x] no

### Description

This error message should probably use a template string so that `pluginName` is not a literal but it's value is printed instead.


styfle:
Should I add one more assert to check the error message string includes the plugin name? https://github.com/rollup/rollup/blob/138033ee89254f7d692c2aee809988107b46e564/test/hooks/index.js#L846

shellscape:
@styfle nice catch! yes, please do add a test. definitely would not hurt.

styfle:
@shellscape Done!

shellscape:
thanks!

Issue #2461, opened 2018-09-17
Chunk imports are undefined


arxpoetica:
- Rollup Version: 0.66
- Operating System (or Browser): MacOS High Sierra 10.13.6 (17G65) / Google Chrome 68.0.3440.106 (Official Build) (64-bit)
- Node Version: 8.1.0

### How Do We Reproduce?

This is Sapper specific, so a little hard to reproduce without a Sapper setup. So I set up a temporary repo like so for install:

```bash
git clone https://github.com/arxpoetica/rollup-ticket.git
cd rollup-ticket
npm install
npm run dev
```

Then go to http://localhost:3000 and see the following:

### Expected Behavior

The CSS in the `<Popup.../>` component should show up (red background, white text).

### Actual Behavior

The CSS does not show up.

To get the CSS to show, do either of the following:

* In the main `routes/index.html` file, remove or comment out line 10: `// import debounce from './debounce.js'`

or

* In the `routes/alt.html` file, remove or comment out lines 3 and 8: `<!-- <Popup message="This is a block of text. Notice, the text always shows up, but not always the CSS styles."/> -->` and `// Popup: '../components/Popup.html'`

FYI @Rich-Harris in the Sapper community forums had this to day:

> essentially the bug is that Rollup doesn't correctly report the dependencies of each chunk, in certain cases where they are required by multiple chunks, which means Sapper fails to generate correct CSS

Rich-Harris:
I think we can probably turn this into a repo that doesn't involve Svelte or Sapper — will try and have a crack at it soon if @arxpoetica doesn't beat me to it

arxpoetica:
@Rich-Harris I probably won't have time in the immediate future -- sorry for the Sapper-only orientation.

matt3224:
Following on from our discord chat i thought id add the something here with a link to my project https://github.com/matt3224/blog in-case that helps at all

Rich-Harris:
I've whittled the bug down to its essence: https://github.com/Rich-Harris/rollup-repro-2461

It looks like what's happening is that a chunk is registering another chunk as an import *before* the imported chunk has been given a name. Given modules like this...

```js
// main.js
export default [
  () => import('./a.js'),
  () => import('./b.js')
];

// a.js
import './d.js'
import c from './c.js';

export default () => c();

// b.js
import c from './c.js';

export default () => c();

// c.js
export default () => console.log('c');

// d.js
export default {};
```

[...]

PR #2483, opened 2018-09-29
Assign chunk IDs before creating outputBundle chunks

Rich-Harris:
This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

#2461 

### Description

The `chunk` object that is passed to the `renderChunk` hook (and which forms the properties of the output bundle object) contains, inter alia, an `imports` array specifying the IDs of the external modules and chunks the chunk depends on.

Currently, the loop that assigns the IDs to the chunk is the same as the loop that gets the imported IDs; consequently the imported IDs are frequently `undefined`.

This PR separates ID assignation into a separate loop.

### DIFF
--- a/src/rollup/index.ts
+++ b/src/rollup/index.ts
@@ -284,9 +284,6 @@ export default function rollup(
 							// name all chunks
 							for (let i = 0; i < chunks.length; i++) {
 								const chunk = chunks[i];
-								const imports = chunk.getImportIds();
-								const exports = chunk.getExportNames();
-								const modules = chunk.renderedModules;
 
 								if (chunk === singleChunk) {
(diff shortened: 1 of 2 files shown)


guybedford:
The CI failure seems to do with the assignment of numbers to named chunks with the same chunk name - I checked it out and it seems fine to update the tests to what is being output, and right as well that the first chunk (`1`) is the one being named 1.

Rich-Harris:
Huh, it looks like it's only generating one chunk instead of three in that test? Investigating...

PR #2607, opened 2018-12-21
Hoisting fixes (#2606)

lye:
When hoisting a var, use UNKNOWN_EXPRESSION as the initializer instead
of the actual one. This prevents the initializer from being used during
the optimization step when it hasn't yet been assigned at runtime.

This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

#2606 

### Description

The bug is described in #2606. I more-or-less copied the examples as new test cases.


nifgraup:
Confirming 5739cbd3e36f7bf885153bff17ecddc507b84421 fixes the test case in#2606. (and that this test https://github.com/terser-js/terser/blob/master/test/compress/issue-2871.js passes when terser is bundled with rollup).

PR #3049, opened 2019-08-11
Fix a warning of unresolved dependencies in an interop test

giraffate:
This PR contains:
- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

I fixed a test to suppress a warning.

Breaking Changes?
- [x] no

### Description
Fix a warning of unresolved dependencies in an interop test. Following are before and after test results.

Before
```
main.js → stdout...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
turbocolor (imported by main.js)
created stdout in 23ms

✓ interop: does not include the interop block (354ms)

```

After
```
main.js → stdout...
created stdout in 26ms

✓ interop: does not include the interop block (371ms)
```

### DIFF
--- a/test/cli/samples/interop/_config.js
+++ b/test/cli/samples/interop/_config.js
@@ -1,4 +1,4 @@
 module.exports = {
 	description: 'does not include the interop block',
-	command: 'rollup -i main.js -f cjs --no-interop'
+	command: 'rollup -i main.js -f cjs --no-interop --external turbocolor'
 };


shellscape:
Thanks for the PR 🍺  The reason that the checks for the PR failed are because the unresolved warning was actually intentional and provided coverage over a branch that produced that warning. Your fix is technically correct, but it prevents that warning from being displayed :)

Issue #3225, opened 2019-11-11
Decouple write (to disk) from rollup.watch


jdanyow:
### Feature Use Case

As a dev tooling author I'm using the rollup javascript API to generate bundles in memory. The rollup.rollup API works great for this use case because it separates bundle generation from bundle writing. Would it be possible to make the rollup.watch API match? Maybe if write options are not passed this line could be skipped:

https://github.com/rollup/rollup/blob/8318bb42ff91c510f740af07d8e41c4b3bd99918/src/watch/index.ts#L213

### Feature Proposal

Provide a way to configure rollup.watch such that it does not write the generated bundles to disk.

shellscape:
Thanks for opening an issue. Could you link to your project or any relevant source that would provide context for the request? As this would be a breaking change, maximum information is important.

jdanyow:
Sure, thanks for considering this. I can't link to the repo because it's private but here's the code.

This class is part of a system that bundles chai tests and executes them in headless chrome (puppeteer) or in cloudflare workers. The bundle never needs to be written to disk because it's served from memory with the http module or posted to the cloudflare api to enable testing in the codes true target environment (cloudflare worker).

Specifically, in the `async bundle() {` method below, by the time the watcher's BUNDLE_END event has fired the bundle has been generated and written to disk. I would like to be able to generate myself and avoid the write to disk because the file isn't needed and causes collisions / requires cleanup.

```js
// @ts-check

const EventEmitter = require('events');
const { rollup, watch } = require('rollup');
const glob = require('fast-glob');
const typescript = require('rollup-plugin-typescript2');
const virtual = require('rollup-plugin-virtual');
// const resolve = require('rollup-plugin-node-resolve');
// const commonjs = require('rollup-plugin-commonjs');


/**
 * Bundles specs into a single file using rollup
 */
class Bundler extends EventEmitter {
  /** @type {import('rollup').OutputOptions} */
  outputOptions = { file: 'temp.js', format: 'iife', globals: { 'chai': 'chai' } };

  /**
   * @param {string[]} globs 
   * @param {boolean} watch 
   */
  constructor(globs, watch) {
    super();
    this.globs = globs;
    this.watch = watch;
    this.code = '';
    this.bundled = new Promise(resolve => this.setBundled = resolve);
  }

  async bundle() {
    const inputOptions = await this.getInputOptions();

    if (this.watch) {
      /** @type {import('rollup').RollupWatchOptions} */
      const watchOptions = {
[...]

PR #3454, opened 2020-03-21
feat(watch): add `skipWrite` option

lukeed:
This PR contains:
- [ ] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

List any relevant issue numbers: #3225

### Description

This change allows programmatic `rollup.watch` users to opt _out_ of automatic `bundle.write` calls. This definitely won't be the common use case, but it's been requested (a few times) in the past for certain scenarios.


docs/01-command-line-reference.md
```diff
@@ -272,20 +273,20 @@ Print the help document.
 Use the specified plugin. There are several ways to specify plugins here:
 
 - Via a relative path:
-  
+
   ```
   rollup -i input.js -f es -p ./my-plugin.js
   ```
-  
+
```
lukeed: Sorry, these look to be my editor's auto-trimming

docs/999-big-list-of-options.md
```diff
@@ -1096,6 +1096,13 @@ Default: `true`
 
 Whether to clear the screen when a rebuild is triggered.
 
+#### watch.skipWrite
+Type: `boolean`<br>
+<!-- CLI: `--watch.skipWrite`<br> -->
```
lukeed: This CLI option _will_ work, but I figure it's better not to list it (similar to `watch.chokidar`) since its application makes zero sense via the CLI layer.

src/watch/watch.ts
```diff
@@ -174,7 +176,7 @@ export class Task {
 				return;
 			}
 			this.updateWatchedFiles(result);
-			await Promise.all(this.outputs.map(output => result.write(output)));
+			this.skipWrite || (await Promise.all(this.outputs.map(output => result.write(output))));
```
lukeed: This is where the new option is applied.

(Not a fan of how the linter formatted it)

timiyay:
Nice. This `skipWrite` option will give me a way to deal with Rollup challenge I'm currently facing.

I need to get my hands on the return value from `rollup.write` when running in watch mode. To do this, I have to listen for `BUNDLE_END` and call `event.result.write`. This means I write the bundle to disk twice in quick succession, which feels dirty.

`skipWrite` would let me take responsibility for calling `write()` within my event listener, giving me my return value while avoiding the double write.

Is there anything I can do to help get this ready for release?

lukeed:
Thanks!

PR #3617, opened 2020-06-05
Update sourcemapPathTransform to also take the path to the sourcemap …

dgoldstein0:
…file as a second argument

This PR contains:
- [x] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

List any relevant issue numbers: #3615

### Description

This aims to solve the problem in #3615.  I need some help figuring out the best way to write the test for this.  I manually tested by making the same change to rollup in my project and verified I could generate sourcemaps correctly by utilizing this change.


test/function/samples/transform-source-function-gets-two-args/_config.js
```diff
@@ -0,0 +1,18 @@
...
+			name: 'myModule',
+			sourcemap: true,
+			file: path.resolve(__dirname, 'main.js'),
+			sourcemapPathTransform: (relativePath, sourcemapPath) => path.resolve(sourcemapPath, relativePath)
+		}
+	},
+	generateError: {
+		code: 'VALIDATION_ERROR',
+		message: 'sourcemapPathTransform function must return a string.'
+	}
```
dgoldstein0, line 17: this section is wrong but I didn't find a way to assert about the sourcemap output or even just the js output... pointers here would be helpful.

lukastaegert: Why don't you just extend this one to use the new parameter: https://github.com/rollup/rollup/tree/master/test/sourcemaps/samples/transform-source-paths ? Alternatively, you can put a modified one next to it.

docs/999-big-list-of-options.md
```diff
@@ -670,7 +670,7 @@ The location of the generated bundle. If this is an absolute path, all the `sour
 `sourcemapFile` is not required if `output` is specified, in which case an output filename will be inferred by adding ".map"  to the output filename for the bundle.
 
 #### output.sourcemapPathTransform
-Type: `(sourcePath: string) => string`
+Type: `(sourcePath: string, sourcemapPath: string) => string`
 
 A transformation to apply to each path in a sourcemap. For instance the following will change all paths to be relative to the `src` directory.
```
lukastaegert, line 675: I believe a short explanation of what `sourcePath` is as opposed to `sourcemapPath` here would help avoid a lot of confusion.

dgoldstein0:
No I don't mind.  I was planning on following up tomorrow but I'm happy to
not need to do that.  I'll still glance over your changes to get an idea of
what I can do better for any future PRs.

PR #3628, opened 2020-06-11
Acorn 7.3.0 upgrade

guybedford:
Acorn [just released 7.3.0 with optional chaining support](https://github.com/acornjs/acorn/pull/891#issuecomment-642478745) so this PR switches back from the fork we were using to provide this.

I did check that the final ESTree changes don't affect any of the internal code here and as far as I can tell there aren't any issues with this currently, but let me know if I've missed anything.

This PR contains:
- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [x] no

Breaking Changes?
- [x] no


guybedford:
Because the build is performed based on the previous build I've left in the fork, perhaps a way to do this will be to remove the fork in a subsequent PR once this is merged?

chicoxyzzy:
7.3.1 was released a bit later todey

guybedford:
I've also added a commit here to upgrade the plugins, as there was a fix in the private class elements package to work with optional chaining here - https://github.com/acornjs/acorn-private-class-elements/commit/c9857bb69ca3deb4cdd22339174897beb0af4c52.

I also removed the import meta plugin since that is in Acorn core since https://github.com/acornjs/acorn/pull/943.

PR #3633, opened 2020-06-13
Turn namespace assignment error into a warning

guybedford:
Similarly to https://github.com/rollup/rollup/pull/3475 this turns the static error into a warning that will become a runtime error only on the execution paths that actually attempt it.

This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### DIFF
--- a/src/ast/nodes/MemberExpression.ts
+++ b/src/ast/nodes/MemberExpression.ts
@@ -262,7 +262,8 @@ export default class MemberExpression extends NodeBase implements DeoptimizableE
 			this.object instanceof Identifier &&
 			this.scope.findVariable(this.object.name).isNamespace
 		) {
-			return this.context.error(
+			this.scope.findVariable(this.object.name).include();
+			this.context.warn(
 				{
 					code: 'ILLEGAL_NAMESPACE_REASSIGNMENT',
 					message: `Illegal reassignment to import '${this.object.name}'`
--- a/src/ast/nodes/MemberExpression.ts
+++ b/src/ast/nodes/MemberExpression.ts
@@ -258,18 +258,18 @@ export default class MemberExpression extends NodeBase implements DeoptimizableE
 	}
 
 	private disallowNamespaceReassignment() {
-		if (
-			this.object instanceof Identifier &&
-			this.scope.findVariable(this.object.name).isNamespace
-		) {
-			this.scope.findVariable(this.object.name).include();
-			this.context.warn(
-				{
-					code: 'ILLEGAL_NAMESPACE_REASSIGNMENT',
-					message: `Illegal reassignment to import '${this.object.name}'`
-				},
-				this.start
-			);
+		if (this.object instanceof Identifier) {
+			const variable = this.scope.findVariable(this.object.name);
+			if (variable.isNamespace) {
+				variable.include();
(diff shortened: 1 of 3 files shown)


src/ast/nodes/MemberExpression.ts
```diff
@@ -262,7 +262,8 @@ export default class MemberExpression extends NodeBase implements DeoptimizableE
 			this.object instanceof Identifier &&
 			this.scope.findVariable(this.object.name).isNamespace
```
lukastaegert, line 263: For style points, you could split this condition and assign the return value of `this.scope.findVariable` to a local variable to avoid the double lookup.

guybedford: Sure, I've added the refactoring.

lukastaegert:
I know this is already merged but what was the reason to `.inlcude()` the namespace on reassignment? I mean, this leads to the creation of a namespace object, but this is not what is reassigned. The original variable is what is reassigned, wouldn't it be enough to include that?

https://rollupjs.org/repl/?version=2.16.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMColMjBhcyUyMGxpYiUyMGZyb20lMjAnLiUyRmxpYiclM0IlNUNubGliLmZvbyUyMCUzRCUyMCd0aGlzJTIwd2lsbCUyMGJlJTIwYXNzaWduZWQlMjB0byUyMGZvbyclM0IlNUNubGliLmJhciUyMCUzRCUyMCd0aGlzJTIwd2lsbCUyMGJlJTIwYXNzaWduZWQlMjB0byUyMHVuZGVmaW5lZCclM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJsaWIuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBmb28lMjAlM0QlMjAnZm9vJyUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE

guybedford:
Without the include the entire statement was being removed as it wasn’t
being detected as having a side effect of throwing. I don’t know what
mechanism is best to handle possible throw side effects currently.

lukastaegert:
In that case I would suggest #3637

PR #3775, opened 2020-09-12
add ability to specify stdin file extension via --stdin=ext

kzc:
This PR contains:
- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

Addresses missing functionality mentioned in #3312.

### Description

Presently rollup can only process `esm` via stdin. This PR adds the ability to specify the CLI stdin file extension via `--stdin=ext` allowing other inputs such as JSON:
```
echo '{"foo": 42, "bar": "ok"}' | rollup --stdin=json -p json
``` 
A synthetic file name `"-.ext"` is used internally to represent the stdin filename with an extension in order to work with existing plugins with suffix-based logic. Stdin based input can still be disabled with `--no-stdin` if required.

Unfortunately this new stdin extension feature will not work with plugins that interrogate the filesystem to read input files such as `@rollup/plugin-typescript`. For this scenario other plugins like `@rollup/plugin-sucrase` or `rollup-plugin-esbuild` could be used instead.


kzc:
The code coverage failure seems rather insignificant - a partial miss in one `if` condition.

@lukastaegert Feel free to take over this PR to change its testing and implementation as you see fit. It works fine my own use. I think others could benefit from this new functionality as well.

kzc:
The code was simplified and coverage now passes.

lukastaegert (approved):
Thanks, looks good!

PR #3794, opened 2020-09-22
fix(types): Allow methods to return void in addition to undefined

Alorel:
This PR contains:
- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [x] no

Breaking Changes?
- [x] no

### Description

Many plugin methods that are allowed to return null or undefined are forced to explicitly return one or the other. For example, the following two variations of the `outputOptions` hook are functionally identical, but one causes a typings error while the other doesn't:

image

image

This PR adds a consistent type alias for `null | undefined | void` and updates relevant `null | undefined` instances with the new alias.

### DIFF
--- a/src/rollup/types.d.ts
+++ b/src/rollup/types.d.ts
@@ -221,7 +221,7 @@ interface PartialResolvedId {
 	syntheticNamedExports?: boolean | string;
 }
 
-export type ResolveIdResult = string | false | null | undefined | PartialResolvedId;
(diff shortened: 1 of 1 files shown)


Alorel:
I will close this. It appears that Typescript doesn't accept `void` for the nullish coalescing operator and this isn't an important enough change to warrant a breaking release.

PR #3824, opened 2020-10-16
fix: plugin warnings not showing warning.loc

luciotato:
Plugin warnings without warning.id but with warning.loc were not emitting loc information
loc information emitting is fundamental for user experience (to locate the source of the error/warning)

This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

### Description

Plugin warnings without `warning.id `but with `warning.loc` were not emitting `warning.loc` information
`warning.loc` information emitting is fundamental for user experience (to locate the source of the error/warning)
added 3 lines, to emit the `warning.loc` info if present in a format amicable to IDEs problem matchers: file:line:col

### DIFF
--- a/cli/run/batchWarnings.ts
+++ b/cli/run/batchWarnings.ts
@@ -189,15 +189,13 @@ const deferredHandlers: {
 				for (const warning of items) {
 					if (warning.url && warning.url !== lastUrl) info((lastUrl = warning.url));
 
-					if (warning.id) {
-						let loc = relativeId(warning.id);
+					const id = warning.id || warning.loc?.file;
+					if (id) {
+						let loc = relativeId(id);
 						if (warning.loc) {
 							loc += `: (${warning.loc.line}:${warning.loc.column})`;
 						}
 						stderr(bold(loc));
-					}
-					else if (warning.loc) {
-							info(yellow(`${rollup.relativeId(warning.loc.file)}:${warning.loc.line}:${warning.loc.column}`)+' error '+message);
 					}
 					if (warning.frame) info(warning.frame);
 				}
--- a/cli/run/batchWarnings.ts
+++ b/cli/run/batchWarnings.ts
@@ -196,6 +196,9 @@ const deferredHandlers: {
 						}
 						stderr(bold(loc));
 					}
+					else if (warning.loc) {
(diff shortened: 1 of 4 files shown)


cli/run/batchWarnings.ts
```diff
@@ -196,6 +196,9 @@ const deferredHandlers: {
 						}
 						stderr(bold(loc));
 					}
+					else if (warning.loc) {
+							info(yellow(`${rollup.relativeId(warning.loc.file)}:${warning.loc.line}:${warning.loc.column}`)+' error '+message);
+					}
```
lukastaegert, line 201: I'm. not happy about this being the only non-title being formatted in yellow, couldn't we stick with the formatting used in the previous case? Also why do you want to repeat the error message that has just been written in the title? Moreover, `file` is optional in `.loc`, and this only makes sense if `file` is actually present otherwise you would not know what the location means.

Also this repeats the way that locations are formatted from the previous section, but in a slightly inconsistent way with regard to spacing. Here would be one suggestion to replace everything starting from line 192 with less duplication:

```js
const id = warning.id || warning.loc?.file;
if (id) {
  let loc = relativeId(id);
  if (warning.loc) {
    loc += `: (${warning.loc.line}:${warning.loc.column})`;
  }
  stderr(bold(loc));
}
```

lukastaegert (changes requested):
This looks like a useful improvement to me. However, some work would be needed to get this ready for release, see my comment.

We would also need tests in order for this to be merged. And there are two different cases to be considered, one including `loc.file` and one where it is missing. These could be combined into a single test that emits two warnings, though. You could take for instance `test/cli/samples/custom-frame` as an inspiration.
Please let me know if you would be willing to go through with this.

cli/run/batchWarnings.ts
luciotato: Lukas, I changed the code as you suggested.
I won't be able to add tests.

lukastaegert:
I can push a test to your branch, but you would need to "allow edits by maintainer", which should be a checkbox somewhere to the bottom right of your PR:
Bildschirmfoto 2020-10-21 um 06 38 14

luciotato:
done.

PR #3848, opened 2020-10-30
[types] add options hook to AsyncPluginHooks

intrnl:
This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [x] no

Breaking Changes?
- [x] no

### Description

According to the docs, options hook is supposed to be [async](http://rollupjs.org/guide/en/#options), and I checked the [code](https://github.com/rollup/rollup/blob/1fa9374/src/rollup/rollup.ts#L101) as well.

This pull request fixes the corresponding types.

### DIFF
--- a/src/rollup/types.d.ts
+++ b/src/rollup/types.d.ts
@@ -347,7 +347,10 @@ export interface PluginHooks extends OutputPluginHooks {
 	buildStart: (this: PluginContext, options: NormalizedInputOptions) => Promise<void> | void;
 	load: LoadHook;
 	moduleParsed: ModuleParsedHook;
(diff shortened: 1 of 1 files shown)


intrnl:
Hmm, there's something I noticed,

If the `await inputOptions` in this [line](https://github.com/rollup/rollup/blob/1fa9374/src/rollup/rollup.ts#L116) resolves to `null`, then it wouldn't follow what the docs says,
> Returning null does not replace anything

lukastaegert (approved):
Good catch

PR #3936, opened 2021-01-25
Ensure test after() callback is always executed

Benjamin-Dobell:
This PR contains:
- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
**N/A** - The fix is for the test suite itself.

Breaking Changes?
- [x] no

### Description

Took me a little while to track down what was going on here. Basically, if the `emit-uint8array-no-buffer` test fails, then all subsequent tests will fail to execute. This is because the `emit-uint8array-no-buffer` test has a `before` function which deletes `global.Buffer`, and and an `after` function which is supposed to restore it.

However, prior to this PR, the `after` callback is not guaranteed to be called so `global.Buffer` remains undefined. The lack of `global.Buffer` causes issues with, [magic-string](https://github.com/Rich-Harris/magic-string/blob/389c1f1688b5f58db6589601f57b901a2aa43ce1/src/SourceMap.js#L9), but likely elsewhere too.

https://github.com/rollup/rollup/blob/889fa5267ad93cc706fdbff2024e9c4d4f273749/test/form/samples/emit-uint8array-no-buffer/_config.js#L6-L11


lukastaegert (approved):
Good catch!

PR #3937, opened 2021-01-25
Modernize references to Node.js and webpack

ludofischer:
This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?
- [x] no

Breaking Changes?
- [x] no

### Description
To avoid confusing novices, update  Node.js and webpack references.

* Remove mention of experimental flag in Node.js since last two LTS versions
  support ESM without it
* Change webpack 2 to just webpack since 3,4 and 5 also support ESM and webpack 2 docs are not on the webpack
  website

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ rollup main.js --format umd --name "myBundle" --file bundle.js
 
 Developing software is usually easier if you break your project into smaller separate pieces, since that often removes unexpected interactions and dramatically reduces the complexity of the problems you'll need to solve, and simply writing smaller projects in the first place [isn't necessarily the answer](https://medium.com/@Rich_Harris/small-modules-it-s-not-quite-that-simple-3ca532d65de4). Unfortunately, JavaScript has not historically included this capability as a core feature in the language.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,7 +65,7 @@ rollup main.js --format umd --name "myBundle" --file bundle.js
...
+This finally changed with ES modules support in JavaScript, which provides a syntax for importing and exporting functions and data so they can be shared between separate scripts. The specification is now implemented in all major browsers and in Node.js. Rollup allows you to write your code using this module system, either outputting optimized ES modules for use in these native environments, or compiling it back down to existing supported formats such as CommonJS modules, AMD modules, and IIFE-style scripts. This means that you get to *write future-proof code*, and you also get the tremendous benefits of...
```
shellscape, line 68: The removal of the flag is only true for Node 14. Since Rollup supports LTS, this should stay in, or be modified to state that the flag may be required for older versions.

dasa: This flag was also removed in 12.17.0: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2020-05-26-version-12170-erbium-lts-targos

shellscape: Right, but not 12.x prior to that and not 10.x both of which are still in LTS https://nodejs.org/en/about/releases/ We won't be able to remove that bit until April 2022. So, it should be reinstated or reworded.

ludofischer: I have updated the Node.js version range, and I could not resist changing some of the surrounding text. I have tried to clarify why compiling to other formats is a feature. I have tried to make the problem -> solution structure more explicit, give concrete examples of environments that do not support ES modules. I have also tried to use to the same word (ES modules) instead of alternating (ES modules - this syntax - the specification).

lukastaegert (approved):
Thanks!

PR #3972, opened 2021-02-23
Don't allow `:` in file names.

NfNitLoop:
This PR contains:
- [X] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [X] yes (*bugfixes and features will not be merged without tests*)
- [ ] no

Breaking Changes?
- [X] no

### Description

File paths with a colon in them can refer to an NTFS "alternate data
stream" which does not operate like a normal file and can break many use
cases. They should be avoided (except for the case of the Windows drive
letter, like "C:").

This reverts some of the test changes made in
87586dfb41df57de791923ac6d8285b66607fca2 which removed the `:` from
the virtual entry names. This seems to be a convention in some plugins.
The test now tests that those names don't result in invalid Windows file
paths.

I originally uncovered this in my own project, and it's blocking me from properly bundling Rollup'd code (via Snowpack) in my Windows build:  https://github.com/NfNitLoop/feoblog/issues/16


lukastaegert (approved):
Looks good

PR #4012, opened 2021-03-24
fix `sourcemap` reference in docs

tjenkinson:
This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?
- [x] no

Breaking Changes?
- [x] no

### Description
It's lowercase in the config.

>Usually, a plugin will only omit the sourcemap if it (the plugin, not the bundle) was configured with `sourcemap: false`

Looks like the [commmonjs plugin uses `sourceMap`](https://github.com/rollup/plugins/tree/master/packages/commonjs#sourcemap) 🙈, so not sure if the line about plugins should be left as is, or maybe CommonJS should also accept `sourcemap` to be more inline with rollup itself?

### DIFF
--- a/docs/08-troubleshooting.md
+++ b/docs/08-troubleshooting.md
@@ -64,9 +64,9 @@ There are occasional valid reasons for `this` to mean something else. If you're 
 
 ### Warning: "Sourcemap is likely to be incorrect"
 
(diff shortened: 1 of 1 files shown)


lukastaegert (approved):
Thanks!

PR #4020, opened 2021-03-27
Add explanation for "iife"

DavidBruant:
This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

### Context

I was reading the rollup doc with a beginner in JavaScript 
and i realized that the meaning of "iife" was not explained anywhere 

### Description

I'm adding the meaning in a place that i feel makes sense in the documentation
This is only my best guess

I'm happy to add the explanation elsewhere and write a different text if the one i proposed is not satisfying


lukastaegert (approved):
Good addition!

PR #4077, opened 2021-05-12
Add assetExtname replacement in entryFileNames

BPScott:
This PR contains:
- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?
- [ ] no

Breaking Changes?
- [x] no

List any relevant issue numbers: #3107 #3116

### Description

When using `preserveModules` by default rollup adds the original extension of non-js/ts files to the generated filename. E.g. `foo.txt` compiles to `foo.txt.js` (so that it doesn't clash with `foo.js`). This behaviour was added in #3116. This is great, however it is a little tricky to emulate when customising the output filename with `entryFileNames`.

If I wanted the same-as-default behaviour but outputting `.mjs` extensions I could pass a function to entryFilenames but it is pretty dang clunky:

```js
{
  // ...
  preserveModules: true,
  entryFilenames: (chunkInfo) => {
    const NonAssetExtensions = ['.js', '.jsx', '.ts', '.tsx'];    
    const isAssetfile = !NonAssetExtensions.some((nonAssetExt) =>
      (chunkInfo.facadeModuleId || '').endsWith(nonAssetExt),
    );

    return `[name]${isAssetfile ? '[extname]' : ''}.mjs`;
  };
```

Instead I would like to propose adding an `assetExtname` replacement, that returns the extension prefixed with a `.` (same as extname) if the extension is not one of the "non asset extensions".

With this change, having rollup's default behaviour but outputting `.mjs` filenames can be done by setting entryFilenames to `'[name][assetExtname].mjs'`.


lukastaegert (approved):
Looks good, makes sense to me. Bonus points for updating the documentation. Will need to wait for GitHub Actions to work properly again before merging, though.

BPScott:
Pushed a change to fix the linting error

PR #4086, opened 2021-05-23
package.json: update exports for node 16

jaeh:
This PR contains:
- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?
- [x] no

Breaking Changes?
- [ ] no
- [x] maybe 

### Description

node 16 shows a deprecationWarning when using ./dist/ as exports field:

```
DeprecationWarning: Use of deprecated folder mapping "./dist/" in the "exports" field module resolution 
of the package at /{cwd}/node_modules/rollup/package.json imported from /{cwd}/build/tasks/rollup.js.
Update this package.json to use a subpath pattern like "./dist/*".
```

node 12, 13 and 14 seem to work fine with the change, node 14 works fine in my codebase that has node >= 14 as requirement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,6 @@
       },
       "default": "./dist/es/rollup.browser.js"
     },
-    "./dist/": "./dist/"
+    "./dist/*": "./dist/*"
   }
 }


guybedford (commented):
Out of interest, which entry point are you using that you need to use the `dist/` form at all here? I was hoping we could deprecate this in the next major so we don't expose private chunks unnecessarily.

jaeh:
i actually did 
`import * as rollup from  'rollup/dist/es/rollup.js'`
in my build task,
just changed the line to
`import * as rollup from 'rollup'`
and now all is fine, no warnings in node 16.

i wrote this line sometime last summer, i am sure i had a reason for it back then,
but i can not remember now, maybe i did not have `"type": "module"` in package.json and added it afterwards.

my build task works on node 14 and node 16 with the changed import and the ./dist key removed,
deprecating this export would maybe force some other people to change the import path too,
but i doubt there will be any problems other than that.

feel free to close this pull request :)

guybedford:
Thanks for the update, yes I would rather move to removal with the warning as a useful deterrent in the mean time! If you're happy to close we can do that. Otherwise if we can use explicit subpaths based on use cases that would be good too.

PR #4191, opened 2021-07-27
Fix incorrect watcher event documentation

plmrry:
The watcher event has an `output` property, not an `outputFiles` property.

This PR contains:
- [x] documentation

Are tests included?
- [x] no

Breaking Changes?
- [x] no

### Description
The watcher event has an `output` property, not an `outputFiles` property.


lukastaegert (approved):
Thanks!

PR #4220, opened 2021-09-03
chore: use forceConsistentCasingInFileNames in ts-config

dnalborczyk:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

just for good house keeping. recommended setting.

https://www.typescriptlang.org/tsconfig/#forceConsistentCasingInFileNames

> TypeScript follows the case sensitivity rules of the file system it’s running on. This can be problematic if some developers are working in a case-sensitive file system and others aren’t. If a file attempts to import fileManager.ts by specifying ./FileManager.ts the file will be found in a case-insensitive file system, but not on a case-sensitive file system.

> When this option is set, TypeScript will issue an error if a program tries to include a file by a casing different from the casing on disk.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,6 +2,7 @@
 	"compilerOptions": {
 		"allowSyntheticDefaultImports": true,
 		"diagnostics": true,
+		"forceConsistentCasingInFileNames": true,
 		"isolatedModules": true,
 		"module": "ESNext",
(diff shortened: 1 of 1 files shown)


lukastaegert (approved):
👍

PR #4265, opened 2021-11-05
support loading ESM plugins from the CLI via `--plugin`

kzc:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

n/a

### Description

Rollup presently only supports CJS plugins from the CLI via `--plugin`. This PR adds seamless ESM plugin support to the CLI. To maintain backwards compatibility rollup will still attempt to resolve the plugin via `require` first, and only upon error will it attempt to load the plugin via a dynamic `import()`. This is due to the different module resolution semantics of CJS and ESM modules in NodeJS.

The use of dynamic import to load ESM plugins works with Node v12+.  No rollup CLI functionality is lost on NodeJS versions that do not support loading ESM via dynamic `import`.

### DIFF
--- a/build-plugins/esm-dynamic-import.ts
+++ b/build-plugins/esm-dynamic-import.ts
@@ -13,7 +13,7 @@ export default function addBinShebangAndEsmImport(): Plugin {
 		name: 'esm-dynamic-import',
 		renderDynamicImport({ moduleId }) {
 			importFound = true;
-			if (moduleId.endsWith('loadConfigFile.ts')) {
+			if (moduleId.endsWith('commandPlugins.ts') || moduleId.endsWith('loadConfigFile.ts')) {
 				return { left: 'import(', right: ')' };
 			}
 		}
--- a/cli/run/commandPlugins.ts
+++ b/cli/run/commandPlugins.ts
@@ -3,40 +3,42 @@ import { InputOptions } from '../../src/rollup/types';
 import { stdinPlugin } from './stdin';
 import { waitForInputPlugin } from './waitForInput';
 
-export function addCommandPluginsToInputOptions(
+export async function addCommandPluginsToInputOptions(
 	inputOptions: InputOptions,
 	command: Record<string, unknown>
-): void {
+): Promise<void> {
 	if (command.stdin !== false) {
 		inputOptions.plugins!.push(stdinPlugin(command.stdin));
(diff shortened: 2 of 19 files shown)


kzc:
Interesting... the new ESM CLI plugin tests succeed on Node v12 here, but not on my local machine. The Node v12 version used by github CI must be more recent than my local version. 

The Node v10 failures for the new ESM CLI plugin tests were expected - both here and on my local machine. I will add a condition to skip the ESM CLI plugin tests on Node v10, as no old functionality is lost on that Node version.

lukastaegert (approved):
Looks good to me, thanks a lot!

dnalborczyk:
> Interesting... the new ESM CLI plugin tests succeed on Node v12 here, but not on my local machine. The Node v12 version used by github CI must be more recent than my local version.
> 
> The Node v10 failures for the new ESM CLI plugin tests were expected - both here and on my local machine. I will add a condition to skip the ESM CLI plugin tests on Node v10, as no old functionality is lost on that Node version.

the `experimental-modules` flag was removed in `v12.17`, and esm was considered stable in `v12.22`.

dnalborczyk:
I think `rollup` should probably make the move to full `esm` for `v3` as well and move the ecosystem forward! it should be perfectly fine when used from the cli as esm is supported from node.js `12.22+`. some friction could potentially arise when used in code, but that can be mitigated by using dynamic imports in cjs if needed.

kzc:
Even if rollup v3 moves to an all ESM toolchain for the CLI, rollup v2 should support both ESM and CJS during the transition period.  Although it's technically possible to mimic the semantics of `require` with dynamic `import()`, it's non-trivial. `require` module resolution has a lot of peculiar historical behaviors.

dnalborczyk:
> Even if rollup v3 moves to an all ESM toolchain for the CLI, rollup v2 should support both ESM and CJS during the transition period. 

yes, I agree. that was just a thought I had in mind since it touched the `esm` topic.

mreinstein:
@kzc  this is freakin awesome, thank you!! ❤️

kzc:
Glad it's useful.  It simplified my workflow.

PR #4270, opened 2021-11-14
Use entryFileNames when generating filenames for virtual modules

### LINKED ISSUE #4269: When preserveModules is true, virtual files should use output.entryFileNames when determining their filename
### Rollup Version

2.59.0

macos

Not relevant to rollup output. But a version with support for `.mjs` files is needed to showcase the bug

### Link To Reproduction

https://replit.com/@BPScott/rollup-plugin-babel-repl-helpers-bad-filename

### Expected Behaviour

Originally raised in https://github.com/rollup/plugins/issues/1031, but I think the issue might be better solved at the rollup-level rather than the plugin-level.

When `output.preserveModules:true` is set and there are virtual files, rollup should use `output.entryFileNames` to generate the filename of created virtual files.

When given the following output options, virtual files should be output with a `.mjs` extension, regardless of their original file extension:

```
 output: {
    format: 'esm',
    dir: 'output',
    entryFileNames: '[name][assetExtname].mjs',
    preserveModules: true,
  },
```

I.e. in the given example that uses plugin-babel - `./_virtual/_rollupPluginBabelHelpers.js` should be named `./_virtual/_rollupPluginBabelHelpers.mjs`.

### Actual Behaviour

When `output.preserveModules:true` and `output.entryFilenames` is set and there are virtual files, rollup does not use `entryFileNames` to generate the filename of created virtual files.

In the above demo the `_rollupPluginBabelHelpers` virtual file is output with the filename `./_virtual/_rollupPluginBabelHelpers.js`.

This then breaks running compiled output in node as node assumes that contents of `./_virtual/_rollupPluginBabelHelpers.js` is commonjs, but it is actually esm.

### Bonus thoughts

I believe the fix for this lives within [Chunk.ts's `generateIdPreserveModules` ](https://github.com/rollup/rollup/blob/66b3139b4be1d52342088dc00ef89fb9412f639a/src/Chunk.ts#L428). I've tried moving the definition of `fileName`  above the `isAbsolute()` if block, and changing setting the virtual path block to be "path = `_virtual/${fileName}`;". This seems to give me the filename I'm after, but I get lots of test failures that I can't seem to resolve by updating expected sample output, so clearly I'm missing a trick somewhere.

BPScott:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Resolves #4269
Resolves rollup/plugins#1031
Resolves rollup/plugins#1037

### Description

When `output.preserveModules:true` is set and there are virtual modules, rollup creates files on disk for each of those virtual modules.

Prior to this PR the filename was the sanitised virtual module id but no further transformation was done. This meant that if the virtual module id had no extension, then the filename on disk lacked an extension, and if did specify an extension then the filename on disk would always be that extension, even if `output. entryFileNames` tried to use a different filename.

This causes issues in cases where an environment can not load files that lack an extension (such as web browsers when the server can't set the correct mime-type - https://github.com/rollup/plugins/issues/1037) or when the extension needs to be changed (such as when you want to create `.mjs` files for node esm support - https://github.com/rollup/plugins/issues/1031)

With this PR, the filenames of virtual modules are processed with `output.entryFileNames` in the same way as non-virtual modules - which by default will add a `.js` extension to generated files and shall allow for customization of the extension of the filename.

### Notes

This PR has been split into three commits to try and help ease reviewing

- e67174ea contains the fix itself within Chunk.ts
- 8c5030d4 contains updates to all existing tests to reflect that files for virtual modules now get an extension by default:
  - The virtual module `\0virtualModule` previously generated `_virtual/_virtualModule` and now generates `_virtual/_virtualModule.js`
  -  The virtual module `\0other.js?commonjs-exports.js` previously generated `_virtual/_other.js_commonjs-exports` and now generates `_virtual/other.js_commonjs-exports.js`
  - SystemJS references to modules include the `.js` extension, this seems to be consistent with how files produced from non-virtual modules are referenced
  - AMD references to modules omit the `.js` extension, but include other extensions, this seems to be consistent with how files produced from non-virtual modules are referenced
- 8f3747b3 contains a new test that demonstrates the various entryFilenames keys shall be replaced in virtual filenames too


BPScott:
Rebased atop latest master

lukastaegert (approved):
Thanks, looks great!

BPScott:
Thank you @lukastaegert <3

PR #4286, opened 2021-11-29
fix: escape dynamic import paths

danielroe:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This PR addresses issues when using rollup for a project with `'` in the path, for example `/Users/Daniel's Projects/project/`. Although normal imports are escaped correctly, it seems that dynamic imports are treated differently and we need to escape them as well.

Context: https://github.com/nuxt/framework/issues/2192.

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -921,13 +921,15 @@ export default class Chunk {
 				}
 				const renderedResolution =
 					resolution instanceof Module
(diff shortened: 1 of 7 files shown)


lukastaegert (approved):
Thanks, great catch + fix!

PR #4288, opened 2021-11-30
refactor: modify mkdirpath

### LINKED ISSUE #4287: RangeError: Maximum call stack size exceeded while building

When rollup is run on system with restrictive open files ulimit the error returned from `mkdirpath` function should be `EMFILE: too many open files.`

The error reported is `RangeError: Maximum call stack size exceeded`.

### Modification Proposal

Current implementation of `mkdirpath` treats every error from `fs.readdirSync` the same, calls itself recursively to create parent directory.

Few ideas to improve this function:
 - recursion only while error is ENOENT: no such file or directory
 - remove try..catch and use `fs.existsSync(path)`
 - use `fs.mkdirSync` with "recursive" option

mgrabowski84:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

resolves #4287

### Description

Refactors `mkdirpath` to use `fs.mkdirSync` recursive option (if directory is nested all subsequent directories will be created, same as using `mkdir -p` in Bash).


lukastaegert (approved):
Makes sense, thanks!

merceyz:
The `recursive` option was added in v10.12.0 but rollup supports >=10 so this is a breaking change
https://github.com/rollup/rollup/blob/49d41f78987a1dd13488df009fe2db1b5d1b6e7c/package.json#L131
https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsmkdirsyncpath-options

dnalborczyk:
> The `recursive` option was added in v10.12.0 but rollup supports >=10 so this is a breaking change
> 
> https://github.com/rollup/rollup/blob/49d41f78987a1dd13488df009fe2db1b5d1b6e7c/package.json#L131
> 
> 
> [nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsmkdirsyncpath-options](https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsmkdirsyncpath-options)

I think this should be permissible, given that node 10.x is end-of-life and unmaintained. rollup (and most OSS in general) is only testing the latest LTS (minor/patch) versions.

PR #4298, opened 2021-12-11
refactor: use set for reserved words

dnalborczyk:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no


lukastaegert (approved):
I am sure this will be slightly slower from a performance perspective (because object access is incredibly fast, especially if the object is never mutated and initialized with a literal) but I guess it will not be noticeable.

PR #4299, opened 2021-12-11
feat: add additional string prototype methods

dnalborczyk:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

@lukastaegert  could you look at `match` and `matchAll` if `returnsBoolean` is what is expected?


lukastaegert (changes requested):
No you are right, `.match` (and `.matchAll`) definitely do NOT return booleans. `.match` returns either an array with extra properties or `null`, and `.matchAll` returns an iterator. We should use `returnsUnknown` for both.
Similarly, `.at` can return `undefined` and should therefore also be a `returnsUnknown`. Checking the other `...at` methods, `.codePointAt` should also be changed to unknown (charAt and charCodeAt seem to return consistent types, though).

Could you adjust those?
Otherwise, I am not sure it makes any sense to mark methods as deprecated, after all, nothing gets every removed from JavaScript so we need to keep supporting it. 😉

dnalborczyk:
> No you are right, `.match` (and `.matchAll`) definitely do NOT return booleans. `.match` returns either an array with extra properties or `null`, and `.matchAll` returns an iterator. We should use `returnsUnknown` for both. 

ah, ok. I was about to ask. 👍 

> Similarly, `.at` can return `undefined` and should therefore also be a `returnsUnknown`. Checking the other `...at` methods, `.codePointAt` should also be changed to unknown (charAt and charCodeAt seem to return consistent types, though). Could you adjust those? 

I'll fix those as well.

> Otherwise, I am not sure it makes any sense to mark methods as deprecated, after all, nothing gets every removed from JavaScript so we need to keep supporting it. 😉

the comments are more of a reminder. e.g. `trimRight` and `trimLeft` are only mentioned in the MDN as a side note, as being an alias for `trimStart` and `trimEnd`.

dnalborczyk:
> > Otherwise, I am not sure it makes any sense to mark methods as deprecated, after all, nothing gets every removed from JavaScript so we need to keep supporting it. 😉

> the comments are more of a reminder. e.g. trimRight and trimLeft are only mentioned in the MDN as a side note, as being an alias for trimStart and trimEnd.

in the same vein I'm gonna add those other _deprecated_ "HTML" string methods (blink, link, small, etc.), which are still in V8 (incl. node.js) and possibly other engines.

lukastaegert (approved):

PR #4303, opened 2021-12-15
feat: file extensions opt-in and lookup

dnalborczyk:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

resolves https://github.com/rollup/rollup/issues/4301

Todo:
- [ ] tests
- [ ] docs
- [ ] accept `string` option? (can also be given as single array value instead ['.js']
- [ ] how do extension-less files (currently) work? what would the option be? (empty string?)

Initially I thought to implement a `requireFileExtensions` (boolean) option, but I thought it might be better to re-use the same option for the [extension lookup](https://github.com/rollup/rollup/blob/2e91c85fc95bc722e5a7141d7fa0acde4aab61aa/src/utils/resolveId.ts#L51) , which can also be specified now. similar to [webpack resolve extensions](https://webpack.js.org/configuration/resolve/#resolveextensions)

**false** _(default)_
currently an `opt-in`, and therefore non-breaking. in that case the `internal default extensions` would be used, which are currently `['.mjs', '.js']`.

**true**
all given file paths are expected to have a file extension. they are required.

**string[] e.g. ['.js'], ['.js', '.mjs']**
replaces the `default extensions`, which also include the order of file lookups, for example ['.js', '.mjs] would first look for `.js` files, then for `.mjs` files. it also means that file extensions are not required, but still allowed.

personally I would _strongly suggest_ to make this option an `opt-out` (default: true) for v3, as node.js, the web, and other engines (nginx, moddable etc.) work the same way (including I believe webpack). rollup seems to be the only one standing out.


lukastaegert:
@dnalborczyk what is the state of this one? While I did not do a deep review yet, the proposal looks sound to me.

lukastaegert:
Carefully closing this as stale at this point. Feel free to re-open if you think we should add this.

PR #4321, opened 2021-12-27
fix: replace comparison

dnalborczyk:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

@lukastaegert jus a follow-up from: https://github.com/rollup/rollup/pull/4302#discussion_r775314654.

let me know if you have any questions. I can explain it a bit more (it's a bit late right now 😄 ) I'm trying to achieve that the types can eventually be trusted. 😉  although, someone might have to start likely with `acorn` to achieve that: https://github.com/acornjs/acorn/issues/946

### DIFF
--- a/src/utils/getOriginalLocation.ts
+++ b/src/utils/getOriginalLocation.ts
@@ -5,7 +5,7 @@ export function getOriginalLocation(
 	location: { column: number; line: number; name?: string; source?: string }
 ): { column: number; line: number } {
 	const filteredSourcemapChain = sourcemapChain.filter(
-		(sourcemap): sourcemap is ExistingDecodedSourceMap => sourcemap.mappings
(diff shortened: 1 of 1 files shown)


lukastaegert (changes requested):
I am sorry but I do not see how this improves anything. The types have been working before, the previous check for truthiness was correct as in all cases where `.mappings` was falsy, it was definitely not a valid SourceMap. So why are you so determine to add this? Sure, the TypeScript types say it should be enough to compare against `undefined`, but in the end, source maps are generated by plugins and may have been generated completely without any types. TypeScript cannot give any guarantees here. And at this point, maybe some build setup was relying on this, who knows. The more general check we had was basically for free, so I would prefer to keep it.

PR #4333, opened 2021-12-30
refactor: convert utils.transform to async function

dnalborczyk:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Typescript found another potential bug during the conversion:

```
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2454: Variable 'curPlugin' is used before being assigned.
src/utils/transform.ts (156:25)

156   throwPluginError(err, curPlugin.name, { hook: 'transform', id });
```

which makes sense, since the assignment is not guaranteed in the `try` block, if the code throws anywhere prior.

### DIFF
--- a/src/utils/transform.ts
+++ b/src/utils/transform.ts
@@ -37,7 +37,7 @@ export default async function transform(
 	const emittedFiles: EmittedFile[] = [];
 	let customTransformCache = false;
 	const useCustomTransformCache = () => (customTransformCache = true);
-	let curPlugin: Plugin | undefined;
+	let pluginName = '';
 	const curSource: string = source.code;
 
 	function transformReducer(
@@ -85,7 +85,7 @@ export default async function transform(
 			[curSource, id],
 			transformReducer,
 			(pluginContext, plugin): TransformPluginContext => {
-				curPlugin = plugin;
+				pluginName = plugin.name;
 				return {
 					...pluginContext,
 					addWatchFile(id: string) {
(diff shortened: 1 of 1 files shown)


lukastaegert (changes requested):
> which makes sense, since the assignment is not guaranteed in the try block, if the code throws anywhere prior.

Actually it is guaranteed to be assigned because it is assigned before the first plugin hook is executed. There is nothing that can throw before this point. Which is why it is impossible to write a test to cover your change.
Maybe there is another way to make TypeScript happy than to add unnecessary optionals.

dnalborczyk:
> Actually it is guaranteed to be assigned because it is assigned before the first plugin hook is executed. There is nothing that can throw before this point. Which is why it is impossible to write a test to cover your change. Maybe there is another way to make TypeScript happy than to add unnecessary optionals.

I know what you mean, I imagine it's not easy to track what's going on in the `try block` (what is guaranteed and what not).

```js
let foo
try {
    foo = { name: 'name'}
} catch (err) {
    console.log(foo.name) // Object is possibly 'undefined'.(2532)
}
```

also lifting the callback out of the try/catch block didn't help:

```
let foo
function bar() {
    foo = { name: 'name'}
}
bar()
console.log(foo.name) // Object is possibly 'undefined'.(2532)
```

anyhow, I initialized the plugin name variable.

lukastaegert (approved):

PR #4390, opened 2022-02-07
refactor: add @types/estree explicitly, fix dynamic type imports

dnalborczyk:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

- **@types/estree** was available only implicitly thru other dependencies.
- also replaced the dynamic type imports in declarations. (I believe that's meant for non-module imports, where there's no other way to import types otherwise.)
- simplified `NodeWithFieldDefinition` union type definition
- bumped some deps.

before:
image

after:
image


dnalborczyk:
https://github.com/rollup/rollup/runs/5100138440?check_suite_focus=true seems is a left-over failing test case from fs.promises/chunks.

lukastaegert (approved):

PR #4463, opened 2022-04-13
use void for options hook instead of undefined

ycmjason:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other: typing

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

currently we are required to return `undefined` in the options hook when using TS. this change would allow the options hook to have no return value.


lukastaegert (approved):
Good point, the intention is actually to allow you not to write a return. I also changed this for the other hooks.

PR #4473, opened 2022-04-24
chor: Make husky optional

frank-dspeed:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

when you optain the src via a other method then git clone for example via zip tar gz or other complex git operations the .git folder is not present and so the npm install fails. on prepare as husky can not detect .git 

i want to propose to make husky optional if the directly structure does not match

## Enables
Building the source and work with it.


lukastaegert (changes requested):
While it is a good idea, this does not work for two reasons:
* at least for me, `existsSync` needs an absolute path. This is easily remedied by going for 
  ```
  require("fs").existsSync(require("path").resolve("./git"))
  ```
  though
* If the exit code is 1, the entire command will fail and you basically cannot install at all while the intention is to just not install husky. One way to fix this could be to add
  ```
   || node -e 'console.log("Skipping husky installation")'
  ```
  which will also give some nice feedback so one may at least notice when we accidentally disabled husky for local installs as well.

frank-dspeed:
@lukastaegert as always nice feedback from you i will come up with a new solution that also applys for the case that npm run build fails we will add to both steps

```
|| echo 'Skipping because of prev Error Message'
|| echo 'Skipping because of prev Error Message'
```

the error messages generated by the original commands  will already contain all needed infos. We only swallow the Error return code

frank-dspeed:
Interristing news and observation i did try the node -e console log as also echo both did not get showed on stdout or stderr but the swallowing of the error codes via || did also not work 

this has not always been the case i verifyed it and npm got some upgrades it now does not use the shell that has spawned the npm process. Which is part of security considerations and logical 

how ever the only cross platform real working solution would be a ./npm-prepare.js

dnalborczyk:
@frank-dspeed maybe this will help you: https://typicode.github.io/husky/#/?id=disable-husky-in-cidockerprod

frank-dspeed:
@dnalborczyk your partial correct i did not think about the npm flags but they will work for that case.
thanks a lot. They will enable at last npm install without problems. good job.

frank-dspeed:
i will throw a additional .npmrc into the repo to set the flags without touching the existing code

PR #4657, opened 2022-10-10
feat: support nested plugin

sxzz:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Support nested plugin

```js
{
  plugins: [[p1, p2], p3, p4],
}
```

Since [Vite](https://vitejs.dev/config/shared-options.html#plugins) supported this feature already. I think Rollup should have it too.

Consider the case:
```js
const plugin = {
  buildStart: {
    order: 'pre',
    handler() {
      // actual implementation
    }
  }
}
```

When using object hooks, there's no way to put order both `pre` and `post` in the one plugin ATM.

---

After this PR:

```js
const plugin = [
	{
	  buildStart: {
	    order: 'pre',
	    handler() {
	      // actual implementation
	    }
	  }
	},
	{
	  buildStart: {
	    order: 'post',
	    handler() {
	      // actual implementation
	    }
	  }
	}
]
```


src/utils/options/normalizeInputOptions.ts
```diff
@@ -54,7 +55,7 @@ export function normalizeInputOptions(config: InputOptions): {
 		moduleContext: getModuleContext(config, context),
 		onwarn,
 		perf: config.perf || false,
-		plugins: ensureArray(config.plugins),
+		plugins: flatten(ensureArray(config.plugins)),
```
lukastaegert, line 58: Two things:
 - In the end, `flatten` just reimplements `[].flat(Infinity)`, which I would prefer to use
 - `ensureArray` is only called on the original array. However, it also filters falsy plugins, which should also happen in nested arrays.

```suggestion
		plugins: ensureArray(config.plugins).flat(Infinity).filter(Boolean),
```

test/function/samples/nested-plugin/_config.js
```diff
@@ -0,0 +1,21 @@
...
+			return code.replace('foo = 1', 'foo = 2');
+		}
+	},
+	{
+		name: 'nested-plugin-2',
+		transform(code) {
+			return code.replace('answer = 41', 'answer = 42');
+		}
+	}
+];
```
lukastaegert, line 14: Maybe add a `false` or `undefined` to the array to test the nested filtering. The idea being that plugins can use simple switches to deactivate nested plugins. Rollup should just ignore this.

lukastaegert (changes requested):
Very nice! See my comment. We would also need some documentation for this, though. You should update the types and mention array support in the `plugins` and `output.plugins` sections of the `999-big-list-of-options.md`.

One last thing to consider is that ideally, the `--configPlugin` option should support arrays as well. Which means we should probably replace https://github.com/rollup/rollup/blob/250c2b2f4d0f89ae2e499f8932e4b0661e82226c/cli/run/commandPlugins.ts#L25-L25 with something like

```js
const plugins = ensureArray(commandPlugin).flat(Infinity).filter(Boolean);
```

as well.

src/utils/options/normalizeInputOptions.ts
sxzz: I just noticed `target` in `tsconfig` is `ES2020` now. I implemented this feature before 3.0 was released. Thanks for suggestions!

lukastaegert: Ah right, there was a release yesterday 😄

cli/run/commandPlugins.ts
```diff
@@ -8,11 +8,14 @@ export async function addCommandPluginsToInputOptions(
 	inputOptions: InputOptions,
 	command: Record<string, unknown>
 ): Promise<void> {
+	if (!Array.isArray(inputOptions.plugins)) {
```
lukastaegert, line 11: Actually at this point, it is guaranteed that `plugins` is an array, I just forgot to reflect it in the types. The reason is that `mergeInputOptions` runs first on the options that are passed here, and that one ensures `plugins` is an array. A nicer solution would be the following:

In `types.d.ts`, introduce a new type like

```ts
export interface InputOptionsWithPlugins extends InputOptions {
	plugins: (Plugin | null | false | undefined)[];
}
```

and change

```ts
export interface MergedRollupOptions extends InputOptionsWithPlugins {
	output: OutputOptions[];
}
```

Then use the new type in `mergeInputOptions` and change the type in `addCommandPluginsToInputOptions` from `InputOptions` to the more correct `MergedRollupOptions`.

cli/run/commandPlugins.ts
```diff
@@ -96,9 +99,10 @@ async function loadAndRegisterPlugin(
 			)}" for Rollup to recognize it.`
 		);
 	}
-	inputOptions.plugins!.push(
-		typeof plugin === 'function' ? plugin.call(plugin, pluginArg) : plugin
-	);
+	if (!Array.isArray(inputOptions.plugins)) {
```
lukastaegert, line 102: Again with the type changes, this should be unnecessary. Removing this check will then also fix coverage.

lukastaegert (approved):
Very nice!

sxzz:
Thanks for your guidance. I was very impressed with your presentation at ViteConf!

sxzz:
@lukastaegert Coverage seems to have failed. I just removed `!` type assertion and formatted it.
image

lukastaegert:
> Coverage seems to have failed

No worries. At some point I guess we should add the missing test, but this was indeed unrelated to your PR.

PR #4877, opened 2023-02-24
feat: mark files in a `node_modules` as ignore-listed by default

bmeurer:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

https://github.com/rollup/rollup/issues/4847

### Description

Following up on the addition of the `sourcemapIgnoreList` output option, this adds a default which marks all files within a `node_modules` folder as ignore-listed. This way most projects using rollup will need no additional configuration regarding the ignore-listing and automatically benefit.

Doc: https://goo.gle/devtools-ignoreList-adoption

### DIFF
--- a/docs/configuration-options/index.md
+++ b/docs/configuration-options/index.md
@@ -1415,9 +1415,9 @@ The location of the generated bundle. If this is an absolute path, all the `sour
 
 ### output.sourcemapIgnoreList
 
-|       |                                                                  |
-| ----: | :--------------------------------------------------------------- |
-| Type: | `(relativeSourcePath: string, sourcemapPath: string) => boolean` |
+|  |  |
+| --: | :-- |
+| Type: | `boolean \| (relativeSourcePath: string, sourcemapPath: string) => boolean` |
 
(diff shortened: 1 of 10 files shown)


lukastaegert (changes requested):
Thank you!

While I approve the change in general, now the only way to switch it off is to pass a function that always returns `false`.
I think we should adjust the types and handling slightly:
- E.g. allow an explicit `false` to turn off the feature. That would be consistent with how this works for other options like `treeshake` or `watch`
- Either translate `false` to an "always false" predicate and have always a function in the `normalizedOutputOptions`, or hand through the `false` value. I am slightly leaning to the former approach of always having a function for simplicity. In that case, we can remove the conditional logic in `renderChunks` that is currently not covered.

And of course, documentation should be adapted to show the new types and explain the default logic.

bmeurer:
> Thank you!
> 
> While I approve the change in general, now the only way to switch it off is to pass a function that always returns `false`. I think we should adjust the types and handling slightly:
> 
> * E.g. allow an explicit `false` to turn off the feature. That would be consistent with how this works for other options like `treeshake` or `watch`
> * Either translate `false` to an "always false" predicate and have always a function in the `normalizedOutputOptions`, or hand through the `false` value. I am slightly leaning to the former approach of always having a function for simplicity. In that case, we can remove the conditional logic in `renderChunks` that is currently not covered.
> 
> And of course, documentation should be adapted to show the new types and explain the default logic.

Done, thanks!

lukastaegert (approved):

PR #4932, opened 2023-04-09
Allow shouldTransformCachedModule to return null

bluwy:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Allow `shouldTransformCachedModule` to return null. The hook uses the `hookFirst` approach, which only delegates other plugins if the return value is null.

https://github.com/rollup/rollup/blob/0bcf0a672ac087ff2eb88fbba45ec62389a4f45f/src/utils/PluginDriver.ts#L141-L154

Since the types only restricts returning a `boolean`, it makes it that the first plugin that implements this will take precedence over everything. Allowing to return null should fix this.

Perhaps there could be discussion if this was intentional.

### DIFF
--- a/docs/plugin-development/index.md
+++ b/docs/plugin-development/index.md
@@ -564,14 +564,14 @@ type ShouldTransformCachedModuleHook = (options: {
 	meta: { [plugin: string]: any };
 	moduleSideEffects: boolean | 'no-treeshake';
 	syntheticNamedExports: boolean | string;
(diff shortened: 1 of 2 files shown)


lukastaegert (approved):
Good catch!

PR #4943, opened 2023-04-18
Add a test for reserved keywords used as import/export specifiers

Andarist:
It's just a test as I it crossed my mind that this might need rechecking after https://github.com/rollup/rollup/pull/4939 landed. Everything works OK but having an extra test won't hurt


lukastaegert (approved):
Great to have this 👍

PR #4954, opened 2023-04-23
test: enable typecheck for _config files

antfu:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This would be helpful for contributors to write tests, and see what is available as test options.

Currently, it only enables IDE type-checking as some of the functions are not fully type-covered. We could improve the along the way, but right now, being able to typecheck the Rollup options field `options` is a huge improve to me already.


antfu:
I tried to understand what each configuration stands for in the tsdocs - feel free to directly update them if you see any improvements.

lukastaegert (commented):
This is actually a great idea! One thing that could be improved, though: The different test types do not support the same fields. Many are shared, of course, like `description` or `solo`. But only function tests support `exports` and `global` and some other props, while `formats` I think is specific to chunking-form (would need to check). It would be nice to have this difference also in the type check. One thing that happens quite often is that a test is converted between `function` and `form`. At that point, the additional check would help to highlight if the config needs to be changed.

One idea could be to have different helpers, e.g. `defineFunctionTest`, `defineChunkingTest`, `defineFormTest`,... (and I do not think we need Rollup in the name to keep the names shorter).

While you would get no warning if you use the wrong global, using the wrong global would immediately turn the actual test red because each test type only injects the global for this type (or injects something that throws a more helpful error for the others).

Or maybe you have a better idea?

antfu:
Interesting! I think maybe I could make it work using the same function name in different directly. Let me play with it!

antfu:
Managed to make it work! Now the type of `defineTest` will be based on the folder. I think it would also be helpful when moving tests - the type check will tell which config entries needs to be changed.

Screenshot 2023-04-23 at 23 00 23
Screenshot 2023-04-23 at 23 02 06

lukastaegert (commented):
That looks amazing! Now the only thing is that we have a million of type errors...

antfu:
Maybe we could disable the typecheck for now (but we will still get the autocomplete, etc.) and opt-in by `// @ts-check` per file. That way we could move forward and I will try to improve them from time to time

lukastaegert:
I finalized the types for the available properties per test now and would merge this now. There are still some subtle things I am not happy about, though. Mostly, there are some issues as TypeScript does not like importing types from outside the current tsconfig file. One way to solve this could be to go back to a single tsconfig for tests and have different test helpers per test type, but maybe for another PR.

lukastaegert (approved):

antfu:
Thanks!

PR #5045, opened 2023-06-22
Fix `this.error` with `pos` in `transform` hook

### LINKED ISSUE #5044: "Cannot destructure property 'line' of 'locate(...)' as it is undefined" happens when `pos` is passed to `this.error`
### Rollup Version

3.25.1

stackblitz

_No response_

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-1umn8w?file=rollup.config.js

### Expected Behaviour

The correct error is output.

### Actual Behaviour

```
[!] (plugin vite:import-analysis-build) TypeError: Cannot destructure property 'line' of 'locate(...)' as it is undefined.
```
happens

refs https://github.com/vitejs/vite/issues/13571

sapphi-red:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #5044
resolves https://github.com/vitejs/vite/issues/13571

### Description

The original source was passed to `augmentCodeLocation` instead of the current source.

### DIFF
--- a/test/function/samples/plugin-error-transform-pos/_config.js
+++ b/test/function/samples/plugin-error-transform-pos/_config.js
@@ -25,7 +25,7 @@ module.exports = defineTest({
 		hook: 'transform',
 		id: path.join(__dirname, 'main.js'),
 		watchFiles: [path.join(__dirname, 'main.js')],
-        frame: "1: export const foo = 'foo'bar\n                             ^",
+        frame: "11: export const foo = 'foo'bar\n                             ^",
         loc: {
             column: 26,
             file: path.join(__dirname, 'main.js'),
--- a/test/function/samples/plugin-error-transform-pos/_config.js
+++ b/test/function/samples/plugin-error-transform-pos/_config.js
@@ -1,4 +1,5 @@
 const path = require('node:path');
+const ID_MAIN = path.join(__dirname, 'main.js');
 
 module.exports = defineTest({
 	description: '`this.error(...)` accepts number as second parameter (#5044)',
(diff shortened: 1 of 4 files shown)


test/function/samples/plugin-error-transform-pos/_config.js
```diff
@@ -0,0 +1,36 @@
...
+		]
+	},
+	error: {
+		code: 'PLUGIN_ERROR',
+		message: 'error',
+		plugin: 'plugin2',
+		hook: 'transform',
+		id: path.join(__dirname, 'main.js'),
+		watchFiles: [path.join(__dirname, 'main.js')],
+		frame: "11: export const foo = 'foo'bar\n                             ^",
```
sapphi-red, line 28: I'm not sure why the test fails when `frame` is `1: export const foo = 'foo'bar\n                             ^`.

TrickyPi: The reason is that `deindent` function is slicing the string from index 1 (see https://github.com/rollup/rollup/blob/833314c1621420e031340c1894e7863e062f7761/test/utils.js#L104). You can use "Template literals" like this
```js
frame: `
1: export const foo = 'foo'bar
                                               ^
`,
```

lukastaegert (approved):
Thanks a lot for spotting this! I took the liberty of adjusting the test as suggested by @TrickyPi, looks good to me otherwise.

test/function/samples/plugin-error-transform-pos/_config.js
sapphi-red: Ah, I see. Thank you!

PR #5054, opened 2023-07-05
allow manualChunks to be false

SIGSTACKFAULT:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other - only changed types

Are tests included?

- [ ] yes
- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: N/A

### Description

allow manualChunks to be false so it can be overridden when configs are merged. this probably only affects me.

### DIFF
--- a/src/rollup/types.d.ts
+++ b/src/rollup/types.d.ts
@@ -727,7 +727,7 @@ export interface OutputOptions {
 	inlineDynamicImports?: boolean;
 	interop?: InteropType | GetInterop;
 	intro?: string | AddonFunction;
-	manualChunks?: ManualChunksOption;
+	manualChunks?: ManualChunksOption | false;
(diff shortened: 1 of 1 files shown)


lukastaegert (commented):
I assume an explicit `undefined` would not work? The problem is, there is quite a bit more that needs to happen to add a new config value:

- adapt normalized options, or adapt the normalizer
- adapt documentation
- add a test

Also, `false` feels slightly off for this option. Would `null` also work? What problem is this actually solving? Is Vite generating manual chunks that you do not want? Wouldn't it make more sense to adapt/add an option on Vite side? After all, this fixes no problem inherent to Rollup but to a secondary tool.

Itzyboo1 (approved):

PR #5077, opened 2023-07-27
feat: Add WASM browser build

TrickyPi:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

relavant #5073 

#### Maybe we could accept The JavaScript wrapper 
Currently, I didn't specific the value of `--target` to `web` for `wasm-pack`, beacause we don't want to generate js wrapper, but `wasm-pack` also gernerate some js files that are not for instanitating wasm, and we have to import js files directly instead of wasm files. In the docs of `@rollup/plugin-wasm`, it metioned that "The best setting to use with wasm-bindgen or wasm-pack is --target web. This will generate your WebAssembly files (such as sample_bg.wasm) with a JavaScript wrapper (such as sample.js). The JavaScript wrapper contains all of the instantiation code you'll need.". So maybe we could accept generating the js wrapper.


lukastaegert:
> So maybe we could accept generating the js wrapper.

Ah yes, just do what the documentation says. Another thing to check: I do not think that WASM can create a NodeJS Buffer as return value, definitely not in the browser. So we need to adapt the converter logic slightly to work with an ArrayBuffer. This should not be a problem except for reading strings, where you have to do something with TextDecoder I think.

TrickyPi:
> I do not think that WASM can create a NodeJS Buffer as return value, definitely not in the browser. So we need to adapt the converter logic slightly to work with an ArrayBuffer. This should not be a problem except for reading strings, where you have to do something with TextDecoder I think.

Yes, I'm running into this problem.
截屏2023-07-28 13 23 14


> Ah yes, just do what the documentation says

Ok

rust/parse_ast/src/lib.rs
```diff
@@ -29,6 +31,7 @@ impl Emitter for TestEmitter {
   }
 }
 
+#[wasm_bindgen]
```
TrickyPi, line 34: I'm not sure if it affects the napi build, should we create a new package for the wasm build?

rollup.config.ts
```diff
@@ -135,10 +138,13 @@ export default async function (
 		],
 		plugins: [
 			replaceBrowserModules(),
+			replace(wasmReplacement),
+			replace(getEnvironmentReplacement(true)),
 			alias(moduleAliases),
 			nodeResolve({ browser: true }),
 			json(),
 			commonjs(),
+			wasm(),
```
TrickyPi, line 147: I don't set the `targetEnv` option to `browser` for running browser test successfully. But it is generating compatible code for different environments.
image

src/rollup/rollup.ts
```diff
@@ -51,6 +52,8 @@ export async function rollupInternal(
 	);
 	initialiseTimers(inputOptions);
 
+	/** WASM_INIT_PLACEHOLDER */
```
TrickyPi, line 55: The wasm needs to be instantiated, but it is an async process, so I have to put it in an async function(the browser format of Rollup is umd that does not support top-await), I think `rollupInternal` async function is an appropriate place.

rust/parse_ast/src/lib.rs
lukastaegert: Yes, please move it into a separate Cargo workspace, just how I did it for the Napi bindings, even if it means creating a wrapper function that just forwards results. At the very least, this allows to keep dependencies separate to speed up builds, and it probably would cause problems for the Napi build anyway.

rollup.config.ts
lukastaegert: But doesn't that mean the browser build also contains Node code that we do not need? Of course it makes sense to also have a Node Wasm build, but then I would say should be a proper build with CLI published under a separate name. The browser build is really hard to use as it of course needs plugins for everything.

src/rollup/rollup.ts
lukastaegert: I am not super fond of using the replace plugin as it needs to scan every single file for one replace, and in your setup several times. Replacing imports is usually much cheaper in compile time. Here, you could supply a dummy init function for NodeJS that is replaced with the WASM init function in the browser build.

rust/parse_ast/src/lib.rs
TrickyPi: done

src/rollup/rollup.ts
TrickyPi: done

rollup.config.ts
TrickyPi: I set the `targetEnv` option to `browser`. I'm now thinking that how to fix browser tests, maybe we need a headless browser such as playwright to run browser tests?

lukastaegert: Yes, I think so too.

lukastaegert: BTW CI tests should be green now (everything that is not working is commented out), there is a TypeScript issue in your code and maybe you need to add some comments to silence the linter.

browser/src/readString.ts
```diff
@@ -0,0 +1,5 @@
+const textDecoder = new TextDecoder();
+
+export default function readString(astBuffer: Uint8Array, start: number, length: number) {
+	return textDecoder.decode(astBuffer.slice(start, start + length));
```
lukastaegert, line 4: Instead of using `.slice()`, which copies part of the ArrayBuffer (see https://devdocs.io/javascript/global_objects/typedarray/slice), you could use `.subarray()`, which should avoid that, see https://devdocs.io/javascript/global_objects/typedarray/subarray.

TrickyPi: done

rollup.config.ts
TrickyPi: I tried using `puppeteer` to run browser tests, but it's not easy to migrate current browser tests due to runtime context. So I polyfilled some browser global variables like `fetch` and `WebAssembly.compileStreaming`, maybe this is also an appropriate solution?

lukastaegert: Sure, I think this is fine for now.

browser/src/initWasm.ts
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+// eslint-disable-next-line import/no-unresolved
+import init from '../wasm/bindings_wasm';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+// eslint-disable-next-line import/no-unresolved
+import wasm from '../wasm/bindings_wasm_bg.wasm';
```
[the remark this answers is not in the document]
lukastaegert: Good work!

lukastaegert (approved):

TrickyPi:
@lukastaegert I have just received your invitation to join Rollup, I'm sooooooo happy!!!

lukastaegert:
@TrickyPi I also sent you a message on the Rollup Discord. If there is something you want to discuss in a less public space, you can use Discord, otherwise PRs are fine as well. You can try to (squash) merge your PR if you want. If you are not allowed to do that, leave me a message.

PR #5114, opened 2023-08-30
fix(perf): avoid superfluous timer wrappings in watch mode

ZhengLiu2825:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

In watch mode, every time bundle starts, inputOptions.plugins will be wrapped in another layer of timer, which will cause incorrect performance timings.


lukastaegert (commented):
Thank you for spotting is. The only issue I see with your implementation is that `isWithTimers` now is global state, which means that completely unrelated runs will now also influence each other. Basically if you have a config file with an array of configs, only the plugins in the first config will be instrumented.
Maybe a cleaner approach would be to leave a flag on each instrumented plugin in `getPluginWithTimers`, e.g.

```ts
function getPluginWithTimers(plugin: any, index: number): Plugin {
	if (plugin._hasTimedHooks) return plugin;
	plugin._hasTimedHooks = true;
	for (const hook of TIMED_PLUGIN_HOOKS) {
		if (hook in plugin) {

```

One could also use a Symbol to hide the flag, or keep a `Set<Plugin>` of instrumented plugins as global state. What do you think?

lukastaegert (approved):
I changed your code as I suggested and will merge it now.

PR #5144, opened 2023-09-18
[v4.0] Imporve the performance of generating ast and rollup ast nodes

TrickyPi:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,7 +10,8 @@
 		"resolveJsonModule": true,
 		"skipLibCheck": true,
 		"strict": true,
-		"target": "ES2022"
+		"target": "ES2022",
+		"useDefineForClassFields": false
 	},
 	"include": ["typings/**/*.d.ts", "src", "cli", "browser", "rollup.config.ts", "native", "wasm.d.ts"],
 	"exclude": ["dist", "node_modules", "test/typescript", "browser/dist"]
--- a/src/utils/convert-ast.ts
+++ b/src/utils/convert-ast.ts
@@ -55,18 +55,21 @@ const nodeConverters: ((position: number, buffer: Uint32Array, readString: ReadS
 		const parameters = convertNodeList(buffer[position++], buffer, readString);
 		const body = convertNode(buffer[position++], buffer, readString);
 		const annotations = convertAnnotationList(position, buffer);
-		return {
-			type: 'ArrowFunctionExpression',
-			start,
-			end,
-			async,
-			body,
-			expression,
-			generator,
-			id: null,
-			params: parameters,
-			_rollupAnnotations: annotations
-		};
+		return addAnnotationProperty(
+			{
+				type: 'ArrowFunctionExpression',
+				start,
+				end,
+				async,
+				body,
+				expression,
+				generator,
+				id: null,
+				params: parameters
+			},
+			annotations,
+			ANNOTATION_KEY
+		);
 	},
 	// index:3; AssignmentExpression
 	(position, buffer, readString): estree.AssignmentExpression & AstNode => {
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -10,7 +10,8 @@
 		"resolveJsonModule": true,
 		"skipLibCheck": true,
 		"strict": true,
-		"target": "ES2022"
+		"target": "ES2022",
+		"useDefineForClassFields": false
```
dnalborczyk, line 14: I believe this is always true when target is `ES2022`. https://www.typescriptlang.org/tsconfig#useDefineForClassFields

that said, I believe it might not have much of a meaning, as fields are not transpiled with that target.

lukastaegert: Oh, you are in for a surprise: Just try it in the TS Playground. If you set the target to ES2022, then class fields are only rendered as class fields if you (explicitly) set `useDefineForClassFields` to `true`.
https://www.typescriptlang.org/play?useDefineForClassFields=true&target=9#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A

If you set it to `false`, they are instead assigned in the constructor.
https://www.typescriptlang.org/play?target=9&useDefineForClassFields=false#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A

I did not like the effect of `true` before ES2022 because it created huge overhead in the generated code, but I hoped that native class fields would be faster. But apparently they are not, at least not yet, and the performance gains are striking.

lukastaegert:
For completeness, here are the performance measurements that @TrickyPi sent me.

Before this change:
perf

After this change:
perf2

Basically, the analyze AST step that instantiates all internal AST node classes went from 90ms to 140ms only because we were now using proper class fields. Setting `useDefineForClassFields` to `false` changed this number back to 90ms.

tsconfig.json
TrickyPi: Sorry for not including the detailed effect of this PR. Yes, the default of `useDefineForClassFields` is `true` if the target is `ES2022`, Howerver we need to explicitly set it to `false` to place all class fields within the constructor during TS compilation. in that case, we could reduce Rollup runtime.

The performance before
before

The performance after changes
now

dnalborczyk: > Oh, you are in for a surprise: Just try it in the TS Playground. If you set the target to ES2022, then class fields are only rendered as class fields if you (explicitly in the Playground) set `useDefineForClassFields` to `true`. [typescriptlang.org/play?useDefineForClassFields=true&target=9#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A](https://www.typescriptlang.org/play?useDefineForClassFields=true&target=9#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A)
> 
> If you set it to `false`, they are instead assigned in the constructor. [typescriptlang.org/play?target=9&useDefineForClassFields=false#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A](https://www.typescriptlang.org/play?target=9&useDefineForClassFields=false#code/MYGwhgzhAEBiD29oG8BQ0PVJCsCWApiACbQC80ALgE4CuBA3KgL5A)

I wonder if this is a bug in typescript, or intentional behavior? 🤔   I would expect typescript to not transpile, regardless of the `useDefineForClassFields` setting when he target is `ES2022/ESNext`.

> I did not like the effect of `true` before ES2022 because it created huge overhead in the generated code, but I hoped that native class fields would be faster. But apparently they are not, at least not yet, and the performance gains are striking.

this is surprising: https://v8.dev/blog/faster-class-features

lukastaegert (approved):

PR #5198, opened 2023-10-13
fix: prevent `ENOENT` error on temp config removal

jzempel:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

...existing `load-config-file` tests continue to pass

Breaking Changes?

- [x] no

### Description

We're observing `rollup` [errors](https://app.circleci.com/pipelines/github/zendeskgarden/react-containers/3886/workflows/5efdc3a2-c161-4df5-b8b2-c296817e2524/jobs/7556/parallel-runs/0/steps/0-105) (link may archive after a few months)  in a monorepo that runs parallel builds. Example:

`Error: ENOENT: no such file or directory, unlink '/home/circleci/project/utils/build/rollup.config-1697053868023.cjs'`

These errors are becoming more and more common as the performance of the build tooling improves. This PR substitutes the [promise](https://nodejs.org/api/fs.html#promise-example) signature of Node unlink with the async [callback](https://nodejs.org/api/fs.html#callback-example) signature. Doing so allows us to safely "eat" the non-critical error of attempting to delete a non-existent temporary config file.


lukastaegert (changes requested):
While I do not know why you have these issues (is some parallel process removing files?) I have no issues swallowing errors here. But I do not understand why you need the non-promise unlink function for this, and I do not like it as it is inconsistent.
You can also "safely eat errors" by either awaiting and writing it in a try-catch with empty handler, or append a `.catch(doNothing)`.

Though I would actually feel better if we would still show a warning in that case as that points to an issue with your build system.

jzempel:
@lukastaegert do you think we have something acceptable for merge here? Thank you 🙏

lukastaegert (approved):

PR #5203, opened 2023-10-14
fix: export xxhashBase64Url from wasm

sapphi-red:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Rollup wasn't working in stackblitz because `xxhashBase64Url` was not re-exported from wasm.
Reproduction: https://stackblitz.com/edit/rollup-repro-ykzjaj?file=package.json

This PR fixes that and add a lint script to detect these kinds of mistakes.

P.S. I noticed the stackblitz template for the bug report uses rollup v3.


native.wasm.js
```diff
@@ -0,0 +1,4 @@
+const { parse, xxhashBase64Url } = require('./wasm-node/bindings_wasm.js');
+
+exports.parse = parse;
+exports.xxhashBase64Url = xxhashBase64Url;
```
TrickyPi, line 4:
```suggestion
const { parse, xxhash_base64_url } = require('./wasm-node/bindings_wasm.js');

exports.parse = parse;
exports.xxhashBase64Url = xxhash_base64_url;
```
Currently, `wasm-pack` doesn't convert `under_score_case` to `camelCase`.

sapphi-red: Good catch! I felt renaming on the wasm side will make the code more consistent. So I did it in that way.

TrickyPi: Nice work! I wasn't aware of this attribute `js_name` before.

lukastaegert (approved):
Thanks so much!

PR #5336, opened 2024-01-10
Normalize resolveId source

btea:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description
On the Windows platform, the path containing double backslashes may cause an error when reading the file.


lukastaegert:
Unfortunately, this looks to me like a breaking change. Could you please give a test case which problem you are trying to solve?

btea:
This is a issue submitted by someone in the vite repository. I tried to fix this problem on the vite side, and it just normalized the id received by the plugin transform hook. So I tried normalizing the ids on the rollup side.

Below are the relevant warehouse links and error screenshots, I hope they can be helpful.

https://github.com/movahhedi/vite-bug-inline-css

image

PR #5422, opened 2024-03-09
feat: show all cause in Error

devohda:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #5166 

### Description

In ES2022+, [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) has `cause` field optionally. (type is unknown)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

So if `cause` is in Error then show all cause log in cli.


lukastaegert (commented):
Hi, that looks already very good! Just two things:

- This looks very similar to how Node would print such an error. Node, however, would indent every subsequent cause by two more spaces (i.e. two for the first cause, four for the second etc.). Personally I find this helps visually so that it does not look like one huge stack trace. Maybe we want to add this as well? (one easy way to indent is to `.split('\n').map(line => indent+line).join('\n')`)
- It would be nice to test this output somewhat. Of course we should not test stack traces directly as they contain paths on a user's machine, but we can test stuff in between. Here is how you would do it:

If you do not know how to run tests, look at `CONTRIBUTING.md`.
Tests for command line output are located in `test/cli`. Create a new folder in `test/cli/samples` with the following files:

- `main.js`: It does not matter a lot, but we need an entry point. E.g. `console.log('ok')` would be some content that does not produce a warning.
- `rollup.config.mjs`: We define a config that always throws an error with nested errors, e.g.
  ```js
  export default {
    input: "main.js",
    plugins: [
      {
        buildStart() {
          throw new Error("Outer error", {
            cause: new Error("Inner error", {
              cause: new Error("Innermost error")
            })
           });
        }
      }
    ]
  };
  ```
- `_config.js`: The test configuration. This would be the typical contents
  ```js
  const { assertIncludes } = require('../../../utils.js');
  
  module.exports = defineTest({
  	description: 'prints error cause',
  	command: 'rollup --config rollup.config.mjs',
  	// We expect an error and want to make assertions about the output
  	error: () => true,
  	stderr: stderr => {
  		// We just assert the parts of the output that do not change
  		assertIncludes(stderr, '\n[!] (plugin at position 1) Error: Outer error\n    at ');
  		assertIncludes(stderr, '\n  [cause] Error: Inner error\n      at ');
  		assertIncludes(stderr, '\n    [cause] Error: Innermost error\n        at ');
  	}
  });
  ```

lukastaegert (approved):

PR #5445, opened 2024-03-27
docs: add `@shikiji/vitepress-twoslash`

sapphi-red:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

This PR adds `@shikiji/vitepress-twoslash` so that users can hover on the code and see the types.
Example with the vite docs: https://elk.zone/m.webtoo.ls/@sapphi_red/112131987169625108


docs/configuration-options/index.md
```diff
@@ -1702,14 +1775,19 @@ Note `id` can only be used for single-file builds, and cannot be combined with `
...
-  amd: {
-    id: 'my-bundle'
-  }
+	// ...
+	format: 'amd',
+	output: {
+		amd: {
+			id: 'my-bundle'
+		}
+	}
```
sapphi-red, line 1790: These example codes were wrong (`amd` was not wrapped with `output`).

lukastaegert (commented):
This is truly amazing. Makes me wonder if we should start using adding doc comments in the docs, but I would only want to do that if I can then move the entire documentation into the types to avoid the duplicity. Not sure if this is worth it. Just some minor comments for the AMD section, but otherwise I do not see why we cannot merge this, this must have been a ton of work!

lukastaegert (approved):

PR #5451, opened 2024-04-04
chore: generate string constants from config

TrickyPi:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
Previously, to retrieve a string constant from the Rust parser, we had to modify two files. Now, only one place needs to be modified, and changing the order of string constants is also convenient.


lukastaegert (approved):
Nice, yes, this was really something we needed!

PR #5488, opened 2024-04-23
Report performance in CI

TrickyPi:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
Through this report, we can roughly determine whether there are performance regression or improvement in the PR.


lukastaegert (commented):
This is amazing! One question, though: Currently, this is using the Rollup build, which has many plugins that are out of our control etc. and not too close to the core Rollup performance. I recently added a benchmarking script that you get when you run `npm run perf` (replacing an older benchmark that did not work too well...).
What this does is basically run a benchmark used by esbuild: Download the three.js sources and bundle them ten times. This does not require any plugins and is thus a much better measure of bundling performance. It will also report detailed statistics in which parts of the application we gain or lose performance as well as where the memory consumption improved or became worse. Not sure if it is easy to put this into a report, but this is something I thought we should definitely have eventually.

TrickyPi:
> I recently added a benchmarking script that you get when you run `npm run perf`
👍, Oh, I see it.

>  Not sure if it is easy to put this into a report, but this is something I thought we should definitely have eventually.
Yeah, I agree with you! I think it's not hard to include this information in the report. I'll have a look at it!

lukastaegert (approved):
Nice 👍

One thing to consider if we want to save a runner could be to not build the artifacts in the job but move the report phase to the other Github workflow where we already build all artifacts and run it similar to the tests and smoke tests. But on the other hand, it does not make much of a difference.

PR #5562, opened 2024-07-02
feat: implementing decorator support

### LINKED ISSUE #5553: JS Decorator Stage 3 support
### Rollup Version

latest

all

_No response_

### Link To Reproduction

https://rollupjs.org/repl/?version=4.18.0&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMiUyRiUyRiUyMFRSRUUtU0hBS0lORyU1Q25pbXBvcnQlMjAlN0IlMjBjdWJlJTIwJTdEJTIwZnJvbSUyMCcuJTJGbWF0aHMuanMnJTNCJTVDbiU1Q25jbGFzcyUyMEElMjAlN0IlNUNuJTIwJTIwJTQwZm5jJTVDbiUyMCUyMGZuKCklMjAlN0IlNUNuJTIwJTIwJTIwJTIwJTVDbiUyMCUyMCU3RCU1Q24lN0QlNUNuJTVDbmNvbnNvbGUubG9nKGN1YmUoNSkpJTNCJTIwJTJGJTJGJTIwMTI1JTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCUyQyU3QiUyMmNvZGUlMjIlM0ElMjIlMkYlMkYlMjBtYXRocy5qcyU1Q24lNUNuJTJGJTJGJTIwVGhpcyUyMGZ1bmN0aW9uJTIwaXNuJ3QlMjB1c2VkJTIwYW55d2hlcmUlMkMlMjBzbyU1Q24lMkYlMkYlMjBSb2xsdXAlMjBleGNsdWRlcyUyMGl0JTIwZnJvbSUyMHRoZSUyMGJ1bmRsZS4uLiU1Q25leHBvcnQlMjBjb25zdCUyMHNxdWFyZSUyMCUzRCUyMHglMjAlM0QlM0UlMjB4JTIwKiUyMHglM0IlNUNuJTVDbiUyRiUyRiUyMFRoaXMlMjBmdW5jdGlvbiUyMGdldHMlMjBpbmNsdWRlZCU1Q24lMkYlMkYlMjByZXdyaXRlJTIwdGhpcyUyMGFzJTIwJTYwc3F1YXJlKHgpJTIwKiUyMHglNjAlNUNuJTJGJTJGJTIwYW5kJTIwc2VlJTIwd2hhdCUyMGhhcHBlbnMhJTVDbmV4cG9ydCUyMGNvbnN0JTIwY3ViZSUyMCUzRCUyMHglMjAlM0QlM0UlMjB4JTIwKiUyMHglMjAqJTIweCUzQiU1Q24lNUNuJTJGJTJGJTIwVGhpcyUyMCU1QyUyMnNpZGUlMjBlZmZlY3QlNUMlMjIlMjBjcmVhdGVzJTIwYSUyMGdsb2JhbCU1Q24lMkYlMkYlMjB2YXJpYWJsZSUyMGFuZCUyMHdpbGwlMjBub3QlMjBiZSUyMHJlbW92ZWQuJTVDbndpbmRvdy5lZmZlY3QxJTIwJTNEJTIwJ2NyZWF0ZWQnJTNCJTVDbiU1Q25jb25zdCUyMGluY2x1ZGVFZmZlY3QlMjAlM0QlMjBmYWxzZSUzQiU1Q25pZiUyMChpbmNsdWRlRWZmZWN0KSUyMCU3QiU1Q24lNUN0JTJGJTJGJTIwT24lMjB0aGUlMjBvdGhlciUyMGhhbmQlMkMlMjB0aGlzJTIwaXMlMjBuZXZlciU1Q24lNUN0JTJGJTJGJTIwZXhlY3V0ZWQlMjBhbmQlMjB0aHVzJTIwcmVtb3ZlZC4lNUNuJTVDdHdpbmRvdy5lZmZlY3QxJTIwJTNEJTIwJ25vdCUyMGNyZWF0ZWQnJTNCJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSUyQyUyMm5hbWUlMjIlM0ElMjJtYXRocy5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=

### Expected Behaviour

Support JavaScript Decorator Stage 3, since esbuild, oxlint, swc all support that

https://github.com/tc39/proposal-decorators

### Actual Behaviour

```sh
RollupError: main.js (5:2): Unexpected token `@`. Expected identifier, string literal, numeric literal or [ for the computed key
/main.js (5:2)
3: 
4: class A {
5:   @fnc
     ^
6:   fn() {
```

TrickyPi:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #5553 

### Description
This PR does not account for decorators with auto-accessor properties because this protocol has only reached Stage 1 at the moment.


TrickyPi:
Hi Lukas. Do you have any idea how to fix the failed tests? I added some log info for debugging and found that the AST of `rollup.config.ts` is incorrect. See https://github.com/rollup/rollup/actions/runs/9777093939/job/26991182657
image

phk422 (approved):

TrickyPi:
I think I found the problem: the architecture `x86_64-apple-darwin` is actually arm64🤯.

TrickyPi:
I also found that other people have run into the same problem https://github.com/actions/runner-images/issues/9741. To fix this problem, I downgraded the macOS version to macOS 13.

lukastaegert:
> I think I found the problem: the architecture x86_64-apple-darwin is actually arm64

Again? We had this problem once for the regular build and it was fixed in 8b30a5bbb8f654e9d497c26f1b851a0b93794b96. Let me have a look.

TrickyPi:
> > I think I found the problem: the architecture x86_64-apple-darwin is actually arm64
> 
> Again? We had this problem once for the regular build and it was fixed in [8b30a5b](https://github.com/rollup/rollup/commit/8b30a5bbb8f654e9d497c26f1b851a0b93794b96). Let me have a look.

I guess this PR only fixed the build artifacts step, but in the test step, when the OS is macos-latest and the target is x86_64-apple-darwin, the arch value obtained is arm64
https://github.com/rollup/rollup/blob/21f9a4949358b60801c948cd4777d7a39d9e6de0/native.js#L3

lukastaegert:
Yes, the test step should have just been switched to use the ARM artifact as well. I took the liberty of pushing this.

TrickyPi:
> Most Macs today probably use an ARM architecture, and the x86 Macs are becoming less and less relevant, so we should definitely test ARM instead

yes, indeed.

lukastaegert:
Weird, for some reason it cannot find the downloaded artifact.

TrickyPi:
I guess it is cache problem

TrickyPi:
It works now 🎉.

lukastaegert:
But it works by downgrading the runner. Shouldn't we be able to make it work using the ARM build?

TrickyPi:
> But it works by downgrading the runner. Shouldn't we be able to make it work using the ARM build?

Hmm, we can use the ARM build, but we need to use the correct cache key for node_modules; otherwise, other steps might use the wrong cached node_modules.

lukastaegert:
> Promise.resolve().then(() => console.log(6));
console.log(await ({then(resolve) { resolve(5)}}))

If we do not use different Mac versions, I would not expect there to be problems? After all, then there is a one-to-one mapping between os and arch.

TrickyPi:
Oh, I get it! you are right! Thank you for your patient explanation! ❤️

TrickyPi:
image

After manually clearing the previous cache, it works.

lukastaegert (commented):
This is a very good start! At the moment, this does not implement `accessor` support as outlined here https://github.com/tc39/proposal-decorators, but I think that is fine and can be added later. After all, not a single browser currently supports decorators, so this stuff is rather hard to test.

Talking about testing, I supposed the TypeScript classes still need to be extended. For once, side effect detection currently does not work. As I understand it, decorators are called once when the class declaration is executed. That would mean that all classes here would need to be retained: http://localhost:5173/repl/?local=true&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmZ1bmN0aW9uJTIwZm9vRGVjb3JhdG9yJTIwKCklMjAlN0IlNUNuJTIwJTIwY29uc29sZS5sb2coJ2VmZmVjdCcpJTNCJTVDbiU3RCU1Q24lNUNuJTQwZm9vRGVjb3JhdG9yJTVDbmNsYXNzJTIwRmlyc3QlMjAlN0IlN0QlNUNuJTVDbmNsYXNzJTIwU2Vjb25kJTIwJTdCJTVDbiUyMCUyMCU0MGZvb0RlY29yYXRvciU1Q24lMjAlMjBteU1ldGhvZCgpJTIwJTdCJTdEJTVDbiU3RCU1Q24lNUNuY2xhc3MlMjBUaGlyZCUyMCU3QiU1Q24lMjAlMjAlNDBmb29EZWNvcmF0b3IlNUNuJTIwJTIwbXlQcm9wJTIwJTNEJTIwMSU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSUyQyUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTdEJTJDJTIydHJlZXNoYWtlJTIyJTNBdHJ1ZSU3RCU3RA==

Next, inclusion does not work either, here the decorator is not included: http://localhost:5173/repl/?local=true&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmZ1bmN0aW9uJTIwZm9vRGVjb3JhdG9yJTIwKCklMjAlN0IlNUNuJTIwJTIwY29uc29sZS5sb2coJ2VmZmVjdCcpJTNCJTVDbiU3RCU1Q24lNUNuJTQwZm9vRGVjb3JhdG9yJTVDbmNsYXNzJTIwRm9vJTIwJTdCJTdEJTVDbiU1Q25jb25zb2xlLmxvZyhGb28pJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=

Next would be to make sure identifiers are probably renamed when deconflicted, but this cannot be tested without the previous being fixed http://localhost:5173/repl/?local=true&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCcuJTJGZmlyc3QuanMnJTNCJTVDbmltcG9ydCUyMCcuJTJGc2Vjb25kLmpzJyUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyZnVuY3Rpb24lMjBmb29EZWNvcmF0b3IlMjAoKSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZygnZWZmZWN0JyklM0IlNUNuJTdEJTVDbiU1Q24lNDBmb29EZWNvcmF0b3IlNUNuY2xhc3MlMjBGb28lMjAlN0IlN0QlNUNuJTVDbmNvbnNvbGUubG9nKEZvbyklM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlMkMlMjJuYW1lJTIyJTNBJTIyZmlyc3QuanMlMjIlN0QlMkMlN0IlMjJjb2RlJTIyJTNBJTIyZnVuY3Rpb24lMjBmb29EZWNvcmF0b3IlMjAoKSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZygnZWZmZWN0JyklM0IlNUNuJTdEJTVDbiU1Q24lNDBmb29EZWNvcmF0b3IlNUNuY2xhc3MlMjBGb28lMjAlN0IlN0QlNUNuJTVDbmNvbnNvbGUubG9nKEZvbyklM0IlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlMkMlMjJuYW1lJTIyJTNBJTIyc2Vjb25kLmpzJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMm91dHB1dCUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTdEJTJDJTIydHJlZXNoYWtlJTIyJTNBdHJ1ZSU3RCU3RA==

TrickyPi:
Thank you for your professional review 😊. I will look into these issues.

TrickyPi:
The mentioned test has been added and passed. Ready for review again.

src/ast/nodes/Decorator.ts
```diff
@@ -0,0 +1,35 @@
...
+		if (this.expression instanceof CallExpression) {
+			if (this.expression.hasEffects(context)) {
+				return true;
+			}
+			const returnExpression = this.expression.callee.getReturnExpressionWhenCalledAtPath(
+				EMPTY_PATH,
+				NODE_INTERACTION_UNKNOWN_CALL,
+				new PathTracker(),
+				this.expression
+			);
```
lukastaegert, line 21: I was first wondering about the `origin` parameter, which should usually an instance that caches the return value and not `this.expression`.

But then I think you should not need to mess with this at all or write a special logic for call expressions. Because `CallExpression`, or rather `CallExpressionBase`, already implements that logic. So I would assume just always doing

```js
return this.expression.hasEffectsOnInteractionAtPath(
				EMPTY_PATH,
				NODE_INTERACTION_UNKNOWN_CALL,
				context
			);
```
should work?

src/ast/nodes/ClassDeclaration.ts
```diff
@@ -40,7 +40,7 @@ export default class ClassDeclaration extends ClassNode {
 				code.appendLeft(this.end, `${_}${getSystemExportStatement([variable], options)};`);
 			}
 			const renderedVariable = variable.getName(getPropertyAccess);
-			if (renderedVariable !== name) {
+			if (renderedVariable !== name && this.decorators.length === 0) {
```
lukastaegert, line 43: While I cannot prove it, I think the code that is generated without this change

```js
let Foo$1 = @fooDecorator
class Foo {};
```

might look weird but is actually correct. And it would preserve the class name. So maybe we revert this change?

lukastaegert (commented):
Very nice! There are two small remaining comments, but overall I think we are nearly done.

src/ast/nodes/ClassDeclaration.ts
TrickyPi: Yeah 😄, I tried to run the compiled code generated by TS for the above code, and the result is correct.
https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAZwKYEMA2qAmAKASkQG8AoRRCBZObAOkzgHM8BGAZlYCZOCBucogBOqKCCFI8VMMihCQ0OEIBciAGLho8MEQC8APhKCK0mvUYsuVq-2OIA8gCMAVqmh00WKdTkKoS2woKJ1d3T0xvGV9FIToAByE4fygATzjUQMQAX1Ic7Ch1ODgAElZEXUQAAXDcQlIITHRkZEK4Emy+IA

src/ast/nodes/Decorator.ts
TrickyPi: I tried to do it, and it works 👍.

src/ast/nodes/ClassDeclaration.ts
TrickyPi: But the right part of `let Foo$1 = @fooDecorator` should be the deconflicted name.

TrickyPi: Oh, The previous code didn't take this point into account either. I will fix it soon.

src/ast/nodes/Decorator.ts
lukastaegert: Ah you are right, of course we should not affect the side effect of the calling function. But then again, this can be generalized to also flag side effects in cases like this: https://rollup-kp14t621q-rollup-js.vercel.app/repl/?pr=5562&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmZ1bmN0aW9uJTIwZWZmZWN0KCklMjAlN0IlNUNuJTIwJTIwY29uc29sZS5sb2coJ2VmZmVjdCcpJTNCJTVDbiUyMCUyMHJldHVybiUyMCgpJTIwJTNEJTNFJTIwJTdCJTdEJTNCJTVDbiU3RCU1Q24lNUNuJTQwKGZhbHNlJTIwJTdDJTdDJTIwZWZmZWN0KCkpJTVDbmNsYXNzJTIwRm9vJTIwJTdCJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=

I took the liberty of adding the change and extending the tests because otherwise this looks perfect now.

lukastaegert (approved):

TrickyPi:
nice fix [d8ab5c3](https://github.com/rollup/rollup/pull/5562/commits/d8ab5c30c0ea66a119567f69aab72be3f6da08bf) ! I didn't account for that.

znck:
Thanks a lot @TrickyPi. 

PS. I don't see GitHub sponsors enabled for your account. If you accept sponsors, I would like to show appreciation for your contribution.

TrickyPi:
> Thanks a lot @TrickyPi.
> 
> PS. I don't see GitHub sponsors enabled for your account. If you accept sponsors, I would like to show appreciation for your contribution.

I'm really glad to hear this, but my GitHub Sponsor application is waiting to be reviewed by GitHub, which might take some time.

PR #5621, opened 2024-08-22
fix: reset variable's renderBaseName at the end of renderModules

### LINKED ISSUE #5095: UMD code not be transformed correctly when build cjs with preserveModules first and then build umd
### Rollup Version

3.28.0

windows 10

18.17.0

### Link To Reproduction

https://replit.com/@hq229075284/rollup-umd-test#dist/umd/roll.dc.js

### Expected Behaviour

```javascript
function getClassPrefix(name) {
    return `${baseClassPrefix}-${name}`;
}
```

### Actual Behaviour

```javascript
function index.getClassPrefix(name) {
    return `${baseClassPrefix}-${name}`;
}
```

sunnylost:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

-   resolves #5095

### Description

**Disclaimer**: I do not consider this a complete fix. This submission is merely to share my thoughts and approach to solving the issue.

Based on the description in [Rollup issue #5095](https://github.com/rollup/rollup/issues/5095), the `preserveModules` setting in the first output affects the `renderBaseName` property of the `Variable` object. As a result, this property is not reset during the execution of the second output, leading to incorrect variable names being output.

My solution involves collecting the `Variable` objects that have had their `renderBaseName` modified during the execution of `deconflictImportsOther`. Then, at the end of the ~`renderModules()`~`render()` execution in the `Chunk`, I reset the `renderBaseName` properties of these `Variable` objects.

But I feel that this solution is too makeshift and too specific to this issue. I wonder if there are other properties that might need to be reset after chunk rendering. This problem is quite challenging for me.

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -708,6 +708,11 @@ export default class Chunk {
 		}
 		if (footer) magicString.append(footer);
 
+		for (const variable of this.renderBaseNameResetSet) {
+			variable.renderBaseName = null;
+		}
+		this.renderBaseNameResetSet.clear();
+
 		return {
 			chunk: this,
 			magicString,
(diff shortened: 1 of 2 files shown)


Asask-sized:
#5166

PR #5713, opened 2024-11-02
fix: Deoptimize the proxied object if its property is reassigned in the handler functions

### LINKED ISSUE #5709: Rollup treeshakes incorrectly when proxy is used
### Rollup Version

4.24.3

REPL

_No response_

### Link To Reproduction

https://rollupjs.org/repl/?version=4.24.3&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmNvbnN0JTIwcmVwb3J0X3ZhbHVlcyUyMCUzRCUyMCU3QiUyMHJlYXNvbiUzQSUyMCcnJTIwJTdEJTVDbmNvbnN0JTIwc3dpdGNoU3VibWl0JTIwJTNEJTIwKHN0YXRlKSUyMCUzRCUzRSUyMCU3QiU1Q24lMjAlMjBpZiUyMChzdGF0ZSUyMCUzRCUzRCUyMDEpJTIwJTdCJTVDbiUyMCUyMCUyMGNvbnNvbGUubG9nKCclMjMzMzMzMzMnKSU1Q24lMjAlMjAlN0QlMjBlbHNlJTIwaWYlMjAoc3RhdGUlMjAlM0QlM0QlMjAwKSUyMCU3QiU1Q24lMjAlMjAlMjAlMjBjb25zb2xlLmxvZygnJTIzQjhCOEI4JyklNUNuJTIwJTIwJTdEJTVDbiU3RCU1Q24lNUNuY29uc3QlMjB2YWx1ZXNQcm94eSUyMCUzRCUyMG5ldyUyMFByb3h5KHJlcG9ydF92YWx1ZXMlMkMlMjAlN0IlNUNuJTIwJTIwc2V0KHRhcmdldCUyQyUyMHByb3BlcnR5JTJDJTIwdmFsdWUlMkMlMjByZWNlaXZlciklMjAlN0IlNUNuJTIwJTIwJTIwJTIwdGFyZ2V0JTVCcHJvcGVydHklNUQlMjAlM0QlMjB2YWx1ZSU1Q24lMjAlMjAlMjAlMjBpZiUyMChyZXBvcnRfdmFsdWVzLnJlYXNvbiUyMCElM0QlM0QlMjAnJyklMjAlN0IlNUNuJTIwJTIwJTIwJTIwJTIwJTIwc3dpdGNoU3VibWl0KDEpJTVDbiUyMCUyMCUyMCUyMCU3RCUyMGVsc2UlMjAlN0IlNUNuJTIwJTIwJTIwJTIwJTIwJTIwc3dpdGNoU3VibWl0KDApJTVDbiUyMCUyMCUyMCUyMCU3RCU1Q24lMjAlMjAlMjAlMjByZXR1cm4lMjB0cnVlJTVDbiUyMCU3RCU1Q24lN0QpJTVDbnZhbHVlc1Byb3h5LnJlYXNvbiUyMCUzRCUyMCdmb28nJTVDbiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTdEJTdE

### Expected Behaviour

`#333333` is output to console, which is the same behavior with the input.

### Actual Behaviour

`#B8B8B8` is output to console, which is a different behavior from the input.

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolved #5709 

### Description
For fixing the circular dependency issues, I added `src/ast/utils/identifyNode.ts` file. I think we need to sort out the dependencies of some of the current ast files later, especially where `instanceof` is called.

### DIFF
--- a/src/ast/nodes/shared/knownGlobals.ts
+++ b/src/ast/nodes/shared/knownGlobals.ts
@@ -7,6 +7,12 @@ import {
 	NODE_INTERACTION_UNKNOWN_ACCESS,
 	NODE_INTERACTION_UNKNOWN_ASSIGNMENT
 } from '../../NodeInteractions';
+import {
+	isFunctionExpressionNode,
+	isObjectExpressionNode,
+	isParameterVariableNode,
+	isPropertyNode
+} from '../../utils/identifyNode';
 import type { ObjectPath } from '../../utils/PathTracker';
 import {
 	SymbolToStringTag,
@@ -298,7 +304,25 @@ const knownGlobals: GlobalDescription = {
 		resolve: O
 	},
 	propertyIsEnumerable: O,
-	Proxy: O,
+	Proxy: {
+		__proto__: null,
+		[ValueProperties]: {
+			deoptimizeArgumentsOnCall: ({ args: [, target, parameter] }) => {
+				if (isObjectExpressionNode(parameter)) {
+					for (const property of parameter.properties) {
+						if (isPropertyNode(property) && isFunctionExpressionNode(property.value)) {
+							const [{ variable: firstParameter }] = property.value.params;
+							if (firstParameter && isParameterVariableNode(firstParameter)) {
+								firstParameter.addEntityToBeDeoptimized(target);
+							}
+						}
+					}
+				}
+			},
+			getLiteralValue: getTruthyLiteralValue,
+			hasEffectsWhenCalled: returnTrue
+		}
+	},
 	RangeError: PC,
 	ReferenceError: PC,
 	Reflect: O,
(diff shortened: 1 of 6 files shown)


lukastaegert (approved):
Nice, and thank you for looking into this so quickly. The solution looks good to me!


> For fixing the circular dependency issues, I added src/ast/utils/identifyNode.ts file. I think we need to sort out the dependencies of some of the current ast files later, especially where instanceof is called.

I actually have a permanent fix for this mostly prepared for Rollup 5 on a branch. Basically what I want to do is auto-generated the AST types from the same files that we use to generate the AST converters so that they are perfectly in sync. This will also include the internally used types in the AST classes. This will be based on union types so that then, doing a string comparison for Node.type would actually narrow the TypeScript type. Thus, we can replace all instanceof checks with `.type` comparisons without the need to define manual type guards any more.

PR #5738, opened 2024-11-19
chore(deps): Update `magic-string`

timfish:
`magic-string` has published a new releases to pass through the `debugId` property.
https://github.com/Rich-Harris/magic-string/blob/master/CHANGELOG.md#03013-2024-11-18

Since this gets vendored into the build output, it needs the version to be bumped in the lock file.


lukastaegert (commented):
Sorry I missed this, this should now be fixed by the weekly dependency update anyway. I will see that this is released soon.

PR #5743, opened 2024-11-27
fix: supports modify the import attributes key in the config file

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves https://github.com/rollup/rollup/discussions/5299#discussioncomment-11379479

### Description
Introduce the `configImportAttributesKey` option to control the finalized keyword of import attributes in the config file.
BTW, updated the max Node version from 20 to 22 for github action.


lukastaegert (approved):
Not a nice solution but the best thing to do apart from a breaking change I guess

TrickyPi:
It seems that there is a compatibility issue between higher versions of Node and nyc, and I haven’t found the exact problem yet. So, I temporarily used version 22.0.0.

TrickyPi:
I found the minimum incompatible Node.js version: 22.7.0. The issue was caused by PR (https://github.com/nodejs/node/pull/53619), though I feel this might ultimately need to be resolved on the `nyc` side. As a temporary workaround, enable option `--no-experimental-detect-module` to avoid the problem.

PR #5804, opened 2025-01-17
fix: set hasDeoptimizedCache to true as early as possible

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

related https://github.com/rollup/rollup/issues/5770#issuecomment-2598186894

### Description
Unfortunately, it seems quite difficult to create a minimal reproduction for this issue. If there really is a problem, it’s likely caused by `expression.deoptimizeCache()` re-triggering the current `getLiteralValueAtPath` method. So, In this PR, `hasDeoptimizedCache` is set to true as early as possible to check if that fixes the issue.


lukastaegert (approved):
Looks good and sensible to me

PR #5809, opened 2025-01-23
fix: avoid duplicate rollupjs.org prefix

GauBen:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Not yet, I'll check how to do that

Breaking Changes?

- [x] no

### Description

Urls are currently prefixed with `https://rollupjs.org/` twice

```
(!) If you do not supply "output.name", you may not be able to access the exports of an IIFE bundle.
https://rollupjs.org/https://rollupjs.org/configuration-options/#output-name
```

I'm removing it here, but we could also remove all other `getRollupUrl` calls

### DIFF
--- a/cli/run/batchWarnings.ts
+++ b/cli/run/batchWarnings.ts
@@ -262,7 +262,7 @@ const deferredHandlers: Record<string, (warnings: RollupLog[]) => void> = {
 
 function defaultBody(log: RollupLog): void {
 	if (log.url) {
-		info(getRollupUrl(log.url));
+		info(log.url);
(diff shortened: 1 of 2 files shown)


lukastaegert (approved):
Nice catch! I looked through all usages but I think the one you found is the only wrong one. I took the liberty to adapt the one failing (which stated clearly that the url was wrong, but this is the problem with snapshot testing, someone has to look at them closely enough...).

PR #5820, opened 2025-01-31
return null

kingma-sbw:
Required to add default chunking

This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no


lukastaegert (approved):
While technically not needed in non-TypeScript contexts, it definitely does not hurt.

PR #5838, opened 2025-02-07
replace undefined with void 0 for operator void

TrickyPi:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

related #5775

### Description
Replace `undefined` with `void 0` in unary expressions using the void operator. This way, we can preserve the original shape of the source code, because some people prefer using `void 0` instead of `undefined` in their code.


lukastaegert (approved):
Looks good to me 👍

PR #5875, opened 2025-03-12
capture Rust panic messages and output them to the console.

luyahan:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x ] no

### Description
Capture Rust panic messages and output them to the console.


TrickyPi:
Thanks! I think this package should only be used for WASM, so the correct location might be here.
https://github.com/rollup/rollup/blob/70ef1cce7c740030cc2935b563d13950cc1511f5/rust/bindings_wasm/src/lib.rs#L7
For further improvement, a better approach would be to catch the panic message, convert it to a buffer, and then let Rollup (the JS side) report it, just like how it's handled in a Node addon.

lukastaegert (approved):
Thanks!

PR #5898, opened 2025-04-02
feat: treeshake dynamic import chained member expression

privatenumber:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Here's the test-only PR demonstrating that it fails without this change:
https://github.com/privatenumber/rollup/pull/2

Breaking Changes?

- [x] no

### Description

This PR adds support for tree-shaking dynamic imports when the following pattern is observed:
```ts
import(specifier).then(m => m.name)
```

I didn't look into this deeply, but in a Vite library build, one of my libraries generates this syntax to import from a chunk. When a Vite app (which uses Rollup under the hood) consumes the library, the unused exports from the chunk isn’t tree-shaken.

I'm not sure if this is the root cause, but I thought this change would improve tree-shaking regardless at minimal cost and was worth adding.


melissamforbs (approved):

melissamforbs (approved):

lukastaegert (approved):
While this is kind of a special case, it looks good to me. Technically at this point, we do have a logic that can track inclusions of object properties at this point. However, using for this purpose would require a bigger refactoring as `getDeterministicImportedNames` is only run once, so it would instead need to be triggered somehow from the parameter variables. Maybe we will implement it at some point. Until then, good work!

privatenumber:
Amazing! It's an honor to be able to contribute to Rollup—thank you for all your work.

PR #5909, opened 2025-04-09
Correct the behavior when multiple transform filter options are specified

sapphi-red:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no (I think this is not a breaking change as the previous behavior does not work well)

### Description

When multiple filter options were specified for a transform hook, the hook ran when either of it matched.
For example, the following filter matched for modules that whose id is `foo` or whose code contains `a`.
```js
{
  filter: { id: ['foo'], code: 'a' }
}
```
This is not useful.
This PR fixes that behavior. After this PR, the filter above only matches if the module id is `foo` **and** the code contains `a`.
The previous behavior (that is probably rarely needed) can be achieved by using multiple plugins:
```js
const filters = [{ id: ['foo'] }, { code: 'a' }]
const plugins = filters.map(filter => ({ name: 'plugin', transform: { filter, handler() { /* process the module */ } } }))
```

refs https://github.com/rolldown/rolldown/pull/4059

### DIFF
--- a/src/utils/pluginFilter.ts
+++ b/src/utils/pluginFilter.ts
@@ -76,7 +76,7 @@ function normalizeFilter(filter: StringFilter): NormalizedStringFilter {
 	}
 	if (Array.isArray(filter)) {
 		return {
-			include: ensureArray(filter)
+			include: filter
 		};
 	}
 	return {
--- a/src/utils/pluginFilter.ts
+++ b/src/utils/pluginFilter.ts
@@ -64,7 +64,7 @@ function createFilter(
 		if (include?.some(filter => filter(input))) {
 			return true;
 		}
-		return !!include && include.length > 0 ? false : true;
+		return !(include && include.length > 0);
 	};
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/pluginFilter.ts
```diff
@@ -70,7 +64,7 @@ function createFilter(
 		if (include?.some(filter => filter(input))) {
 			return true;
 		}
-		return !!include && include.length > 0 ? FALLBACK_FALSE : FALLBACK_TRUE;
+		return !!include && include.length > 0 ? false : true;
```
sxzz, line 67:
```suggestion
		return !(include && include.length > 0)
```

sapphi-red:
```suggestion
		return !(include && include.length > 0);
```

lukastaegert (approved):

PR #5947, opened 2025-05-04
refactor: store safe variable names in cache for subsequent usage

Aslemammad:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

While testing the [10000 vite benchmark example](https://github.com/rolldown/benchmarks/tree/main/apps/10000) I found out the main bottleneck was the `deconflictTopLevelVariables` function (avoids using the same name for variables when all the chunks get merged together). Even after introducing the `cache` property which required few changes on the vite side, still that caused a small performance improvement (1s-2s).

deconflictTopLevelVariables shines through the flamegraph

After digging in, I came to find that the cache was only being used in the bundle generation step and not in the `generate` or `write` step where `deconflictTopLevelVariables` -> `getSafeName` sits. 

The idea here is to reduce the usage of `getSafeName` by caching the already generated safe names by the same function. The new cache field is simply named `safeVariableNames` which is an object. This results in a 2x performance improvement (12s -> 5s).

12s -> 5s improvment

Teams can achieve a way better build performance specially in CI/CDs (e.g. deploys) once they start using the `cache` property on rollup.


src/rollup/rollup.ts
```diff
@@ -105,7 +105,9 @@ export async function rollupInternal(
 	timeEnd('BUILD', 1);
 
 	const result: RollupBuild = {
-		cache: useCache ? graph.getCache() : undefined,
+		get cache() {
+			return useCache ? graph.getCache() : undefined;
+		},
```
Aslemammad, line 110: This so we have an updated cache once the `generate`/`write` were called.

lukastaegert (changes requested):
Hi, thanks for the interesting impulse. This is definitely something to consider looking into. Nevertheless, the current approach has some issues. I added a test to demonstrate the most severe one: When a module is changed in a way that would produce a name conflict, the names in the updated module are no longer deconflicted, resulting in an invalid bundle.
In general, we have an issue with reproducibility: With this change, the result no longer depends on the input alone but also on the name cache. If top level variables change in one module (or even just the import order), a fresh build might deconflict differently than a build using the cache. I wonder if it is possible to work around this in some way.

Aslemammad:
hey! thank you so much for the review. 

i'll look into this and try to resolve it. is it just the test case need to be resolved, or we have to inspect the whole general issue? if it's the latter, i'd appreciate some pointers that i should look into since my knowledge of the codebase is pretty limited. 

but anyway, i'll try to figure this out soon.

lukastaegert:
I think it is a more general issue, unfortunately I do not see a simple solution yet. Another problem is if the accessed global variables change in a module, which can influence deconflicting in all other modules. Maybe one needs to figure out a performant way to check if the accessed top-level and global symbols changed in a module.

src/utils/deconflictChunk.ts
```diff
@@ -218,7 +219,19 @@ function deconflictTopLevelVariables(
 	includedNamespaces: ReadonlySet<Module>
 ): void {
 	for (const module of modules) {
+		for (const safeVariableName of Object.values(module.info.safeVariableNames ?? {})) {
+			usedNames.add(safeVariableName);
+		}
+	}
```
Aslemammad: > the names in the updated module are no longer deconflicted, resulting in an invalid bundle.

the reason for this was that we did not reflect the changes that are in the `safeVariableNames` property to the `usedNames` variable which is the key component of the deconflicting idea.

and here's the fix for it.

lukastaegert: Yes, that could work. It means the output depends on the order of changes, i.e. if you have files A and B and then introduce a conflict by changing something in B, then B will be deconflicted while if you introduce a conflict by changing something in A, then A will be deconflicted. That means that running a clean build might return a different output then an incremental build, but maybe that is acceptable.
At the moment though, quite a few tests are red and the bootstrap build where we use the current version of Rollup to build Rollup is also failing.

Aslemammad: i'll work on them and fix them! 

i also might have a potential solution for the order change! and it's by only adding values to the `usedNames` once we're at the same exact module while iterating and not pre-filling it before iterating through all the modules. maybe we can experiment with that later? totally up to you.

Aslemammad: apparently, the test errors happen on master too! 

CleanShot 2025-06-24 at 8  36 55@2x

Aslemammad: oh no it's not! it was somehow using the cached build! it's a problem with my pr. 

working on it right now.

Aslemammad:
@lukastaegert hey! i just fixed the tests. it was because of a special case where the name was `toString` and it was referencing the Object.toString rather than returning undefined. 

I also resolved the order issue we talked about earlier by only adding to usedNames iteratively rather than all in once.

Aslemammad:
sorry i just closed this by misclicking the close with comment button. re-opened now!

test/chunking-form/index.js
```diff
@@ -31,19 +29,17 @@ runTestSuiteWithSamples(
...
+							input: [directory + '/main.js'],
+							onLog: (level, log) => {
+								logs.push({ level, ...log });
+								if (level === 'warn' && !config.expectedWarnings?.includes(log.code)) {
+									warnings.push(log);
+								}
+							},
+							strictDeprecations: true,
+							...config.options
+						});
```
lukastaegert: I think this should be reverted and instead we need to fix the underlying issue, which is that the safeVariableNames currently depend on the format. It is an important feature of Rollup that you can generate different formats from the same bundle, and it should be possible to write a test that shows that the current version can cause errors with that.
I think the only way out is to remove all format-specific exceptions in the naming logic and always disallow those identifiers. Here are the places I can find:
- https://github.com/rollup/rollup/blob/9cebb0b6d9f03d11f53a8c3e7e9e3a79f567e3c6/src/Chunk.ts#L1447-L1472
  all names that are potentially added can be consolidated into the initial usedNames set. Also, it might make sense to extract the corresponding array into a global constant to avoid re-creating it on every run.
- https://github.com/rollup/rollup/blob/9cebb0b6d9f03d11f53a8c3e7e9e3a79f567e3c6/src/ast/scopes/ChildScope.ts#L64-L66
  This one is more tricky. Maybe we just add "exports" to RESERVED_NAMES https://github.com/rollup/rollup/blob/9cebb0b6d9f03d11f53a8c3e7e9e3a79f567e3c6/src/utils/RESERVED_NAMES.ts#L1-L1.
- https://github.com/rollup/rollup/blob/9cebb0b6d9f03d11f53a8c3e7e9e3a79f567e3c6/src/Chunk.ts#L1428-L1434 This one is rather tricky. So in some formats, the actual variable name will be deconflicted, but in other formats it will be "exports". To avoid format specific ambiguity, we would somehow need to rework the logic so that the es/system name is always deconflicted, even though the variable is sometimes rendered as `exports.name`

Aslemammad: i just resolved the first three items (format-specific exceptions). it made some tests fail and i updated those tests and now they're working (aside from the previous failing ones).

but on the forth item (format specific ambiguity), it's pretty hard for me to grasp. can you share an example with me or a specific test that can laser focus me on the general issue. sorry for that again.

Aslemammad:
Sorry, I was pretty shortsighted on this. I put time the last 4 days on it and I think it's in a much better place right now. 

Basically all the tests are working now. Even those that were failing because of the inconsistency between the initial build and the cached build.

Here's a summary. 

- Some tests did not reflect the changes to remove all format-specific exceptions! I changed some of those tests to represent the new state (e.g. `export` -> `export$1`). This fixed most of the tests. 

- The inconsistency that was caused by the cached build was due to misplacing the cache operation. The deconfliction was happening in certain conditions. 

```ts
			if (
				variable.included &&
				// this will only happen for exports in some formats
				!(
					variable.renderBaseName ||
					(variable instanceof ExportDefaultVariable && variable.getOriginalVariable() !== variable)
				)
			) {
```

But the cache was not respecting those conditions and was applied on any variable that basically had the same name because it was above that check. Basically the solution was moving the cache hit into the condition above. here's the [diff](https://github.com/rollup/rollup/pull/5947/files#diff-3544e93a4cc9e89497706da048531b2af8e84cf4040913a42d25885d4b3c4076R224-R237). 

This fixed those tests that were failing due to the inconsistency we previously discussed.

- Another inconsistency that appeared in tests like `no-treeshake-default-export-conflict` and `system-multiple-export-bindings` was mainly due to the export-default statements acquiring a name in `ExportDefaultDeclaration.initialise()`. 


`no-treeshake-export-default-conflict/foo.js`:
```ts
export function foo() {
	return 'foo';
}

export default function() {
	return 'default';
}
```

Here the default export would get the `foo` name as it is the module's name. So we'd have two `foo`s, one the named export function at the top, and one the default export. 

In the caching process through the initial build, this would hit an already created cache entry `foo` and assign `foo$1` instead of keeping the previous value of the previous named function as `foo`. 

In the next build, when the deconfliction process would look for a cached value for the named export `foo`, it'd get the `foo$1` as value, and not `foo`, because it was overwritten in the previous build by the default export. 

The solution was to put an extension `_default` to the default export assigned name to not hit the same cache entry, in cases where a module already has the same name acquired to another variable. Here's the [diff](https://github.com/rollup/rollup/pull/5947/files#diff-7031f7f9e9183f7282e30b44ae27e61a2b53bb5a898c450d4ac06e0584991be2R70-R79).

The same happens in `system-multiple-export-bindings/main.js`. 

```ts
// Variable Declaration
export let a = 1, b = 2, c = 3;
export { a as a2, c as c2 }

// Export default expression
export default a;
```

Here we have two `a`s as well.

The rollup tests are so good at covering these small cases that I personally never thought this PR would hit. Thank you so much @lukastaegert and sorry for having this take so long. Let me know if there's any confusion. I'd really really love to help to get this shipped. 

> This one is rather tricky. So in some formats, the actual variable name will be deconflicted, but in other formats it will be "exports". 

I wonder if this is still needed since now tests are working without it

src/rollup/rollup.ts
joaoninguemgostoso-code: Aslemammad:feat/safeVariableNames

src/ast/nodes/ExportDefaultDeclaration.ts
```diff
@@ -67,11 +67,16 @@ export default class ExportDefaultDeclaration extends NodeBase {
 		const declaration = this.declaration as FunctionDeclaration | ClassDeclaration;
 		this.declarationName =
 			(declaration.id && declaration.id.name) || (this.declaration as Identifier).name;
-		this.variable = this.scope.addExportDefaultDeclaration(
-			this.declarationName || this.scope.context.getModuleName(),
-			this,
-			this.scope.context
-		);
+
```
lukastaegert: Can you say a word about this change? How do you trigger a bug without this change? If so, can we add a specific test for this bug?
I am also asking because this will only deconflict the name in case the variable declaration is above the default export. If you switch the order, no deconflicting will happen as then, the scope will not yet contain the variable. Just in case this is important.

lukastaegert: Ah sorry, overlooked that this is in your previous comment

lukastaegert: Ok, I refactored this and moved setting the name into the "bind" phase where the scope is fully populated. As this name is only for internal use of the variable and does not influence the scope, this is not a problem. Next I might add some tests where we switch between different formats, but maybe tomorrow.

src/utils/deconflictChunk.ts
```diff
@@ -227,10 +224,22 @@ function deconflictTopLevelVariables(
 					(variable instanceof ExportDefaultVariable && variable.getOriginalVariable() !== variable)
 				)
 			) {
+				const cachedSafeVariableName = Object.getOwnPropertyDescriptor(
+					module.info.safeVariableNames,
+					variable.name
+				)?.value;
```
lukastaegert, line 232: I understand you are doing it like this because you do not want to get inherited properties. However, if we use an object without prototype or a Map, we could use simple property access (where I think the object without prototype is probably faster).

src/utils/deconflictChunk.ts
```diff
@@ -227,10 +224,22 @@ function deconflictTopLevelVariables(
...
+
+				if (cachedSafeVariableName && !usedNames.has(cachedSafeVariableName)) {
+					usedNames.add(cachedSafeVariableName);
+					variable.setRenderNames(null, cachedSafeVariableName);
+					continue;
+				}
```
lukastaegert, line 238: I see the check for `!usedNames.has(cachedSafeVariableName)` is the secret here to ensure everything is properly deconflicted. Nice work! So the effect is that you always avoid conflicts, but your variable names might "drift" over several builds. Which I think is fine as long as conflicts are always avoided.

src/utils/deconflictChunk.ts
lukastaegert: Ah I understand, this is because it can be deserialized from JSON, and then you should not tamper with the prototype. Then this is fine!

Aslemammad:
hey @lukastaegert, thank you so much for the merge and sorry for the reply! work on my side has been hectic.

this literally shined my week and i can't believe my eyes. thank you so much for the merge.

Aslemammad:
i'll hopefully write a blog post on thundraa.com, and if you think it'd be appropriate, i'd be happy to host it on the rollup blog/docs instead, let me know.

tariqrafique:
@Aslemammad unfortunately your change broke one of our use cases. 
I've validated that it was broken by this commit. Details in #6357 

For a minor version bump this feels like a breaking change.

PR #5949, opened 2025-05-07
ci: use node 24

btea:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description
https://github.com/nodejs/node/releases/tag/v24.0.0


lukastaegert:
I wonder if the problem is the native support for requiring ES modules. Then maybe we need to skip the test Node >= 22, or run a different test instead.

btea:
It seems that the warning messages given by the node when loading es module in versions 22.11.0 and 24.0.0 are different, resulting in different results of the following judgment logic, which leads to test failure.
https://github.com/rollup/rollup/blob/7536ffb3149ad4aa7cda4e7ef343e5376e2392e1/cli/run/loadConfigFile.ts#L63-L67
<details>
<summary>22.11.0 screen</summary>

image
</details>

<details>
<summary>24.0.0 screen</summary>

image
</details>

lukastaegert:
Ah, then the test is still useful. Nice!

lukastaegert (approved):

Issue #5961, opened 2025-05-21
Incorrect combined sourcemaps for if block close token


sapphi-red:
### Rollup Version

4.41.0

### Operating System (or Browser)

Browser

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-2uwvqcc5?file=rollup.config.js,src%2Fmain.tsx

### Expected Behaviour

![Image](https://github.com/user-attachments/assets/fb0d3702-d51f-46b8-87bf-e7af7b9b4a2b)
The output sourcemap contains a mapping from the closing `}` for if block to the original code.
(the plugin using esbuild is passing the sourcemap with that mapping).

### Actual Behaviour

![Image](https://github.com/user-attachments/assets/7253919d-14c6-439d-8231-b2c936e59359)
The mapping does not exist.

TrickyPi:
This issue is similar to issue #5955. The problem occurs because Rollup's rough sourcemap segments match with esbuild's rough sourcemap segments.

<img width="2915" alt="Image" src="https://github.com/user-attachments/assets/3e25edba-1c08-4b7e-aa85-d5e7c70aa15f" />

This issue might be difficult to solve. We have two potential solutions: either adjust the sourcemap generation logic in the magicString used by Rollup, or modify the sourcemap generation logic in esbuild. I find sourcemap merging to be quite a headache.

TrickyPi:
Yeah, set `hires` to true or 'boundary' could resolve this issue, but the final sourcemap segments is so verbose. As for the second resolution, adding a mapping for '}' in all block statements might be a workable way.

custundag:
Hello @TrickyPi ,

Thank you for your previous responses and for investigating the issue. Do you have an estimated timeline for when the sourcemap issue will be resolved?

PR #5985, opened 2025-06-14
Improve approximation of coarse sourcemap segments

### LINKED ISSUE #5955: Incorrect combined sourcemaps for JSX opening element token
### Rollup Version

4.39.0

Chrome 136

_No response_

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-ymg9zpal?file=package.json

### Expected Behaviour

I'd expect `return` token to be mapped correctly

### Actual Behaviour

opening `<` in JSX gets mapped to `return` token despite esbuild and terser producing correct mappings between `return` tokens, see the breakdown of the visualized mappings here: https://github.com/vitejs/vite/issues/19819#issuecomment-2797022326

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #5955 

### Description
Relevant comment: https://github.com/rollup/rollup/issues/5961#issuecomment-2896825083 

Currently, esbuild is generating some coarse segments, though it may produce more precise ones in the future https://github.com/evanw/esbuild/issues/4189. Since we already have a fallback logic in place, we only need to extend it to cover the esbuild case.


pascalspadone:
> Not 100% sure there might be some performance impact but it looks reasonable to me to go with it.

We did notice a performance impact, I opened https://github.com/rollup/rollup/issues/6056 about it. I linked a repro repo with some measurements; in our real world case the build time increased from about 1 to about 2 minutes. This is caused by a huge 3rd party library that we bundle with our code. During our build, the `traceSegment` function is called about 16.5 million times, with an average of about 700k segments.

We will work around this by excluding the lib from sourcemap generation. I intend to open a PR to add a rollup option to support this (a bit like is done in [this gist](https://gist.github.com/seyanaracore/a627fb40a82cd1bf02d3f275c2cb3d77) ); in the meantime we can work around the issue with `patch-package`.

TrickyPi:
> > Not 100% sure there might be some performance impact but it looks reasonable to me to go with it.
> 
> We did notice a performance impact, I opened #6056 about it. I linked a repro repo with some measurements; in our real world case the build time increased from about 1 to about 2 minutes. This is caused by a huge 3rd party library that we bundle with our code. During our build, the `traceSegment` function is called about 16.5 million times, with an average of about 700k segments.
> 
> We will work around this by excluding the lib from sourcemap generation. I intend to open a PR to add a rollup option to support this (a bit like is done in [this gist](https://gist.github.com/seyanaracore/a627fb40a82cd1bf02d3f275c2cb3d77) ); in the meantime we can work around the issue with `patch-package`.

Thank you very much for reporting this performance regression — we didn’t expect it to have such a big impact. I’m thinking there might still be room for optimization in the fallback logic. If it really can’t be improved, I think we could consider reverting this commit or going with your suggestion to exclude minified JS files.

PR #6005, opened 2025-07-07
fix: export field order prefer esm

DylanPiercey:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

I'm not totally clear on how I would approach testing this 😓.

Breaking Changes?

- [x] no

### Description

The way the `exports` field is configured in the `package.json`'s cause the commonjs version of the package to _always_ be used in environments that pass in both a `require` and `import` condition (most bundled envs).

This is because the way export conditions are matched is in the object order it checks each condition to see if it was provided. In this case since `require` is earlier in the list it is always preferred.

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -32,8 +32,8 @@
   "exports": {
     ".": {
       "types": "./dist/rollup.browser.d.ts",
-      "require": "./dist/rollup.browser.js",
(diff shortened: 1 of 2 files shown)


lukastaegert (approved):
Good catch!

PR #6046, opened 2025-07-30
fix: correctly handle wellknown protocols

cyyynthia:
This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Indirectly related to #2574, as proper handling of `Symbol.hasInstance` is needed to correctly implement it.

### Description

Currently, Rollup gets rid of protocol definitions during object tree-shaking. This PR fixes this by never optimizing `Symbol.*` properties, except the ones that can be optimized (such as `hasInstance`).

Involves the refactor of `ObjectEntity` to make it able to track symbols as actual keys (rather than magic values only).


lukastaegert (commented):
Hi, this looks really intriguing! Note that I will not be able to do releases for the next two weeks, so I will not try to rush a release today in case we need fixes.
One thing I noticed at first glance: Was it really necessary to refactor the argument of ObjectEntity from an object to a Map? After all, you can also use symbols as object keys, which it appears to me to be the main reason (and is IMO the major super-power of Symbols), you just need to allow them in the TS type. You can also use them in literals via `[bracket notation]`. This would reduce the diff quite a bit. Also while Maps are fairly well optimized by the runtime, immutable objects can usually be even better optimized, though the differences should likely be negligible.

cyyynthia:
I refactored to use maps because while objects do allow the use of symbols as key, they're ignored by Object.entries, Object.values, and Object.keys which would've required using Reflect.ownKeys instead. I haven't really checked the performance difference between the two, and figured that it'd be simpler to use a Map that has support for entries and values ootb. I can look into using Reflect.ownKeys if that's preferred.

---

Here is a REPL that illustrates what this PR fixes: https://rollupjs.org/repl/?pr=6046&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmNvbnN0JTIwWCUyMCUzRCUyMCU3QiUyMCU1QlN5bWJvbC5oYXNJbnN0YW5jZSU1RCUzQSUyMCgpJTIwJTNEJTNFJTIwdHJ1ZSUyMCU3RCU1Q24lNUNuaWYlMjAobnVsbCUyMGluc3RhbmNlb2YlMjBYKSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZygnZ290Y2hhJyklNUNuJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlN0QlN0Q=

lukastaegert:
> I refactored to use maps because while objects do allow the use of symbols as key

Ah I see, we were indeed relying on `Object.values` in several places. Alternatively, one could try to wiggle around by using `Object.getOwnPropertySymbols`, but then getting the values would require another mapping. But your solution is less error prone.

fz6m:
IMO We should always preserve these Symbols, otherwise this becomes a breaking change for a few reasons.

1. Across iframes with different JS runtimes, Symbols are used for cross iframe comparison. If they are removed by tree shake, the comparison will no longer work, which is a breaking change.

2. In some UMD builds, these Symbols are consumed by external code. Tree shaking them out will cause external logic to fail, which is also a breaking change.

We should not be this aggressive with tree shaking. For special properties like these, we should take a conservative approach.

cyyynthia:
This is precisely what this PR is introducing. Prior to the PR, symbols would be accidentally removed as part of the treeshaking logic.

As of Rollup 4.54, `Symbol.*` properties are unconditionally conserved by default. They are allowed to be optimized away *only* when there is adequate static analysis in place that's capable of detecting the protocol is used or not.

See the following reproduction:
- [v4.53](https://rollupjs.org/repl/?version=4.53.0&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiY29uc3Qgb2JqID0ge1xuICBrZXkxOiAndmFsdWUxJyxcbiAga2V5MjogJ3ZhbHVlMicsXG4gIFtTeW1ib2wuaGFzSW5zdGFuY2VdOiAodikgPT4gdHJ1ZSxcbiAgW1N5bWJvbC5pdGVyYXRvcl06ICgpID0-IG51bGwsXG4gIFtTeW1ib2wud2hhdGV2ZXJdOiAoKSA9PiBudWxsLFxufVxuXG5jb25zb2xlLmxvZyhudWxsIGluc3RhbmNlb2Ygb2JqKSIsImlzRW50cnkiOnRydWUsIm5hbWUiOiJtYWluLmpzIn1dLCJvcHRpb25zIjp7fX0=)
- [v4.54](https://rollupjs.org/repl/?version=4.54.0&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiY29uc3Qgb2JqID0ge1xuICBrZXkxOiAndmFsdWUxJyxcbiAga2V5MjogJ3ZhbHVlMicsXG4gIFtTeW1ib2wuaGFzSW5zdGFuY2VdOiAodikgPT4gdHJ1ZSxcbiAgW1N5bWJvbC5pdGVyYXRvcl06ICgpID0-IG51bGwsXG4gIFtTeW1ib2wud2hhdGV2ZXJdOiAoKSA9PiBudWxsLFxufVxuXG5jb25zb2xlLmxvZyhudWxsIGluc3RhbmNlb2Ygb2JqKSIsImlzRW50cnkiOnRydWUsIm5hbWUiOiJtYWluLmpzIn1dLCJvcHRpb25zIjp7fX0=)

PR #6072, opened 2025-08-18
build(rust): mimalloc-safe/no_opt_arch on aarch64

### LINKED ISSUE #6047: illegal instruction on android arm64
### Rollup Version

v4.46.1

Android ARM64

v24.4.0

### Link To Reproduction

_Not Available_

### Expected Behaviour

```text
$ pnpm exec rollup

rollup version 4.44.2
=====================================
...
```

### Actual Behaviour

```text
$ pnpm exec rollup
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command was killed with SIGILL (Invalid machine instruction): rollup
```

cyyynthia:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no
- [x] N/A

Breaking Changes?

- [x] no

- closes #6047 (hopefully for good)

### Description

initially introduced by #6055 for linux, this did not include android devices as it is considerd its own operating system with linux being the "vendor"; triplets are quite inconsistent and confusing at times!

this change conservatively disables arch-specific optimizations for all aarch64 targets, to ensure broad compatibility.

### DIFF
--- a/rust/bindings_napi/Cargo.toml
+++ b/rust/bindings_napi/Cargo.toml
@@ -38,7 +38,7 @@ mimalloc-safe = { version = "0.1.54", features = ["local_dynamic_tls"] }
 # Disable architecture specific optimizations on aarch64 platforms
 # Mimalloc assumes `armv8.1-a` is supported, which is not always the case
 # Ref: https://github.com/rollup/rollup/issues/6047
-[target.'cfg(target_arch = "aarch64")'.dependencies]
+[target.'cfg(all(target_arch = "aarch64", not(target_vendor = "apple")))'.dependencies]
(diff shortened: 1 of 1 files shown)


lukastaegert (commented):
Hmm, this would also disable optimizations for aarch64-apple-darwin, which is a very common target that is also fully tested in every pipeline. I would feel better if we could keep optimizations for this one, what do you think? I agree about the other aarch64 targets, though.

cyyynthia:
Ooh Apple targets are a very good point. Updated the dependency condition :)

lukastaegert (approved):
Thanks a lot!

Issue #5090, opened 2023-08-08
'manualChunks' gives inconsistent results for object and function


guaijie:
### Feature Use Case

```js
// using object split chunk
{
  output: {
    manualChunks: {antd: ['antd']}
  }
}
// using function split chunk
{
 output: {
    manualChunks: (id) => {
      if(id.includes('antd')) return 'antd'
    }
}
```
using object split chunk
![image](https://github.com/rollup/rollup/assets/30885718/8b6e3246-038d-4167-98c9-dfcb474dacf6)
using function split chunk
![image](https://github.com/rollup/rollup/assets/30885718/e2767cba-ded8-417e-a152-5e1f337178a1)

Why do we get inconsistent results?
Because module paths are resolved differently.
Code splitting with configuration objects yields module paths, but code splitting with functions yields a set of submodule paths that called 'ids'.
Using functions for code splitting makes the split code unusable due to dependency issues.
So I need the 'id' to be the module path, or to provide an api to get the module path
```
manualChunks: (id, { resolve }) => {
  if(id === resolve('antd')) return 'antd'
}
```
I couldn't get the module path because I didn't know which module resolution was used(eg:  esm or cjs)

### Feature Proposal

```
// first method
manualChunks: (id, { resolve }) => {
  if(id === resolve('antd')) return 'antd'
}
// second method
manualChunks: (id) => {
 // The id is a module path, not a submodule path
  if(id.includes('antd')) return 'antd'
}
```

zhanglolo:
same, 
https://github.com/rollup/rollup/discussions/4992#discussioncomment-6689838

maiieul:
@lukastaegert I believe this can be closed.


```ts
// using object split chunk
{
  output: {
    manualChunks: {antd: ['antd']}
  }
}
// using function split chunk
{
 output: {
    manualChunks: (id) => {
      if(id.includes('antd')) return 'antd'
    }
}
```

Those are not the same even though they look pretty similar.

PR #6106, opened 2025-09-13
feat: unify manualChunks implementation and support object form regular expressions

maiieul:
This PR contains:

- [x] bugfix
- [x] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- resolves https://github.com/rollup/rollup/issues/5090

### Description

- Unifies `manualChunks` object and function form APIs into the function form implementation.
- Allows passing regexes in object form.

This will also reduce the changes needed for https://github.com/rollup/rollup/pull/6087, and simplify the implementation of `onlyExplicitManualChunks`.


src/ast/nodes/Program.ts
```diff
@@ -29,9 +29,6 @@ export default class Program extends NodeBase {
 	private hasLoggedEffect = false;
 
 	hasCachedEffects(): boolean {
-		if (!this.included) {
```
maiieul, line 32: This was re-added in https://github.com/rollup/rollup/pull/4921/files to solve https://github.com/rollup/rollup/issues/4908. This solved the error message but also introduced a test that I think is incorrect. Discussing below on the test itself.

test/chunking-form/samples/manual-chunk-not-included/_config.js
```diff
@@ -1,5 +1,6 @@
 module.exports = defineTest({
-	description: 'handles manual chunks where the root is not part of the module graph',
+	description:
+		'does not output a bundle for the manual chunks that are not part of the module graph',
```
maiieul, line 3: I don't think this counts as a breaking change.

With the behavior after #4921, Rollup will artificially output a bundle for the manualChunks pattern provided if it was not in the module graph, but it won't be used by any bundle, unless Rollup merges some dependencies into it, which was the case in this test.

test/function/samples/emit-chunk-manual-asset-source/main.js
```diff
@@ -1 +1,3 @@
+import './manual';
```
maiieul, line 1: makes sure that manual.js is part of the module graph

test/function/samples/emit-chunk-manual/main.js
```diff
@@ -1 +1,3 @@
+import './manual';
```
maiieul, line 1: makes sure that manual.js is part of the module graph

test/function/samples/manual-chunks-conflict/_config.js
```diff
@@ -1,16 +0,0 @@
...
-		output: {
-			manualChunks: {
-				dep1: ['dep.js'],
-				dep2: ['dep.js']
-			}
-		}
-	},
-	generateError: {
-		code: 'INVALID_CHUNK',
-		message: 'Cannot assign "dep.js" to the "dep2" chunk as it is already in the "dep1" chunk.'
```
maiieul, line 14: This won't happen anymore. I removed the related log error message code.

test/function/samples/manual-chunks-include-external-modules/_config.js
```diff
@@ -1,31 +0,0 @@
...
-							external: true
-						};
-					}
-				}
-			}
-		]
-	},
-	generateError: {
-		code: 'EXTERNAL_MODULES_CANNOT_BE_INCLUDED_IN_MANUAL_CHUNKS',
-		message: `"${externalModule}" cannot be included in manualChunks because it is resolved as an external module by the "external" option or plugins.`
```
maiieul, line 29: This won't happen anymore. I removed the related log error message code.

test/function/samples/manual-chunks-include-external-modules3/_config.js
```diff
@@ -1,24 +0,0 @@
...
-		},
-		output: {
-			manualChunks: {
-				external: [externalModule]
-			}
-		}
-	},
-	generateError: {
-		code: 'EXTERNAL_MODULES_CANNOT_BE_TRANSFORMED_TO_MODULES',
-		message: `${externalModule} is resolved as a module now, but it was an external module before. Please check whether there are conflicts in your Rollup options "external" and "manualChunks", manualChunks cannot include external modules.`
```
maiieul, line 22: This won't happen anymore. I removed the related log error message code.

docs/configuration-options/index.md
```diff
@@ -1452,7 +1452,7 @@ Note that it is not necessary for the listed modules themselves to be part of th
 });
 ```
 
-will put all lodash modules into a manual chunk even if you are only using imports of the form `import get from 'lodash/get'`.
+will put all lodash modules into a `lodash` chunk.
```
maiieul, line 1455: Removed the note about modules not needing to be part of the module graph since this is no longer true. See the `manual-chunk-not-included` test below. 

The PR doesn't affect the behavior outlined in the docs though. I tested on a repro with lodash using `import get from 'lodash/get'` and the modules were properly grouped together into the `lodash` chunk. So I guess the statement wasn't really accurate to begin with.

maiieul:
Closing, because this approach can lead to non-deterministic behavior.

It turns out that
```ts
    manualChunks: {
      vendor: ['lodash-es', 'clsx'],
    },
 ```
and
```ts
    manualChunks: (id) => {
      if (id.includes('lodash-es')) return 'vendor';
      if (id.includes('clsx')) return 'vendor';
    },
 ```
are not the same.

In basic use cases such as in this example it is the same and unlikely to pose a problem, but imagine that the developer passes `vendor: [..., 'utils']`; then that would be converted to `if (id.includes('utils')) return 'vendor';`, which would put all modules that include 'utils' in their path into the same chunk. The object form on the other hand would create an entry for the `utils` file if it can resolve it, and add it to the manualChunk alias.

We cannot use `endsWith` or similar because the id might not end with the pattern exactly (e.g. with `lodash-es` it is `/.../node_modules/lodash-es/lodash.js`, etc.)

I tried to play with some ways to solve the problem, but it seems the best way to be deterministic with the object form is to be through the current logic of `addManualChunks` -> `getChunkAssignments` -> `getChunkDefinitionsFromManualChunks` -> `addStaticDependenciesToManualChunk`.

Issue #6089, opened 2025-09-01
dlopen failed in android arm64


hyperz111:
### Rollup Version

v4.50.0

### Operating System (or Browser)

Android ARM64

### Node Version (if applicable)

24.7.0

### Link To Reproduction

N/A

### Expected Behaviour

```console
~/p $ npx rollup -v
rollup v4.50.0
```

### Actual Behaviour

```console
~/p $ npx rollup -v
/data/data/com.termux/files/home/p/node_modules/rollup/dist/native.js:67
                throw new Error(
                ^

Error: Cannot find module @rollup/rollup-android-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (/data/data/com.termux/files/home/p/node_modules/rollup/dist/native.js:67:9)
    at Object.<anonymous> (/data/data/com.termux/files/home/p/node_modules/rollup/dist/native.js:76:76)
    at Module._compile (node:internal/modules/cjs/loader:1738:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1871:10)
    at Module.load (node:internal/modules/cjs/loader:1470:32)
    ... 4 lines matching cause stack trace ...
    at require (node:internal/modules/helpers:152:16) {
  [cause]: Error: dlopen failed: cannot locate symbol "__emutls_get_address" referenced by "/data/data/com.termux/files/home/p/node_modules/@rollup/rollup-android-arm64/rollup.android-arm64.node"...
      at Module._extensions..node (node:internal/modules/cjs/loader:1898:18)
      at Module.load (node:internal/modules/cjs/loader:1470:32)
      at Module._load (node:internal/modules/cjs/loader:1290:12)
      at TracingChannel.traceSync (node:diagnostics_channel:322:14)
      at wrapModuleLoad (node:internal/modules/cjs/loader:238:24)
      at Module.require (node:internal/modules/cjs/loader:1493:12)
      at require (node:internal/modules/helpers:152:16)
      at requireWithFriendlyError (/data/data/com.termux/files/home/p/node_modules/rollup/dist/native.js:49:10)
      at Object.<anonymous> (/data/data/com.termux/files/home/p/node_modules/rollup/dist/native.js:76:76)
      at Module._compile (node:internal/modules/cjs/loader:1738:14) {
    code: 'ERR_DLOPEN_FAILED'
  }
}

[...]

PR #6109, opened 2025-09-16
fix(rust): use prebuilt std when it is available

cyyynthia:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

- attempts to deal with #6089 

### Description

This change links LLVM's compiler-rt which should hopefully bring in the missing symbols for all platforms.

### DIFF
--- a/.github/workflows/build-and-tests.yml
+++ b/.github/workflows/build-and-tests.yml
@@ -126,6 +126,8 @@ jobs:
           - host: ubuntu-latest
             target: aarch64-linux-android
             cross: napi
+            # There are compile issues when using the cache with Android
+            cache-cargo: false
           - host: ubuntu-latest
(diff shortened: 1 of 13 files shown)


robertkirkman:
When it passes, i would like to know if it is possible for me to test this PR on my device in some way, please let me know

lukastaegert (commented):
Nice! I will trigger a beta build to try this out from the branch and get feedback.

BTW looking at this issue we just received for OHOS which has the same issue in a different context https://github.com/rollup/rollup/pull/6115, another solution might be to also exclude android here. Let's see what works.

lukastaegert:
Released as `rollup@4.50.3-0`

PR #6110, opened 2025-09-16
Add support x86_64-pc-windows-gnu toolchain

### LINKED ISSUE #5439: consider support for msys2/mingw
### Feature Use Case

Currently rollup is completely broken on mingw, see https://github.com/msys2/MINGW-packages/issues/20407, I've seen nobody has made an issue about this. But the msvc build will not work for mingw. There is a node.js build of mingw, [here](https://packages.msys2.org/package/mingw-w64-x86_64-nodejs)

### Feature Proposal

If it is not much effort, It'd be lovely to have support on this platform. or any guidance on how to port.

lsq:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- resolves [#20550](https://github.com/vitejs/vite/issues/20550)
- resolves #5439
- resolves #6053

### Description

Supports node built with gun toolchain in windows.


native.js
```diff
@@ -3,6 +3,7 @@ const path = require('node:path');
 const { platform, arch, report } = require('node:process');
 
 const isMusl = () => !report.getReport().header.glibcVersionRuntime;
+const msvcOrGnu = report.getReport().header.osName.startsWith('MINGW32_NT');
```
lukastaegert: I find the name slightly ambiguous, especially since "true" means that it is GNU Windows. But since it is only used in a single place, why not inline it below, then it is completely clear. Also, then it is not evaluated if not needed.
(admittedly, we could also inline isMusl...)

.github/actions/custom-build/action.yaml
```diff
@@ -0,0 +1,16 @@
...
+  shell:
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Build (Matrix override)
+      shell: ${{ inputs.shell }}
+      env: ${{ matrix.settings.env || fromJSON('{}') }}
+      run: ${{ inputs.build }}
```
lukastaegert: What is the advantage of extracting this when it is only used once and only has a single step? Otherwise, I would find it more straightforward to inline it again.

lukastaegert (commented):
Hi, thanks for taking care. Just two questions.

.github/actions/custom-build/action.yaml
lsq: Build action has been removed and changed to inline.

lsq:
> Hi, thanks for taking care. Just two questions.

Thank you for your suggestions. I have completed the modifications according to your suggestions. Please review them at your convenience.

lukastaegert:
There was a wrong path, I pushed a commit to fix this and triggered a beta release to try it out

lukastaegert:
You can try out rollup@4.51.1-0

lsq:
> You can try out rollup@4.51.1-0
It works. @lukastaegert
image
image

PR #6115, opened 2025-09-19
Disable local_dynamic_tls for OpenHarmony

hqzing:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

On the OpenHarmony platform, if I use a Node.js that is statically linked with libc++ (built with the --partly-static parameter) to load rollup, it works smoothly.

However, if I use a Node.js that is dynamically linked with libc++ (built with the default parameters) to load rollup, it will encounter the following error:

```
Error relocating /data/node_modules/@rollup/rollup-openharmony-arm64/rollup.openharmony-arm64.node: __emutls_get_address: symbol not found
```

To avoid this error, we have two options to choose from:

1. Disable local_dynamic_tls for OpenHarmony.
2. Link libc++ into rollup (refer to here: https://ohos.rs/docs/more/emu-tls.html).

To improve the compatibility of rollup on OpenHarmony and reduce maintenance costs, I chose option 1. This solution also refers to the handling in rolldown(https://github.com/rolldown/rolldown/pull/5258).

### DIFF
--- a/rust/bindings_napi/Cargo.toml
+++ b/rust/bindings_napi/Cargo.toml
@@ -24,7 +24,8 @@ mimalloc-safe = { version = "0.1.54" }
 # 		all(
 # 			target_os = "linux",
 # 			not(target_arch = "loongarch64"),
-# 			not(all(target_arch = "riscv64", target_env = "musl"))
+# 			not(all(target_arch = "riscv64", target_env = "musl")),
+# 			not(target_env = "ohos")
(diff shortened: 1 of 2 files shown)


lukastaegert (approved):
Thanks for looking into this!

Issue #6073, opened 2025-08-18
annotation to treeshake top level await


LongTengDao:
### Feature Use Case

currently, there is no way to treeahake this:

```js
export const a=await import("a")
```

import() can be shaked if we wrap it with `/*#__PURE__*/(()=>import("a"))()`, though a bit of complex. but we can't do anything about the await.

### Feature Proposal

`/*#__PURE__*/await promise`

or

`await/*#__NO_SIDE_EFFECTS__*/promise`

or

new annotation words.

TrickyPi:
@LongTengDao Hi, do you think this PR #6120 could solve your scenario? You can set `moduleSideEffects` to indicate that this `import()` is pure, and if the exports are not used, it can be tree-shaken. But there are some limitations—for example, `await import('a').then()` won’t be tree-shaken.

lukastaegert:
Thinking about this even longer, maybe there is a way to solve this without an annotation and without more analysis. I will need to try out a few things.

PR #6120, opened 2025-09-21
Import expression effect check considers moduleSideEffects

TrickyPi:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

relevant: #6073

### Description
Technically, an `await` expression should always have side effects, since it delays the execution of the following code. For example:
```js
async function a() {
  await 1;
  console.log('a');
}
async function b() {
  console.log('b');
}
a();
b();
```
Here, `'b'` is logged before `'a'`.
But if the argument to `await` is an `import()` expression, I think the `await` expression could be safely removed. In that case, the execution timing of the following code is not deterministic anyway, since it is affected by file I/O or network requests.


lukastaegert:
> In that case, the execution timing of the following code is not deterministic anyway

You are mostly right, but it will never be synchronous, and some logic may rely on the fact that it is async. As this is hard to detect, I would suggest instead to replace it with something compact like an `await 0`. Which in general might be a good simplification for await without other side effects.

TrickyPi:
Oh, I misunderstood before. I get it now—if an async function doesn’t have an `await`, its execution behaves like synchronous code. Thanks for reminding me!

TrickyPi:
After thinking it through, it seems difficult to replace an awaiting import expression with `0`. First, the `AwaitExpression`’s `hasEffects` function has to return `true`, otherwise, the `render` function won't run, and the replacement logic inside it won't be executed. But once we do that, the inclusion of await expressions will always be triggered, and we can’t stop this inclusion from propagating to its argument.

I originally thought `shouldBeIncluded` could stop the propagation, but it turns out it misses some cases. The CI failure revealed the reason: some of the propagation is triggered by the inclusion of other variables.

Maybe we have to narrow the scope of this PR and only take import expressions into account.

lukastaegert:
> and we can’t stop this inclusion from propagating to its argument.

That is not true. For instance, a BlockStatement also does not propagate the inclusion to its children, and I would have hoped this can work similarly. It is wrong, though, to have conditional logic in `includeNode` as it only runs once. Not yet sure what the correct solution would be here.

lukastaegert:
I had another thought about this and I think we should not implement this.
There is another use-case for having a non-awaited or chained dynamic import: To pre-load code earlier so that it is loaded when needed at a later stage.
For that reason, I think we should never remove dynamic imports unless the entire import target is tree-shaken, which is slightly different from the goal of this pull request.

Improving handling of completely tree-shaken modules is still a useful thing, though. Consider the exact scenario where someone uses a bare dynamic import of a module to preload it. With the current logic, this would not include any exports of the target, and if there are not further side effects and there is no other import of that module for whatever reason, we would currently generate something like

```js
Promise.resolve().then(function () { return dep; });

var dep = /*#__PURE__*/Object.freeze({
	__proto__: null
});
```

and no further chunk. This could just be completely removed. But maybe we should close this PR and address it in another PR if we want.

TrickyPi:
Thank you so much for still remembering this PR!

> There is another use-case for having a non-awaited or chained dynamic import: To pre-load code earlier so that it is loaded when needed at a later stage.

Oh, that’s definitely a reason to remove this feature — I totally overlooked that before.

> Improving handling of completely tree-shaken modules is still a useful thing, though. Consider the exact scenario where someone uses a bare dynamic import of a module to preload it. With the current logic, this would not include any exports of the target, and if there are not further side effects and there is no other import of that module for whatever reason, we would currently generate something like
> 
> ```js
> Promise.resolve().then(function () { return dep; });
> 
> var dep = /*#__PURE__*/Object.freeze({
> 	__proto__: null
> });
> ```
> 
> and no further chunk. This could just be completely removed. But maybe we should close this PR and address it in another PR if we want.

Yeah, that definitely removes the module import. Thanks for the work in PR #6230. I haven’t dug into the full implementation yet, but it’s clear that a lot of thought and effort went into it.

PR #6129, opened 2025-09-30
fix(build): make relative imports in amd/cjs relative to the current module

dschmidt:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

If you agree with the change itself, I can try to implement a regression test. Right now I would mainly like feedback if there is a reason for the old implementation. 

Breaking Changes?

- [ ] no

Not sure here, yes, it changes the behavior, but I think it was broken before... so not sure what to check here. 

 - https://github.com/vitejs/vite/issues/20860 would be fixed if copied code was updated in Vite

### Description

Make relative imports in amd/cjs relative to the current module instead of the baseURI. 
This is basically a backport of a change I'm trying to land in vite: 
https://github.com/vitejs/vite/pull/20861

Copied description: 
As far as I can tell rollup itself uses `module.uri` to replace `import.meta.url` in `amd` modules. So I assume it's okay to use it.
https://github.com/rollup/rollup/blob/ce6cb93098850a46fa242e37b74a919e99a5de28/src/ast/nodes/MetaProperty.ts#L209

An alternative could be to use `resolve.toUrl(module.id)` but it seems unneccessary if other parts of the code already rely on `module.uri`.

FWIW I've verified that `import.meta.url` gets transpiled to `new URL(module.uri, document.baseURI).href` in my setup with amd modules. Using `import.meta.url` directly for the imports here does not work as it happens too late in the transpilation process (getting errors that `module` may not be used outside of module scope).

There is more debug output and context in the original ticket/PR

### DIFF
--- a/src/ast/nodes/MetaProperty.ts
+++ b/src/ast/nodes/MetaProperty.ts
@@ -205,7 +205,7 @@ const relativeUrlMechanisms: Record<
 > = {
 	amd: (relativePath, asObject: boolean) => {
 		if (relativePath[0] !== '.') relativePath = './' + relativePath;
-		return getResolveUrl(`require.toUrl('${escapeId(relativePath)}'), document.baseURI`, asObject);
+		return getResolveUrl(`require.toUrl('${escapeId(relativePath)}'), new URL(module.uri, document.baseURI).href`, asObject);
 	},
 	cjs: (relativePath, asObject: boolean) =>
 		`(typeof document === 'undefined' ? ${getFileUrlFromRelativePath(relativePath, asObject)} : ${getRelativeUrlFromDocument(relativePath, asObject)})`,


lukastaegert:
I think the main reason I did not use `module.uri` in the pre-code-splitting wrappers was because I did not know about it. I just skimmed the RequireJS documentation and still could not find a hint of its existence. On the other hand, I just ran test programs in NodeJS and the browser with RequireJS and yes, it does exist.
Which is probably why Guy added it for some of the newer helpers.

I agree that this would be better, and I would even be willing to ship this in a minor release. The only thing to worry about would be that people might have started to work around this bug, and we might break the workflow for those people.
The alternative would be an opt-in flag until we hit the next major version.

Do you have any information since when `module.uri` is supported in RequireJS?

lukastaegert:
Ah, it is mentioned in the wiki https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#magic
And I can find mentions in issues at least back until 2011, so it is apparently well-supported.

dschmidt:
Thanks for your feedback, @lukastaegert :) 

I've updated the snapshots and `test:quick` and `test:only` suceed now.

Now I'm wondering - do we actually need the `new URL(..., document.baseURI).href` wrapper around it?
In my tests `module.uri` was always a url, rendering the wrapper with the base useless. On the other hand documentation in the RequireJS wiki isn't too specific:

> This module gives you information about the module ID and location of the current module: 

🤷🏻 
In then end I just copied it from https://github.com/rollup/rollup/blob/ce6cb93098850a46fa242e37b74a919e99a5de28/src/ast/nodes/MetaProperty.ts#L209 as written above. 

Just let me know if you want me to keep the PR as it is, adjust it in some way or how we want to progress here in general :)

lukastaegert:
In my tests, `module.uri` was a path fragment and not a full URL, so the wrapping was needed. I wonder why it was different for you, maybe you configured a base-url in require.js?
I have a test that basically includes an HTML page for testing which sets it up as
```
<script src="../../../../../node_modules/requirejs/require.js" data-main="_actual/amd/main.js"></script>
```

with no further config and this produces path fragments like `_actual/amd/main.js` for `module.uri`, so I would prefer to leave in the wrapping.

lukastaegert:
Let me know if you have further comments and suggestions, in general I would be fine to release this now.

dschmidt:
Very interesting, thanks for testing thoroughly. Obviously I'm all for the most compatible implementation, if it works for you like this: great. Let's go! 

I don't think we set a base url anywhere, we don't provide it here at least: 
https://github.com/opencloud-eu/web/blob/30c00ff21be634dbc4421eaef3070d14ae6250e7/packages/web-runtime/src/container/application/index.ts#L58


I've fixed the format issue

lukastaegert:
Ah, I just found an error: Your code does not automatically add "module" to the dependencies. I will have a look if I can easily fix it.

dschmidt:
To be fair: I tracked this down in the code that was copied into vite (see linked PR) and didn't reproduce it in rollup itself. Maybe vite behaves somehow different than rollup itself. 

Unfortunately I'm nowhere near a PC to reproduce it for the next 8 days..

dschmidt:
Until then here are some of the values I'm seeing in my environment: 

https://github.com/vitejs/vite/pull/20861#issuecomment-3352230222

dschmidt:
So in my environment:
```
unzip.js:4681 require.toUrl(relativePath) ./../assets/zip-web-worker-CqnRs-V2.js
unzip.js:4682 module.id https://host.docker.internal:9200/assets/apps/unzip/js/unzip.js
unzip.js:4683 module.uri https://host.docker.internal:9200/assets/apps/unzip/js/unzip.js
unzip.js:4684 document.baseURI https://host.docker.internal:9200/
```

i.e. require.toUrl gives me a relative path from the entrypoint of my plugin (/assets/apps/unzip/js/unzip.js) to the web worker, which is correct and makes sense. 
BUT: That path is joined with the baseURI of the application (https://host.docker.internal:9200/) which is not the baseURI of the plugin and not even the base of that (/assets/apps/unzip/) would be correct, as the entrypoint is in a subfolder (/assets/apps/unzip/js/)t. Imho the only correct approach is to join the "dirname" of the module with the relative path, because that's exactly what the relative path is, no?! That's what my patch aims to achieve
Afaict that's also exactly what is happening for other module systems than amd

lukastaegert:
> to join the "dirname" of the module with the relative path

Not sure I understand. After all, the dirname usually starts with something machine specific and it cannot know what the entry point of the server is.

> i.e. require.toUrl gives me a relative path from the entrypoint of my plugin (/assets/apps/unzip/js/unzip.js) to the web worker,

This is not what happens in my local tests, here it really is the path from the module containing `require.toUrl` to the asset, not the entry point. I wonder what is different with your setup?

dschmidt:
> > to join the "dirname" of the module with the relative path
> 
> Not sure I understand. After all, the dirname usually starts with something machine specific and it cannot know what the entry point of the server is.

I lack the proper vocabulary ... I just meant concatenating a relative path from one module to another with the baseURI of the app doesn't make sense whatsoever.  Can we agree on this at least? 


> > i.e. require.toUrl gives me a relative path from the entrypoint of my plugin (/assets/apps/unzip/js/unzip.js) to the web worker,
> 
> This is not what happens in my local tests, here it really is the path from the module containing `require.toUrl` to the asset, not the entry point. I wonder what is different with your setup?

I was using the entrypoint and the web worker as examples, it's not always that ofc
I will try to reproduce this next week...

dschmidt:
> The problem is that require.toUrl(...) should already fix the relative path correctly. From the [docs](https://requirejs.org/docs/api.html):

```
main.js:3 require.toUrl('./assets/logo1-FquQRdV3.svg') _actual/amd/assets/logo1-FquQRdV3.svg
main.js:4 module.uri _actual/amd/main.js
```

whereas in my project I do see:
```
unzip.js:4680 relativePath ../assets/zip-web-worker-CqnRs-V2.js
unzip.js:4681 require.toUrl(relativePath) ./../assets/zip-web-worker-CqnRs-V2.js
unzip.js:4682 module.id https://host.docker.internal:9200/assets/apps/unzip/js/unzip.js
```

Aha! I think the problem is that the `relativePath` passed to and returned by `require.toUrl` isn't relative to the including module, but to the base.
I'm using https://vite.dev/guide/build.html#relative-base set to `./` which gives me relative paths from the including module to the included module. AFAICT rollup doesn't have a builtin feature like that, so I'm not exactly sure it's worth continuing here.
Potentially I could compute the relative path from  `_actual/amd/main.js` to `_actual/amd/assets/logo1-FquQRdV3.svg` at runtime but doing that does not seem very reasonable (code bloat, performance). Building the relative base feature in rollup at build time is probably way more effort than I'm currently willing to spend (especially as I'm not using rollup directly anyhow but using vite)

Unless I'm overlooking there's a way to enable module relative imports in rollup and enable my logic only in that case ... I assume we can close this

What do you think @lukastaegert?

dschmidt:
Fwiw I'll build a reproducer for vite, I'll try to add a rollup config as well and we can compare the behavior... so let's keep thls open a little longer

dschmidt:
@sapphi-red found the culprit in vite: the local require is not injected into the module, so the global require is used which has the global base path.

c.f. https://github.com/vitejs/vite/pull/20861#issuecomment-3426901757

So afaict this is a vite only bug and I can close this PR.

Thanks a lot for your efforts and your diligence.

sapphi-red:
Sorry for redirecting incorrectly, and thank you for taking the time to check this. 🙏

PR #6150, opened 2025-10-20
feat: Implement sourcemapUrlPathPrefix behaviour in sourcemap generation process

guyo13:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description
 

Source maps are often hosted and served separately from the main application bundles, with custom authentication and authorization logic to control access to them.

In such scenarios it is beneficial to be able to change the generated comment of `//# sourceMappingURL=` to include a certain path prefix before the bundle file name so that we can route the download request to another backend service.

For example, if my sourcemaps service is reachable via `<my-domain>/sourcemaps/<bundle-name.js.map>` I could configure Rollup with:
```js
{
output: {
			sourcemapUrlPathPrefix: '/sourcemaps/'
		}
}
```

And then when wanting to download the `bundle-name.js.map` file, the browser will read `//# sourceMappingURL=/sourcemaps/bundle-name.js.map` instead of `//# sourceMappingURL=/bundle-name.js.map`.

#### Why `sourcemapBaseUrl` isn't enough

The `sourcemapBaseUrl` existing option in Rollup only accepts full URLs (`http://example.com[/my-path/]`). This requires knowledge of the actual host on which the built bundles will be deployed.

This constraint is undesirable and usually not suitable for example for containerized apps where we don't want to and can't constrain the built image to the actual host URL. An example of this scenario is deploying the built app onto different deployment environments (test, stage, prod, etc...) where the sourcemaps are also served from the environment's sourcemaps service.   

This PR adds the `output.sourcemapUrlPathPrefix` option with this exact use case in mind. It also sanely and clearly handles the case where a `sourcemapBaseUrl` is specified ALSO with a url prefix (`http://example.com/pre-prefix`). In that case the `output.sourcemapUrlPathPrefix` option will be appended and the resulting url will be `http://example.com/pre-prefix/${output.sourcemapUrlPathPrefix}`.


.prettierignore
```diff
@@ -1,2 +1,3 @@
 cli/help.md
 LICENSE.md
+slugs-and-pages-by-legacy-slugs.json
```
guyo13, line 3: added this to preserve the one-line formatting of this file

guyo13:
If this is merged I will also open a PR for Rolldown to implement the same feature.

.prettierignore
TrickyPi: ~~I think this change might not be needed, as `slugs-and-pages-by-legacy-slugs.json` should be automatically generated by the `npm run build:docs` command.~~

lukastaegert (commented):
Nice! I totally see the use case. But I am slightly worried about the inflation of options in this space. As I see it, there is no situation where you would want to use both `sourcemapUrlPathPrefix` and `sourcemapBaseUrl` together. And since we ensure that we always add a trailing slash to the sourcemapBaseUrl anyway, I wonder if instead we could just rework sourcemapBaseUrl to support also strings that just start with a "/"? Of course we could not longer simply use `new URL(sourcemapFileName, sourcemapBaseUrl)`, but considering that souremapFileName is always of the form (I think) `some/path/fileName.map`, we may not need any fancy resolution logic and could just always do some concatenation. What do you think?

.prettierignore
guyo13: I will double check it. Since the file is committed I thought it might be useful :)

TrickyPi: It appears that `slugs-and-pages-by-legacy-slugs.json` is generated based on `legacy-slugs.json`, as shown in `verify-anchors.ts`.
That said, I’m still not entirely sure about the exact purpose or effect of these two JSON files.

TrickyPi: Only modifying `slugs-and-pages-by-legacy-slugs.json` will cause a docs build error — we can see the error by running `npm run build:docs`.

docs/configuration-options/index.md
```diff
@@ -1745,6 +1745,15 @@ If `true`, a separate sourcemap file will be created. If `"inline"`, the sourcem
...
 By default, sourcemap files generated by Rollup uses relative URLs to reference the files they describe. By providing an absolute base URL, e.g. `https://example.com`, sourcemaps will use absolute URLs instead.
 
+### output.sourcemapUrlPathPrefix
+
+|       |                                              |
+| ----: | :------------------------------------------- |
+| Type: | `string`                                     |
+|  CLI: | `--sourcemapUrlPathPrefix <url-path-prefix>` |
+
+When specified, Rollup will prefix sourcemap URLs paths with the given string. If specified, must include a leading `/`, e.g. `/my-sourcemaps-server`. Can be used in conjunction with [sourcemapBaseUrl](#outputsourcemapbaseurl), when both are specified the `sourcemapUrlPathPrefix` gets appended to the path present in `sourcemapBaseUrl`.
```
TrickyPi, line 1755: `#outputsourcemapbaseurl` should be changed to `#output-sourcemapbaseurl`.

guyo13: thanks! wasn't aware of the build:docs script. I'll change it

guyo13:
Thats the exact dilemma I was facing. 
You're right about `souremapFileName` (actually it is always a bare file name (rather than a path) because of the `basename` call), but it will probably wont be safe to just concatenate because if the file name contained some invalid URL characters (e.g space) they are currently not uri-encoded by Rollup, and it is the `new URL` which encodes them (might be worth adding a comment to the code because this process is very implicit).

This actually reveals a small bug in my original implementation and in Rollup (if `sourcemapFileNames` is specified without `sourcemapBaseUrl` which would have it encoded. See reproduction example below*) -
because in `emitSourceMapAndGetComment` if we pass a `sourcemapUrlPathPrefix` without a `sourcemapBaseUrl` we have:

```typescript
const sourcemapFileName = basename(fileName);
		const urlPath = sourcemapUrlPathPrefix.  // THIS IS NOT URI ENCODED. e.g "/my sourcemaps/"
			? sourcemapUrlPathPrefix + sourcemapFileName  // THEREFORE THIS IS NOT URI ENCODED
			: sourcemapFileName;

		if (sourcemapBaseUrl) {
			const baseUrl = new URL(sourcemapBaseUrl);
			const combinedPath =
				baseUrl.pathname + (urlPath.startsWith('/') ? urlPath.slice(1) : urlPath);
			url = new URL(combinedPath, baseUrl.origin).toString();
		} else {
			url = urlPath; // THEREFORE THIS IS NOT URI ENCODED
		}
/// rest of the code
```

#### * Bug - when `sourcemapFileNames` is specified without `sourcemapBaseUrl`
```javascript
const assert = require('node:assert');

module.exports = defineTest({
	description: 'unencoded sourcemapFileNames',
	solo: true,
	options: {
		output: {
			// sourcemapBaseUrl: 'https://example.com/a/',
			sourcemapFileNames: 'bundle. [format].js.map'
		}
	},
	test: (code, map, { format }) => {
		assert.equal(map.file, `bundle.${format}.js`);
		const sourceMappingURL = code.split('\n').slice(-2)[0];
		assert.equal(sourceMappingURL, `//# sourceMappingURL=bundle.%20${format}.js.map`); // Fails due to unencoded space in filename
	}
});
//      AssertionError [ERR_ASSERTION]: '//# sourceMappingURL=bundle. amd.js.map' == '//# sourceMappingURL=bundle.%20amd.js.map'
//      + expected - actual

//      -//# sourceMappingURL=bundle. amd.js.map
//      +//# sourceMappingURL=bundle.%20amd.js.map
```

So all in all I think we should first decide whether we want to fix the above bug (it exist even without my feature contribution) or that Rollup won't validate/encode - say spaces in filenames - by definition.

Then, if the decision is not to validate, I can go on with the simple concatenation-based implementation that would enable `sourcemapBaseUrl` to support also strings that just start with a "/". **This will probably break** users that are currently using  `sourcemapBaseUrl` and **are implicitly or explicitly** relying on the `new URL` call sanitizing the URL for them.

Otherwise, if Rollup does want to validate  `sourcemapFileNames`, I think it merits a discussion on how to do this - the simplest approach is to use `encodeURIComponent` on the `basename`-d file (unless I am missing something). Or rather we could take `new URL(sourcemapFileName, "http://example.com").pathname` (note that with a url of `bundle.js.map` this will return `/bundle.js.map`).

Sorry for the long post, I hope it makes sense :)

lukastaegert:
Hi, sorry for losing track of this PR a little. Your points are very good, and for a major version bump, it would make sense to align one way or another.
With regard to implicit URL-encoding, I would rather NOT do this in the future. While it is most of the time in their interest to have valid urls, it is also a hidden magic that users cannot control, and maybe sometimes they just want to "hack" something together.
To avoid the breaking change for this small change, what if we branch the logic: If `sourcemapBaseUrl` starts with a `/`, then we concatenate, otherwise we use the old logic. And we make a note for the next major version to always do concatenation. What do you think?

Issue #6154, opened 2025-10-24
Top-level await and (non-top-level) dynamic import leads to spurious cycle


alexmoon:
### Rollup Version

4.52.1-4.52.5

### Operating System (or Browser)

all

### Link To Reproduction

https://rollupjs.org/repl/?version=4.52.5&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiaW1wb3J0IHsgQmFyIH0gZnJvbSBcIi4vYmFyLmpzXCJcblxubmV3IEJhcigpLmJhcigpIiwiaXNFbnRyeSI6dHJ1ZSwibmFtZSI6Im1haW4uanMifSx7ImNvZGUiOiJhd2FpdCBQcm9taXNlLnJlc29sdmUoKTtcblxuZXhwb3J0IGNsYXNzIEZvbyB7XG4gIGFzeW5jIGJhcigpIHtcbiAgICBjb25zdCBiYXIgPSBhd2FpdCBpbXBvcnQoXCIuL2Jhci5qc1wiKTtcbiAgICByZXR1cm4gbmV3IGJhci5CYXIoKTtcbiAgfVxufSIsImlzRW50cnkiOmZhbHNlLCJuYW1lIjoiZm9vLmpzIn0seyJjb2RlIjoiaW1wb3J0IHsgRm9vIH0gZnJvbSBcIi4vZm9vLmpzXCJcbmV4cG9ydCBjbGFzcyBCYXIgZXh0ZW5kcyBGb28ge30iLCJpc0VudHJ5IjpmYWxzZSwibmFtZSI6ImJhci5qcyJ9XSwib3B0aW9ucyI6e319

### Expected Behaviour

We expect to be able to use dynamic imports to break cycles between modules.

### Actual Behaviour

Rollup generates a circular dependency warning.

In our app, the dependency cycle is much more involved than in the repl and the modules end up being combined in an invalid order which leads to undefined identifier exceptions at runtime. The issue was introduced with the changes in #6121.

I don't know if it's possible for the heuristics in #6121 to be modified to only apply to dynamic imports at the top level or if a config option could be used to opt out of that behavior?

TrickyPi:
In your reproduction link, the compiled code produces the same result as the original source. I’m curious to see the source code that led to the compiled version which caused the execution error.

alexmoon:
[...]

PR #6221, opened 2025-12-26
Handle TLA dynamic import cycles in chunkAssignment

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

related to #6154 

### Description
The reason this issue was raised is that a dynamic import is treated as a TLA dynamic import when it appears in a module with TLA. As a result, an otherwise normal module graph is considered to contain cycles.

In my opinion, a bundler generally should not care about cycles introduced by TLA dynamic imports, other than possibly emitting a warning. However, in the case described in the issue, this TLA dynamic import cycle is a false positive.

Therefore, in this PR, introducing a fallback logic for chunk assignment to handle these scenarios. But the warning caused by the false positive still remains.

### DIFF
--- a/src/utils/chunkAssignment.ts
+++ b/src/utils/chunkAssignment.ts
@@ -173,7 +173,11 @@ export function getChunkAssignments(
 
 	// Each chunk is identified by its position in this array
 	const chunkAtoms = getChunksWithSameDependentEntries(
-		getModulesWithDependentEntries(dependentEntriesByModule, modulesInManualChunks)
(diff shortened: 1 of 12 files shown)


lukastaegert (approved):
Thanks! While I am not convinced this solution is perfect, I remember having spent some time already thinking about how to improve the handling of TLA situations with no good results. More chunks is usually the answer, and this is definitely a good first step.

Issue #6189, opened 2025-11-24
Cannot Read Properties of undefined - files from the same module can't access eachothers exports/imports


AncientSwordRage:
### Rollup Version

4.52.0

### Operating System (or Browser)

Linux

### Node Version (if applicable)

20+

### Link To Reproduction

https://stackblitz.com/edit/vitejs-vite-frdakn6r?file=package.json

### Expected Behaviour

Files from the same module should be able to access eachothers exports/imports

### Actual Behaviour

export variable is not shared

----

I'm using Vite v6 (which uses rollup) to build my project using Rollup's manual chunks functionality. I have put anything with 'react' in one chunk and every other vendor file into another chunk.

I don't think this is an issue with my vite set up?

I can build the project, but I can see from the errors in my dev console that code that the 'index.js' file tries to require from `'../lib/possibleStandardNames'` isn't accessible?

```
// possibleStandardNames.js for react-property
var ud = Iv(),
  ft = {},
  be = {},
  z1;
function Pv() {
  if (z1) return be;
  z1 = 1;
  var q = 0;
  be.SAME = q; // error here
  var nl = 1;
  return (
    (be.CAMELCASE = nl),
    (be.possibleStandardNames = {
      // ... object continues
    }),
    be
  );
}
// index.js for react-property
var E1;
function ed() {
  if (E1) return ft;
  E1 = 1;
  const q = 0,
    nl = 1,
    vl = 2,
    m = 3,
    dl = 4,
    hl = 5,
    pl = 6;
  function Ul(p) {
    return z.hasOwnProperty(p) ? z[p] : null;
  }
 // etc...
```
and later on, when the index file tries to import from `possibleStandardNames`:

```
const { CAMELCASE: mt, SAME: ot, possibleStandardNames: Qt } = Pv(),
```

it would be better to inline `possibleStandardNames` contents into the `index` if they're in the same chunk?

This issue gives this 'tell tale' stack trace that I've seen numerous times while massaging my manual chunks:

```
react-f2prr9Ke.js:12309  Uncaught TypeError: Cannot set properties of undefined (setting 'SAME')
    at Pv (react-f2prr9Ke.js:12309:11)
    at ed (react-f2prr9Ke.js:13015:66)
    at Ne (vendor-DIrDj7Dq.js:10:3731)
    at vendor-DIrDj7Dq.js:10:6029
    at Xe (vendor-DIrDj7Dq.js:10:6681)
[...]

PR #6225, opened 2026-01-01
Throw a warning for circular chunks

TrickyPi:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

Related to #6189

### Description

Code splitting is one of the most important features of a bundler. For example, manual bundling is often used for rarely changed third-party packages like `React` or `Vue`. However, most users do not know the dependency graph of these third-party packages, so the `manualChunks` option may be configured incorrectly, as in this issue.

In this PR, we introduce a warning for circular chunks, which may be caused by two scenarios. First, manual chunks conflict with each other. In this case, Rollup has no room for further improvement. Second, the issue is caused by `output.onlyExplicitManualChunks`. In this case, we can do better in the future by bundling the static dependencies of manual chunks into a separate chunk, if bundling them into the entry module would cause circular chunks.

### DIFF
--- a/src/Bundle.ts
+++ b/src/Bundle.ts
@@ -21,6 +21,7 @@ import {
 	error,
 	logCannotAssignModuleToChunk,
 	logChunkInvalid,
+	logCircularChunk,
 	logInvalidOption
 } from './utils/logs';
 import type { OutputBundleWithPlaceholders } from './utils/outputBundle';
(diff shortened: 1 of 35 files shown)


lukastaegert (approved):
Looks good to me.

PR #6247, opened 2026-01-24
add support for OpenBSD/amd64

rnagy:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### DIFF
--- a/native.js
+++ b/native.js
@@ -41,7 +41,7 @@ const bindingsByPlatformAndArch = {
 		x64: { base: 'linux-x64-gnu', musl: 'linux-x64-musl' }
 	},
 	openbsd: {
-		arm64: { base: 'openbsd-arm64' }
+		arm64: { base: 'openbsd-arm64' },
 		x64: { base: 'openbsd-x64' }
 	},
 	openharmony: {
--- a/native.js
+++ b/native.js
@@ -41,6 +41,7 @@ const bindingsByPlatformAndArch = {
 		x64: { base: 'linux-x64-gnu', musl: 'linux-x64-musl' }
 	},
 	openbsd: {
+		arm64: { base: 'openbsd-arm64' }
 		x64: { base: 'openbsd-x64' }
 	},
 	openharmony: {
--- a/npm/openbsd-arm64/README.md
+++ b/npm/openbsd-arm64/README.md
@@ -1,0 +1,3 @@
+# `@rollup/rollup-openbsd-arm64`
+
+This is the **aarch64-unknown-openbsd** binary for `rollup`
--- a/npm/openbsd-arm64/package.json
+++ b/npm/openbsd-arm64/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "@rollup/rollup-openbsd-arm64",
+  "version": "0.0.0",
+  "os": [
+    "openbsd"
+  ],
+  "cpu": [
+    "arm64"
+  ],
+  "files": [
+    "rollup.openbsd-arm64.node"
+  ],
+  "description": "Native bindings for Rollup",
+  "author": "Lukas Taegert-Atkinson",
+  "homepage": "https://rollupjs.org/",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/rollup/rollup.git"
+  },
+  "main": "./rollup.openbsd-arm64.node"
+}
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
       "aarch64-unknown-freebsd",
       "aarch64-unknown-linux-gnu",
       "aarch64-unknown-linux-musl",
+      "aarch64-unknown-openbsd",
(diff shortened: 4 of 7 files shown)


native.js
```diff
@@ -41,6 +41,7 @@ const bindingsByPlatformAndArch = {
 		x64: { base: 'linux-x64-gnu', musl: 'linux-x64-musl' }
 	},
 	openbsd: {
+		arm64: { base: 'openbsd-arm64' }
```
lukastaegert: There is a comma missing
```suggestion
		arm64: { base: 'openbsd-arm64' },
```

rnagy: ah right, fixed, thanks

lukastaegert:
In order for this to build, we also need to adapt the GitHub pipelines

ikmckenz:
I think the github action openbsd section could be updated as follows: 

```yaml
  build-openbsd:
    name: Build OpenBSD
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        os:
          - name: openbsd
            architecture: x86-64
            version: '7.8'
            target: x86_64-unknown-openbsd
          - name: openbsd
            architecture: arm64
            version: '7.8'
            target: aarch64-unknown-openbsd

    steps:
      - name: Checkout Commit
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2

      - name: Restore Cargo cache
        uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
        with:
          path: |
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            .cargo-cache
            rust/target/
          key: ${{ matrix.os.target }}-cargo-ubuntu-latest-${{ hashFiles('rust/Cargo.lock') }}
          restore-keys: ${{ matrix.os.target }}-cargo-ubuntu-latest

      - name: Install and Cache Node Dependencies
        uses: ./.github/actions/install-and-cache-node-deps

      - name: Build (OpenBSD)
        uses: cross-platform-actions/action@492b0c80085400348c599edace11141a4ee73524 # v0.32.0
        with:
          operating_system: ${{ matrix.os.name }}
          architecture: ${{ matrix.os.architecture }}
          version: ${{ matrix.os.version }}
          run: |
            sudo pkg_add rust node cmake
            sed -i 's^rustflags.*^^g' rust/bindings_napi/.cargo/config.toml
            npm run build:napi

      - name: Save Cargo cache
        if: github.ref == 'refs/heads/master'
        uses: actions/cache/save@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
        with:
          path: |
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            .cargo-cache
            rust/target/
          key: ${{ matrix.os.target }}-cargo-ubuntu-latest-${{ hashFiles('rust/Cargo.lock') }}

      - name: Upload napi artifact
        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
        with:
          name: bindings-${{ matrix.os.target }}
          path: ${{ env.APP_NAME }}.*.node
          if-no-files-found: error
```

Edit: I've tested the above config in my own fork to run the github actions (https://github.com/ikmckenz/rollup/pull/1) and it does work, but the aarch64 build is sloooooww. x86_64 takes ~6 mins to build, and aarch64 takes ~1 hour 45 minutes. I suspect it could be a networking issue, it takes the build 20 minutes to setup the VM (bulk seems to be spent in the rsync), then another long time in the pkg_add stage, then the build is very slow on compiling. I also tested with vmactions/openbsd-vm action and it was the same result. I am testing vendoring the rust now, but unfortunately that doesn't fix the issue, so perhaps not just network slowness.

Miniontoby:
> I think the github action openbsd section could be updated as follows:

Your cargo cache uses a save path that mentions ubuntu (`${{ matrix.os.target }}-cargo-ubuntu-latest-${{ hashFiles('rust/Cargo.lock') }}`)

Which doesnt sound right, but of course you made a suggestion, but if anyone is going to add that, then that's one thing that might need to be changed first

ikmckenz:
> Your cargo cache uses a save path that mentions ubuntu (`${{ matrix.os.target }}-cargo-ubuntu-latest-${{ hashFiles('rust/Cargo.lock') }}`)
> 
> Which doesnt sound right, but of course you made a suggestion, but if anyone is going to add that, then that's one thing that might need to be changed first

Yeah that is actually in the existing code too, don't think it causes issues, but probably should be cleaned up: https://github.com/rollup/rollup/blob/186b352f3b028c5733c47dadacd565e303567e9d/.github/workflows/build-and-tests.yml#L336

PR #6251, opened 2026-01-29
fix: Isolate and cache `process.report.getReport()` calls in a child process for robust environment detection

alan-agius4:
This PR contains:

- [x] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

Not in this repo.
For some details https://github.com/angular/angular-cli/issues/32234

### Description
This refactoring isolates the `process.report.getReport()` call into a separate child process. This change is necessary because since the introduction of this code, we have observed frequent process crashes on Windows, specifically with the error code -1073740940 (0xC0000374), which corresponds to a Windows heap corruption error.

By moving the report generation to an isolated process, any potential instability or memory corruption is contained and does not affect the main process.

The retrieved report header is cached to ensure that the child process is spawned only once, making this a fast and efficient solution for environment detection.


lukastaegert:
Nice catch. In my experience, though, spawning child processes has quite some performance overhead, which will negatively affect startup time. Maybe we can limit this to windows environments, where the issue was observed, and use the regular call on Linux?

alan-agius4:
> Nice catch. In my experience, though, spawning child processes has quite some performance overhead, which will negatively affect startup time. Maybe we can limit this to windows environments, where the issue was observed, and use the regular call on Linux?

SGTM

lukastaegert (approved):
Thanks, looks good.

PR #6257, opened 2026-02-01
[v5.0] Put binary AST buffers into cache

lukastaegert:
This PR contains:

- [ ] bugfix
- [x] feature
- [x] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no

### Description

This finally removes the need to generate full ESTree ASTs at all in many situations and directly stores the binary buffers in the cache.
All generated ASTs except the ones returned by this.parse and the parseAst(Async) helpers will now be generated lazily, which means that all non-primitive properties of AST nodes are getters that are replaced with the actual properties on first access. This should severely reduce processing time if only some of the nodes are needed.
This will likely have quite some performance impact, especially when a standard ESTree-Walker is used. There might be a follow-up PR to provide an efficient interface to walk an AST without materializing nodes that are not needed.

Additionally, it exposes new helpers for manually converting an AST to binary and back.

#### Exports from `rollup/parseAst`

- `parseAst` / `parseAstAsync` — return eager (fully materialized) ASTs
- `parseLazyAst` / `parseLazyAstAsync` — new helpers returning lazy ASTs (same signatures as `parseAst`/`parseAstAsync`)
- `deserializeAst` — new helper that deserializes a binary buffer to an eager AST (symmetric counterpart to `deserializeLazyAst`)
- `deserializeLazyAst` — new helper that deserializes a binary buffer to a lazy AST
- `serializeAst` — create binary buffer from a JSON AST

The lazy and non-lazy deserializers have interchangeable signatures.

#### BREAKING CHANGES

* The cache is no longer purely JSON but instead JSON with Buffers/typed arrays in some fields. If people want to store this, it should still be possible to do via a JSON.stringify implementation with a custom serializer and storing the buffers in separate files.


lukastaegert:
I will improve coverage in #6271 further

PR #6271, opened 2026-02-20
[v5.0] Implement efficient AST walking API

lukastaegert:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description
With #6257, one issue is that in order to have actual performance improvements, we need to avoid materializing the AST if possible. However, if a plugin like the commonjs plugin uses an ast-walker to find relevant nodes, this will then lazily materialize the AST, making everything even slower (though I will probably move this.parse to return a non-lazy AST again for this reason).
So here is a new, high performance API for parsing and walking an AST without materializing any unneeded nodes. This is how it works

```js
this.parseAndWalk(code, {
	async Identifier(node, { parseChildren, skipChildren }) {
		// node is typed as Identifier
		console.log('Found identifier:', node.name);
	},
	async CallExpression(node, { parseChildren, skipChildren }) {
		// node is typed as CallExpression
		if (
			node.callee.type === 'Identifier' &&
			node.callee.name === 'require'
		) {
			// Process require() calls
			parseChildren(); // Visit arguments
			// Now perform any logic that relies on the arguments being visited
			// ...
		} else {
			skipChildren(); // Do not visit arguments for other calls
		}
	}
});
```

In the background, Rollup will first transform the set of queried nodes into a 128bit bitset and send this to Rust. In Rust in another thread, the parsing logic will additionally build a buffer with "walking information", which is basically a list which nodes should be traversed in which order and to which child it should jump if parseChildren or skipChildren is used.
When the vistor handlers are called, a lazy AST nodes is conjured up. I think this is one of the fastest way to walk an AST in JavaScript with ESTree-compliant nodes.

In order to make this fully useful for the CommonJS plugin, we also need to add sufficient information to detect global variables, which I currently still think about how to implement it.


lukastaegert:
I also added a functionality to check if an identifier is contained in a scope. This is now ready for merge from my side.

PR #6314, opened 2026-03-23
feat: replace __proto__: null with Object.setPrototypeOf in generated code

guimard:
## Summary

Replace `{ __proto__: null, ... }` object literals with `Object.setPrototypeOf({ ... }, null)` in all generated output code.

The `__proto__` property in object literals does not work when Node.js runs with `--disable-proto=throw`, which is used in some environments (e.g. Debian package testing). `Object.setPrototypeOf` is the proper API that works in all environments and preserves the same semantics: creating a null-prototype object for namespace and interop helpers.

This affects three code generation paths:
- `src/ast/variables/NamespaceVariable.ts` — namespace objects
- `src/utils/interopHelpers.ts` — `_interopNamespaceDefaultOnly`
- `src/finalisers/system.ts` — star re-export setters and `_starExcludes`

Ref: [Debian#1093290](https://bugs.debian.org/1093290)


lukastaegert:
I see the problem, though I do not understand at all why an environment would disallow `__proto__` but not `Object.setPrototypeOf`. I would really like to avoid this massive refactoring, though.
Besides that it is much less nice to read, there is also usually a performance penalty of switching the prototype AFTER object generation. The `__proto__` syntax in object literals is well-optimized and directly creates the objects as we want them.
I know there are good reasons to forbid assignments to `__proto__`, but disallowing object literal usage, which is definitely not the same thing, is stupid IMO.
I guess we will need to move away from this syntax eventually anyway, but not like this.

PR #6350, opened 2026-04-11
fix: reset variable render names between outputs in the same generate

### LINKED ISSUE #6296: Invalid UMD output when mixed outputs include CJS manualChunks: generated code contains illegal syntax like `function vendor.xxx()`
### Rollup Version

4.59.0

MacOS 14.6.1

20.19.6

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-iaxxfmyq

### Expected Behaviour

UMD output should remain valid JavaScript regardless of other output entries.
Mixing umd and cjs outputs should not cause invalid identifier/function declarations in UMD codegen.

Config:
```js
import nodeResolve from '@rollup/plugin-node-resolve';

export default {
  input: './src/main.js',
  output: [
    {
      name: 'main',
      file: './dist/index.js',
      format: 'umd',
    },
    {
      dir: './dist/',
      format: 'cjs',
      manualChunks: (id) => {
        if (/node_modules/.test(id)) {
          return 'vender';
        }
      },
    },
  ],
  plugins: [nodeResolve()],
};
```

Entry:
```ts
import { _ as type_of } from '@swc/helpers/esm/_get.js'

console.log(type_of)
```

### Actual Behaviour

UMD output contains invalid syntax (function vender.xxx()) and cannot be parsed/executed.

Build output
```js
// ...

        return object;
    }

    function vender._get(target, property, receiver) {
//                 ^ Error Here
        if (typeof Reflect !== "undefined" && Reflect.get) vender._get = Reflect.get;
        else {

// ...
```

barry3406:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] yes
- [x] no

Fixes #6296

Description:

When the same input bundle is generated for multiple outputs in a single `output` array, `Variable.renderBaseName` set during one output's import deconfliction was left on the shared `Variable` instance and leaked into the next output.

Concretely: generating a chunked CJS output calls `deconflictImportsOther`, which sets `setRenderNames(chunk.variableName, exportName)` on imported variables — so a function `_` imported from a `vendor` chunk ends up with `renderBaseName='vendor'`, `renderName='_'`. The state lives on the shared `Variable` instance, not on a per-output structure. When the next output (UMD) goes through `deconflictTopLevelVariables`, the `!variable.renderBaseName` skip leaves the stale `vendor` in place, and the local function declaration ends up rendered as `function vendor._(...)` — invalid JavaScript.

Fix: reset `renderBaseName`/`renderName` to `null` for top-level variables of the chunk's modules at the start of `setIdentifierRenderResolutions`, before either the exports loop or `deconflictChunk` reassigns them. Each output then starts from a clean slate.

Added a regression test in `test/misc/misc.js` that exercises the exact scenario from #6296: a single bundle generated as both chunked CJS (with a `vendor` manualChunk) and UMD, asserting the UMD output contains no dotted function declarations and is parseable as valid JavaScript.


lukastaegert (approved):
Sorry for missing this, this looks like a really nice fix. Will merge and release it once pipelines agree.

PR #6361, opened 2026-04-19
chore: remove tslib from devDeps

K-tecchan:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

#### Summary
 Remove tslib from devDependencies and exclude it from the generated LICENSE.md.

 #### Background
 tslib was added as a dev dependency in 4678513 (2018) when the project switched to rollup-plugin-typescript. However, TypeScript's importHelpers compiler option has never been enabled in this project's tsconfig.json. Without importHelpers: true, TypeScript inlines helper functions directly into each file rather than importing them from tslib, meaning tslib has never been used in the build output.

#### Why tslib appeared in LICENSE.md
@rollup/plugin-typescript unconditionally resolves tslib during initialization via its internal getTsLibPath() function, regardless of whether importHelpers is enabled. This causes rollup-plugin-license to detect tslib as a bundled dependency even though no tslib code is present in any of the dist output files.

tslib first appeared in LICENSE.md when PR #5420 was reverted in 10bc150 (#5736). Prior to PR #4355, @rollup/plugin-typescript was listed in the external configuration in rollup.config.ts, which prevented the plugin's internal tslib resolution from being visible to rollup-plugin-license. After PR #4355 removed the external list, the plugin's resolveId hook began registering tslib, making it detectable as a bundled dependency — even though it was never actually included in the output.

Additionally, even after removing tslib from devDependencies, it remains in node_modules as a transitive dependency (via @tybys/wasm-util), so @rollup/plugin-typescript continues to resolve it successfully. This is why the LICENSE.md is not automatically corrected by simply removing the devDependencies entry.

 #### Verification
- Confirmed that no tslib code exists in any .js files under dist/ or dist/es/ (only a source path reference remains in dist/bin/rollup.map)
- Confirmed that importHelpers is not set in tsconfig.json

####  Changes
- package.json: Remove tslib from devDependencies
- build-plugins/generate-license-file.ts: Exclude tslib from the license file generation
- LICENSE.md: Regenerated — tslib section and 0BSD license type removed

### DIFF
--- a/build-plugins/generate-license-file.ts
+++ b/build-plugins/generate-license-file.ts
@@ -10,7 +10,7 @@ async function generateLicenseFile(
 	const coreLicense = await readFile('LICENSE-CORE.md', 'utf8');
 	const licenses = new Set<string>();
 	const dependencyLicenseTexts = [...dependencies]
-		.filter(({ name }) => name !== '@rollup/browser')
+		.filter(({ name }) => name !== '@rollup/browser' && name !== 'tslib')
 		.sort(({ name: nameA }, { name: nameB }) => (nameA! > nameB! ? 1 : -1))
 		.map(({ name, license, licenseText, author, maintainers, contributors, repository }) => {
 			let text = `## ${name}\n`;
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -13,7 +13,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 # Licenses of bundled dependencies
 The published Rollup artifact additionally contains code with the following licenses:
-MIT, ISC, 0BSD
+MIT, ISC
 
 # Bundled dependencies:
 ## @jridgewell/sourcemap-codec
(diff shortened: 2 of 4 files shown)


build-plugins/generate-license-file.ts
```diff
@@ -10,7 +10,7 @@ async function generateLicenseFile(
 	const coreLicense = await readFile('LICENSE-CORE.md', 'utf8');
 	const licenses = new Set<string>();
 	const dependencyLicenseTexts = [...dependencies]
-		.filter(({ name }) => name !== '@rollup/browser')
+		.filter(({ name }) => name !== '@rollup/browser' && name !== 'tslib')
```
lukastaegert, line 13: Thanks for the detailed analysis and fix. While I like the change in general, I do not feel good about this particular change. After all, we could start relying on the transitive tslib in the future, but this would prevent the license from showing up. If we really do not need it but `@rollup/plugin-typescript` always adds it, we should rather prevent it from being added to the build in the first place via a rollup plugin.

K-tecchan: Thank you for the feedback! Just to make sure I understand correctly — instead of filtering tslib in generate-license-file.ts, would you prefer adding a small Rollup plugin in rollup.config.ts that prevents @rollup/plugin-typescript from resolving tslib in the first place? 

Something like:
```ts
{
  name: 'ignore-tslib',                                                                                                                                   
  resolveId(id) {
    if (id === 'tslib') {
      return false;
    }
  }
}
```

TrickyPi:
Actually, the bundled Rollup code includes code from tslib. I found one example of this:
the original code is here: https://github.com/rollup/rollup/blob/a6be82b8abd747458afdc7484319f7b5deb92535/cli/run/build.ts#L33
and the transformed code is here (in dist/bin/rollup): 
Screenshot 2026-04-22 at 3 27 55 pm
where `__addDisposableResource` is a function from tslib

K-tecchan:
> Actually, the bundled Rollup code includes code from tslib. I found one example of this: the original code is here:
> 
> https://github.com/rollup/rollup/blob/a6be82b8abd747458afdc7484319f7b5deb92535/cli/run/build.ts#L33
> 
> 
> and the transformed code is here (in dist/bin/rollup):
> Screenshot 2026-04-22 at 3 27 55 pm
> where `__addDisposableResource` is a function from tslib

Sorry, I didn't realize this function was coming from tslib.
Since tslib ends up in the bundle, I think the right move is probably to keep the tslib entry in LICENSE.md and just drop it from devDependencies. Would that work, or would you prefer I close this PR instead?

TrickyPi:
Hmm, I think we should keep `tslib` in `devDependencies`, since it is a peer dependency of `@rollup/plugin-typescript`. This avoids relying on `tslib` being installed indirectly through other packages.

K-tecchan:
@TrickyPi @lukastaegert 
> Hmm, I think we should keep `tslib` in `devDependencies`, since it is a peer dependency of `@rollup/plugin-typescript`. This avoids relying on `tslib` being installed indirectly through other packages.

Got it! That makes sense. I'll go ahead and close this PR. 
Thank you for taking the time to review it.

PR #6390, opened 2026-05-22
Preserve emitted asset file modes

### LINKED ISSUE #6217: Preserve file permission on asset emit

Emitted assets should work the same as pre-bundled (be executable for instance).

File content is read as a "string, Uint8Array or Buffer" then written to a destination regardless of initial permissions.

A concrete example of this is the popular package "trash":
- it contains a dynamic url to a MacOS executable
- it can be bundled as asset via `import { importMetaAssets } from "@web/rollup-plugin-import-meta-assets";` for instance
- the emitted file loses executable permissions

### Modification Proposal

Add mode to options in `emitAsset()`:

```js
const stats = await stat(originalFileName);

const consumedAsset: ConsumedAsset = {
  // ...
  source, 
  type: 'asset',
  mode: stats.mode
}
```
https://github.com/rollup/rollup/blob/c5f3e1d3162ccb36e18704138c21457a084ef358/src/utils/FileEmitter.ts#L377-L385

Chmod after write file in `writeOutputFile()`:

```js
if (outputFile.type === 'asset') {
  await chmod(fileName, outputFile.mode);
}
```

https://github.com/rollup/rollup/blob/c5f3e1d3162ccb36e18704138c21457a084ef358/src/rollup/rollup.ts#L312-L323

puneetdixit200:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- resolves #6217

### Description

When an emitted asset is associated with an `originalFileName`, `bundle.write()` now looks up that original file mode and uses it when writing the asset. If the configured fs exposes `chmod`, Rollup also applies the mode after writing so existing output files are corrected as well.

If `originalFileName` is only metadata or the configured fs cannot stat it, Rollup keeps the previous write behavior.

Validation:

- `./node_modules/.bin/mocha test/misc/fs-override.js`
- `./node_modules/.bin/eslint src/rollup/rollup.ts src/rollup/types.d.ts test/misc/fs-override.js`
- `./node_modules/.bin/tsc --noEmit -p . --skipLibCheck`
- `env -u NO_COLOR FORCE_COLOR=3 npm run test:quick`

### DIFF
--- a/src/rollup/rollup.ts
+++ b/src/rollup/rollup.ts
@@ -312,14 +312,51 @@ function getSortingFileType(file: OutputAsset | OutputChunk): SortingFileType {
 async function writeOutputFile(
 	outputFile: OutputAsset | OutputChunk,
 	outputOptions: NormalizedOutputOptions,
-	{ fs: { mkdir, writeFile } }: NormalizedInputOptions
+	{ fs: { chmod, mkdir, stat, writeFile } }: NormalizedInputOptions
 ): Promise<unknown> {
 	const fileName = resolve(outputOptions.dir || dirname(outputOptions.file!), outputFile.fileName);
 
 	// 'recursive: true' does not throw if the folder structure, or parts of it, already exist
 	await mkdir(dirname(fileName), { recursive: true });
 
-	return writeFile(fileName, outputFile.type === 'asset' ? outputFile.source : outputFile.code);
+	if (outputFile.type === 'asset') {
+		const mode = await getOriginalAssetMode(outputFile, stat);
+		await writeFile(fileName, outputFile.source);
+		if (mode !== undefined) {
+			await chmod?.(fileName, mode);
+		}
+		return;
+	}
+
+	return writeFile(fileName, outputFile.code);
+}
+
+async function getOriginalAssetMode(
+	outputFile: OutputAsset,
+	stat: NormalizedInputOptions['fs']['stat']
+): Promise<number | undefined> {
+	const [originalFileName] = outputFile.originalFileNames;
+	if (!originalFileName) {
+		return;
+	}
+	try {
+		const { mode } = (await stat(originalFileName)) as { mode?: number };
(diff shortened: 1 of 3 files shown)


src/rollup/rollup.ts
```diff
@@ -312,14 +312,52 @@ function getSortingFileType(file: OutputAsset | OutputChunk): SortingFileType {
...
+	{ fs: { chmod, mkdir, stat, writeFile } }: NormalizedInputOptions
 ): Promise<unknown> {
 	const fileName = resolve(outputOptions.dir || dirname(outputOptions.file!), outputFile.fileName);
 
 	// 'recursive: true' does not throw if the folder structure, or parts of it, already exist
 	await mkdir(dirname(fileName), { recursive: true });
 
-	return writeFile(fileName, outputFile.type === 'asset' ? outputFile.source : outputFile.code);
+	if (outputFile.type === 'asset') {
+		const mode = await getOriginalAssetMode(outputFile, stat);
```
lukastaegert, line 323: As we are only doing anything if "chmod" is available, getting the original asset mode should be guarded by a check for chmod, otherwise it is completely unnecessary.

lukastaegert (commented):
Interesting idea, just one minor comment

lukastaegert (changes requested):
Thinking about this longer, though, I wonder if it would not make more sense to add the file mode as an additional attribute when emitting assets. That would be much less surprising, though it would be a bigger change as one would need to carry the attribute through the logic and also add it to emitted assets. If the attribute is missing, we carry it over as "null".
Initially, emitting with a file mode would file if chmod is missing. In Rollup 5, we could make "chmod" part of the "fs" type and always permit it.
That would put the job of running `fs.stat` on the emitter, but would could add an example to the docs that showcases this.

I think I would like this approach better, much fewer surprises. And I think most of the time you actually do not need the `fs.stat` as you likely already know the mode you want to set.

PR #6393, opened 2026-05-27
Use import attributes for importing JSON

selfisekai:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

### Description

Use `import with { type: "json" }` for importing JSON files. This is not required for rollup to be built by rollup, but allows to also build rollup with esbuild instead.

### DIFF
--- a/src/browser-entry.ts
+++ b/src/browser-entry.ts
@@ -1,4 +1,4 @@
-import '../typings/package.json';
+import package_ from '../package.json' with { type: 'json' };
 
-export { version as VERSION } from '../package.json' with { type: 'json' };
+export const VERSION = package_.version;
 export { defineConfig, default as rollup } from './rollup/rollup';
--- a/src/browser-entry.ts
+++ b/src/browser-entry.ts
@@ -1,4 +1,4 @@
 import '../typings/package.json';
 
-export { version as VERSION } from 'package.json';
+export { version as VERSION } from '../package.json' with { type: 'json' };
 export { defineConfig, default as rollup } from './rollup/rollup';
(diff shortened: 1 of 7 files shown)


src/browser-entry.ts
```diff
@@ -1,4 +1,4 @@
 import '../typings/package.json';
 
-export { version as VERSION } from 'package.json';
+export { version as VERSION } from '../package.json' with { type: 'json' };
```
lukastaegert: As I understand it, this would require support for named exports, which is not part of the official standard for `type: 'json'`. I think you would need to import and then do something like

```js
import package_ from '../package.json' with { type: 'json' };
export const VERSION = package_.version;
```

src/node-entry.ts
```diff
@@ -1,3 +1,3 @@
-export { version as VERSION } from 'package.json';
+export { version as VERSION } from '../package.json' with { type: 'json' };
```
lukastaegert: Another named re-export.

lukastaegert (commented):
Initially, I did not implement this as this would have added the entire package.json to the bundle. But as we now have object property tree-shaking, this is no longer an issue, I double-checked that this approach would just include an object `{ version: "<version>" }`, which is fine for me.
You should probably address my comment, though, if you want full support for other build tools. Admittedly at some point, we might also add proper JSON support to rollup.

lukastaegert (approved):
I took the liberty of adding my changes myself as this should not break anything and there is no need to leave it lying around.

PR #6398, opened 2026-05-30
feat: add native FreeBSD/powerpc64le binding

pkubaj:
Register the freebsd-ppc64 binding (powerpc64le-unknown-freebsd) so
Rollup runs natively on FreeBSD/powerpc64le instead of failing over to
the WASM build. The fallback is not usable here because a
rollup.config.js that imports from "rollup" loads native.js, which
currently throws for this platform/arch.

Currently no CI due to that platform being Tier 3 in Rust.
Tested on FreeBSD 15.1-BETA1 with Rust 1.95.0.

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no - requires a native builder, but it will likely work if other freebsd targets and other ppc64 targets work

Breaking Changes?

- [x] no


lukastaegert (commented):
Thanks for taking the initiative. However without CI changes, no artifact will be built, making this not so useful.

lukastaegert:
If it helps, we could move to a newer nightly build, provided it supports all necessary platforms and features.

pkubaj:
OK, I need some explanation. There is freebsd-arm64 support upstreamed here to rollup and artifacts are available, even though aarch64-freebsd is Tier 3 in Rust, so there's no prebuilt binaries. Why is that? What are the next steps for me to push? Should it be getting powerpc64le-freebsd to Tier 2 in Rust or something else?

lukastaegert:
To be honest I do not really care about tiers. If you manage to set up a running build pipeline in CI for that target, then we can ship that. That is the part you did not do. For that, you need to extend build-and-tests.yml, like in this commit https://github.com/rollup/rollup/commit/72b0f0ee02f6766f8e59c1724700da6647c6b51f

pkubaj:
CI added.

lukastaegert (approved):

lukastaegert:
Please check the failing build. Getting things to actually build is usually the hardest part.

pkubaj:
I've just sent an upstream cross PR for adding powerpc64le-unknown-freebsd: https://github.com/cross-rs/cross/pull/1781

PR #6403, opened 2026-06-02
Does not add an extra leading line feed for addons

### LINKED ISSUE #6401: always one LF above plugin added banner
### Rollup Version

4.61.0

node and deno and browser

v26.2.0

### Link To Reproduction

```js
import * as rollup from 'rollup'

export const modules = {
  "main.js": "import foo from 'foo.js'; console.log(foo);",
  "foo.js": "export default 42;",
};

export const out = await rollup
  .rollup({
    input: "main.js",
    plugins: [
      {
        name: "loader",
        resolveId(source) {
          if (modules.hasOwnProperty(source)) {
            return source;
          }
        },
        load(id) {
          if (modules.hasOwnProperty(id)) {
            return modules[id];
          }
        },
        banner(chunk) {
          // deal with chunk
          return "BANNER";
        },
        footer(chunk) {
          // deal with chunk
          return "FOOTER";
        }
      },
    ],
  })
  .then((bundle) => bundle.generate({ format: "iife" }))
  .then(({ output }) => (output[0].code));
```

### Expected Behaviour

```js
> console.log(out.startsWith("BANNER"))
true
```

### Actual Behaviour

```js
> console.log(out.startsWith("\nBANNER"))
true
```

TrickyPi:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

resolves #6401

### DIFF
--- a/src/utils/addons.ts
+++ b/src/utils/addons.ts
@@ -9,9 +9,8 @@ export interface Addons {
 	outro: string;
 }
 
-const concatSeparator = (out: string, next: string) => (next ? `${out}${out && '\n'}${next}` : out);
-const concatDblSeparator = (out: string, next: string) =>
-	next ? `${out}${out && '\n\n'}${next}` : out;
+const concatSeparator = (out: string, next: string) => [out, next].filter(Boolean).join('\n');
+const concatDblSeparator = (out: string, next: string) => [out, next].filter(Boolean).join('\n\n');
 
 export async function createAddons(
 	options: NormalizedOutputOptions,
--- a/src/utils/addons.ts
+++ b/src/utils/addons.ts
@@ -9,8 +9,9 @@ export interface Addons {
 	outro: string;
 }
 
-const concatSeparator = (out: string, next: string) => (next ? `${out}\n${next}` : out);
-const concatDblSeparator = (out: string, next: string) => (next ? `${out}\n\n${next}` : out);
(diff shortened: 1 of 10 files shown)


src/utils/addons.ts
```diff
@@ -9,8 +9,9 @@ export interface Addons {
 	outro: string;
 }
 
-const concatSeparator = (out: string, next: string) => (next ? `${out}\n${next}` : out);
-const concatDblSeparator = (out: string, next: string) => (next ? `${out}\n\n${next}` : out);
+const concatSeparator = (out: string, next: string) => (next ? `${out}${out && '\n'}${next}` : out);
+const concatDblSeparator = (out: string, next: string) =>
+	next ? `${out}${out && '\n\n'}${next}` : out;
```
lukastaegert: Maybe `join` would make the syntax slightly nicer? E.g.


```suggestion
const concatSeparator = (out: string, next: string) => [out, next].filter(Boolean).join('\n');
const concatDblSeparator = (out: string, next: string) => [out, next].filter(Boolean).join('\n\n');
```

TrickyPi: Yeah, makes sense!

lukastaegert (approved):

PR #6404, opened 2026-06-03
fix: set report.excludeNetwork=true before getReport() to avoid blocking PTR lookups

jdz321:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

`native.js` calls `report.getReport()` during module initialization to detect the platform (`isMingw32`/`isMusl`). `report.getReport()` enumerates all libuv TCP handles and performs a reverse DNS (PTR) lookup on each one. When active TCP sockets are present and the remote IP's PTR query returns `SERVFAIL`, the call blocks until DNS times out.  

This affects projects where a plugin calls `fetch()` during a build, leaving a keep-alive socket in undici's pool, and a later plugin import triggers `native.js` to load.

Fix: set `report.excludeNetwork = true` before calling `report.getReport()` and restore the previous value afterwards. For the Windows `spawnSync` path, `excludeNetwork` is set inside the child script.

Ref: https://github.com/nodejs/node/issues/55576
Ref: https://github.com/nodejs/node/pull/55602

### DIFF
--- a/native.js
+++ b/native.js
@@ -6,12 +6,18 @@ const { spawnSync } = require('node:child_process');
 const getReportHeader = () => {
 	try {
 		if (platform !== 'win32') {
-			return report.getReport().header;
+			// Avoid blocking reverse DNS (PTR) lookups on open TCP socket handles.
+			// See: https://github.com/nodejs/node/issues/55576
+			const previousExcludeNetwork = report.excludeNetwork;
+			report.excludeNetwork = true;
+			const header = report.getReport().header;
+			report.excludeNetwork = previousExcludeNetwork;
+			return header;
 		}
 
 		// This is needed because report.getReport() crashes the process on Windows sometimes.
 		const script =
-			"console.log(JSON.stringify(require('node:process').report.getReport().header));";
+			"const r=require('node:process').report;r.excludeNetwork=true;console.log(JSON.stringify(r.getReport().header));";
 		const child = spawnSync(process.execPath, ['-p', script], {
 			encoding: 'utf8',
 			timeout: 3000,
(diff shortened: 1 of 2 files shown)


lukastaegert (approved):
I was not aware, nice improvement!

jdz321:
Here's a fun one, let me share how we stumbled into this bug.

While building a project with Vite 8 + vite-plugin-pwa, the Vite build itself completed quickly, but the PWA build step took an extra ~30 seconds to finish. After investigation, the root cause turned out to be a `fetch()` call in another Vite plugin that requests an internal IP address. Removing that plugin or changing the URL to a public address made the delay disappear.

The reason this only appears with Vite 8 is subtle: Vite 7 and earlier use rollup as the build engine, so `native.js` is loaded (and `report.getReport()` is executed) before any plugin `fetch()` calls happen. With Vite 8, the build engine switched to rolldown, but vite-plugin-pwa still depends on rollup internally. This means `native.js` is now loaded lazily — inside a `closeBundle` hook — which happens *after* the `fetch()` call has already left a keep-alive socket in undici's connection pool. At that point `report.getReport()` enumerates the live socket handles and performs a reverse DNS (PTR) lookup on the remote IP. Since the internal IP's PTR query returns `SERVFAIL` instead of `NXDOMAIN`, the DNS resolver retries until it times out, blocking the process for ~30 seconds.

jdz321:
the test now skips on Windows since `getReport()` runs inside a `spawnSync` child process there, making it impossible to intercept via `process.report.getReport` mock in the current process. The Windows fix (setting `excludeNetwork` in the child script) is verified by the script string itself.

lukastaegert (approved):

PR #6408, opened 2026-06-04
fix: insert conflict numbers before first extension in multi-extension filenames

### LINKED ISSUE #5822: When resolving file name conflicts, add the number used to prevent a conflict before the first dot of the filename instead of the last

When Rollup resolves file name conflicts, it adds a number to the conflicting file before the last dot of the file name. _I.e._ if it is creating a file named `foo.bar.js` but a file already exists with said name, then it names the new file `foo.bar2.js`.

This is fine in most scenarios, however, in some cases, a composition of multiple "extensions" in a file name is actually relevant meta-information for the type of file.

In particular, I'm using `rollup-plugin-dts` to generate type definitions for my library, which has multiple entry points. I have at some point in my Rollup configuration something like:

```js
// ...
{
  input: arrayOfEntryPoints,
  output: [{ dir: "dist", chunkFileNames: "chunk.d.ts" }],
  plugins: [dts()],
},
// ...
``` 

When generating chunks, I'd expect Rollup to emit files named: `chunk.d.ts`, `chunk2.d.ts`, `chunk3.d.ts`, ...

However, as previously mentioned, Rollup currently adds the number used to prevent file conflicts just before the last dot of the file, so `chunk.d.ts`, `chunk.d2.ts`, `chunk.d3.ts`, ... is generated instead.

This is problematic for this particular use-case since TypeScript type definitions are expected to have a file extension of `.d.ts`. 

### Modification Proposal

I propose that Rollup instead adds conflict-preventing numbers before the first dot of the file name, rather than the last.

I believe that such behaviour is the least likely to cause problems related to file extensions. I can think of other scenarios where the current behaviour might cause problems (_e.g._ think of a plugin which generates `tar.gz` archives), whereas I can't think of any scenarios where this proposal would cause issues (other than breaking compatibility with the previous behaviour).

LeSingh1:
Closes #5822.

When rollup encounters a filename conflict it calls `makeUnique`, which
appended a counter just before the last file extension. For a
`chunkFileNames` pattern like `chunk.d.ts` that produced `chunk.d2.ts`
and `chunk.d3.ts` on subsequent conflicts instead of `chunk2.d.ts` and
`chunk3.d.ts`.

The problem is that `path.extname` only returns the last segment
(`.ts`), so the base after stripping becomes `chunk.d` and the number
lands in the wrong spot. TypeScript declaration files emitted by
rollup-plugin-dts are the most common case where this breaks things
because TypeScript's type resolver requires the `.d.ts` extension as a
whole.

The fix replaces the `extname` call with a small helper that scans for
the first dot in the filename (skipping a leading dot for hidden files
like `.env.local`) and treats everything from that position onwards as
the extension to preserve. Single-extension names are unaffected because
`chunk.js` has its first dot at the same position as its last dot.

A new chunking-form test verifies `chunk.d.ts` resolves conflicts as
`chunk2.d.ts` and `chunk3.d.ts` across all four output formats (es,
cjs, amd, system).

### DIFF
--- a/src/utils/renderNamePattern.ts
+++ b/src/utils/renderNamePattern.ts
@@ -1,7 +1,6 @@
 import { error, logFailedValidation } from './logs';
 import type { OutputBundleWithPlaceholders } from './outputBundle';
 import { lowercaseBundleKeys } from './outputBundle';
-import { extname } from './path';
 import { isPathFragment } from './relativeId';
 
 export function renderNamePattern(
@@ -42,7 +41,7 @@ export function makeUnique(
 	{ [lowercaseBundleKeys]: reservedLowercaseBundleKeys }: OutputBundleWithPlaceholders
 ): string {
 	if (!reservedLowercaseBundleKeys.has(name.toLowerCase())) return name;
-	const extension = extname(name);
+	const extension = getFirstExtension(name);
 	name = name.slice(0, Math.max(0, name.length - extension.length));
 	let uniqueName: string,
 		uniqueIndex = 1;
@@ -50,4 +49,13 @@ export function makeUnique(
 		reservedLowercaseBundleKeys.has((uniqueName = name + ++uniqueIndex + extension).toLowerCase())
 	);
 	return uniqueName;
+}
+
+// Returns everything from the first dot onwards, treating files that start with
+// a dot (e.g. ".gitignore") as having no leading dot for extension purposes so
+// that ".env.local" yields ".local" rather than the full name.
+function getFirstExtension(name: string): string {
+	const startIndex = name.startsWith('.') ? 1 : 0;
+	const dotIndex = name.indexOf('.', startIndex);
+	return dotIndex === -1 ? '' : name.slice(dotIndex);
 }
--- a/src/utils/renderNamePattern.ts
+++ b/src/utils/renderNamePattern.ts
@@ -41,21 +41,13 @@ export function makeUnique(
 	{ [lowercaseBundleKeys]: reservedLowercaseBundleKeys }: OutputBundleWithPlaceholders
 ): string {
 	if (!reservedLowercaseBundleKeys.has(name.toLowerCase())) return name;
-	const extension = getFirstExtension(name);
-	name = name.slice(0, Math.max(0, name.length - extension.length));
+	const dotIndex = name.indexOf('.', 1);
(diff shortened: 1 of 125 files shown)


src/utils/renderNamePattern.ts
```diff
@@ -51,3 +50,12 @@ export function makeUnique(
 	);
 	return uniqueName;
 }
+
+// Returns everything from the first dot onwards, treating files that start with
+// a dot (e.g. ".gitignore") as having no leading dot for extension purposes so
+// that ".env.local" yields ".local" rather than the full name.
+function getFirstExtension(name: string): string {
+	const startIndex = name.startsWith('.') ? 1 : 0;
```
lukastaegert: I think this check is unnecessary, why not just always start with `1`?

src/utils/renderNamePattern.ts
```diff
@@ -42,7 +41,7 @@ export function makeUnique(
 	{ [lowercaseBundleKeys]: reservedLowercaseBundleKeys }: OutputBundleWithPlaceholders
 ): string {
 	if (!reservedLowercaseBundleKeys.has(name.toLowerCase())) return name;
-	const extension = extname(name);
+	const extension = getFirstExtension(name);
 	name = name.slice(0, Math.max(0, name.length - extension.length));
```
lukastaegert: As we are rolling our own logic here, why not inline the logic from `getFirstExtension` directly into `makeUnique`? Then we can directly work with `dotIndex` in the calculation here, and get rid of the `Math.max`.

lukastaegert (commented):
I like the idea, just some minor comments.

LeSingh1:
Thanks for the review. Applied both comments:

- Inlined the getFirstExtension logic directly into makeUnique and dropped the helper entirely. dotIndex is now computed inline with indexOf('.', 1), and base and extension are derived from it directly, so there's no longer a Math.max needed.
- Since the search always starts at index 1, the startsWith('.') conditional check was no longer needed and is gone.

Also fixed a prettier line-length violation in the test _config.js that was causing the Lint and Audit CI job to fail.

src/utils/renderNamePattern.ts
lukastaegert: Nice,thanks!

lukastaegert (approved):

lukastaegert (approved):

PR #6411, opened 2026-06-11
Skip side-effect-free external imports when hoisting is disabled

### LINKED ISSUE #6111: Unexpected hoisting when `output.hoistTransitiveImports = true` and `treeshake.moduleSideEffects = false | 'no-external'`
### Rollup Version

v4.50.2

macOS 15.5

_No response_

### Link To Reproduction

https://rollupjs.org/repl/?version=4.50.2&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiZXhwb3J0IHtjcmVhdGVDbGllbnQsIHVuc3RhYmxlX19hZGFwdGVyLCB1bnN0YWJsZV9fZW52aXJvbm1lbnR9IGZyb20gJ0BzYW5pdHkvY2xpZW50J1xuXG4vLyBTb21laG93IHVuY29tbWVudGluZyB0aGUgYmVsb3cgd2lsbCBzdG9wIHRoZSBob2lzdGluZ1xuLy8gZXhwb3J0IGNvbnN0IGZvbyA9ICdiYXInIiwiaXNFbnRyeSI6dHJ1ZSwibmFtZSI6Im1haW4uanMifSx7ImNvZGUiOiJleHBvcnQge3VzZVByZXNlbnRhdGlvblF1ZXJ5fSBmcm9tICdAc2FuaXR5L25leHQtbG9hZGVyL2hvb2tzJ1xuZXhwb3J0IHt1c2VPcHRpbWlzdGljfSBmcm9tICdAc2FuaXR5L3Zpc3VhbC1lZGl0aW5nL3JlYWN0J1xuXG4vLyBVbmNvbW1lbnRpbmcgdGhpcyBhbHNvIHN0b3BzIHRoZSBpbXBvcnQgaG9pc3Rpbmdcbi8vIGV4cG9ydCBjb25zdCBmb28gPSAnYmFyJyIsImlzRW50cnkiOnRydWUsIm5hbWUiOiJob29rcy5qcyJ9XSwib3B0aW9ucyI6eyJvdXRwdXQiOnsiZm9ybWF0IjoiZXMiLCJob2lzdFRyYW5zaXRpdmVJbXBvcnRzIjpmYWxzZX0sInByZXNlcnZlRW50cnlTaWduYXR1cmVzIjoic3RyaWN0IiwidHJlZXNoYWtlIjoic21hbGxlc3QifX0=

### Expected Behaviour

Since `output.hoistTransitiveImports` are `false`, and `preserveEntrySignatures` are `'strict'`, there should be no hoisting of imports.
Setting `treeshake.moduleSideEffects` to `false` or `'no-external'` should only treeshake, no matter if the entries are purely re-exports from other packages, or a mixture.

### Actual Behaviour

In the repro, the output `main.js` should not have the `import '@sanity/next-loader/hooks'` and `import '@sanity/visual-editing/react'` bare imports, they are only used by the `hooks.js` output.

What's even weirder is that it seems connected to chunks that only re-export from other chunks. If any other export is added, like `export const foo = 'bar'`, then the hoisting no longer happens.

morgan-coded:
When `hoistTransitiveImports: false` is used with `experimentalMinChunkSize`, an entry chunk can pick up bare external imports from merged pure re-export chunks even though the entry never references those packages.
This keeps binding-less, side-effect-free external chunks out of the rendered imports when hoisting is disabled, and keeps `chunk.imports` in sync with the emitted code. The default hoisting path is left unchanged.
I added chunking-form coverage for ES, CJS, AMD, and System output. The focused sample goes red on the issue shape and green with the fix; I also ran the full suite, eslint, `tsc --noEmit`, and `git diff --check`.
Closes #6111

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -1245,6 +1245,7 @@ export default class Chunk {
 		const importSpecifiers = this.getImportSpecifiers();
 		const reexportSpecifiers = this.getReexportSpecifiers();
 		const renderedDependencies = new Map<Chunk | ExternalChunk, ChunkDependency>();
+		const skippedDependencies: ExternalChunk[] = [];
 		const fileName = this.getFileName();
 		for (const dependency of this.dependencies) {
 			const imports = importSpecifiers.get(dependency) || null;
@@ -1254,14 +1255,14 @@ export default class Chunk {
 				reexports === null &&
 				dependency instanceof ExternalChunk &&
 				!dependency.moduleSideEffects &&
-				!this.outputOptions.hoistTransitiveImports
+				this.outputOptions.hoistTransitiveImports === false
 			) {
 				// A side-effect-free external dependency without imported or
 				// re-exported bindings can only be present because chunk assignment
 				// placed otherwise unrelated modules into this chunk. When transitive
 				// import hoisting is disabled, rendering it would emit a spurious
 				// side effect import, see https://github.com/rollup/rollup/issues/6111
-				this.dependencies.delete(dependency);
+				skippedDependencies.push(dependency);
 				continue;
 			}
 			const namedExportsMode =
@@ -1295,6 +1296,10 @@ export default class Chunk {
 				reexports,
 				sourcePhaseImport: sourcePhaseImport?.local
 			});
+		}
+
+		for (const dependency of skippedDependencies) {
+			this.dependencies.delete(dependency);
 		}
 
 		return (this.renderedDependencies = renderedDependencies);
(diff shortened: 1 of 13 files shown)


lukastaegert (approved):
Thanks for the PR! I generally agree with your assessment and think the general idea of the fix is fine. What I did not like is the implicit side effect of mutating dependencies in `this.getRenderedDependencies()`. I took the liberty to slightly change this so that dependencies are not mutated and instead relevant places check `this.getRenderedDependencies()` instead.

PR #6415, opened 2026-06-17
fix: advance value past wildcard prefix before suffix check in getLogFilter

JSap0914:
## Bug

The wildcard pattern matching in `testFilter` (used by `getLogFilter`) checked whether the value started with `parts[0]` (the prefix) but did **not** advance the search position past it before checking the suffix. This caused false positive matches whenever the prefix and suffix were the same string and the value was exactly equal to one copy of that string.

**Example:**
```js
const filter = getLogFilter(['code:foo*foo']);
filter({ code: 'foo' }); // was: true — WRONG, should be false
```

`'foo'.startsWith('foo')` is true and `'foo'.endsWith('foo')` is true, so the old code matched, even though the wildcard `*` would need to cover -3 characters to make the match valid. The minimum valid match for `foo*foo` is `foofoo` (where `*` matches the empty string).

The same issue applies to any non-empty prefix and suffix that share content, e.g. `WARN*WARN`, `A*A`, etc.

## Fix

After the prefix check succeeds, slice `value` past the prefix so the subsequent suffix search only operates on the remaining substring:

```ts
if (!value.startsWith(parts[0])) {
    return false;
}
value = value.slice(parts[0].length); // advance past prefix
```

## Verification

```
npx mocha test/misc/get-log-filter.js
```

All 10 tests pass (9 existing + 1 new regression test):

```
  getLogFilter
    ✔ does not match when wildcard prefix and suffix overlap
    ✔ does not filter when there are no filters
    ✔ filters for string matches
    ✔ combines multiple filters with or
    ✔ supports placeholders
    ✔ supports inverted filters
    ✔ supports AND conditions
    ✔ handles numbers and objects
    ✔ handles edge case filters
    ✔ handles nested properties

  10 passing (20ms)
```

> AI-assisted contribution.


lukastaegert (approved):
Nicely spotted, looks good!

PR #6421, opened 2026-06-23
docs: update x_google_ignoreList link to canonical URL

DucMinhNe:
The `output.sourcemapIgnoreList` docs link to the `x_google_ignoreList` source map extension via:

```
https://developer.chrome.com/articles/x-google-ignore-list/
```

That URL now 301-redirects to the canonical page:

```
https://developer.chrome.com/docs/devtools/x-google-ignore-list
```

Updated the href to point straight at the canonical page so it doesn't bounce through a redirect.


lukastaegert (approved):
Nice catch, thanks!

PR #6432, opened 2026-07-02
fix: make isLegal idempotent by not using a global-flag regex

spokodev:
`isLegal` checks an identifier with `illegalCharacters.test(value)`, where
`illegalCharacters` is a module-level regex that carries the global flag
(`/[^\w$]/g`). `RegExp.prototype.test` on a global regex advances and remembers
`lastIndex`, and because the regex is a shared singleton that state leaks across
calls. `isLegal` is therefore not idempotent: the same input returns different
answers depending on what was tested before it.

For example, `isLegal('foo.bar')` returns `false` on the first call and `true`
on the next. More seriously, after a legal name is checked, the following check
of an illegal name can wrongly return `true`, so an invalid `output.name` for
`iife` or `umd` output (for example `my=name` or `a-b`) can be accepted instead
of rejected, which emits invalid JavaScript.

The fix uses a separate non-global regex for the membership test in `isLegal`.
The global regex is kept for `makeLegal`, whose `String.prototype.replace`
resets `lastIndex` and is unaffected.


lukastaegert (approved):
I just had a good agent try really hard to find a red test for this, but in the end it found no way to generated an error from this issue, so I agree with it that it is likely only a theoretical bug. But of course the change is correct, and it does not make sense to use a global regex here.

PR #6442, opened 2026-07-09
fix: expose cached module meta during cache checks

### LINKED ISSUE #6441: `this.getModuleInfo(moduleId).meta` is an empty object when `this.load` is called in `shouldTransformCachedModule`
### Rollup Version

4.62.2

StackBlitz

_No response_

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-6edwwvwm?file=package.json,rollup.config.js,src%2Fcounter.js

### Expected Behaviour

When editing `counter.js` while `npm run build:watch` is running, the value set by the `setModuleInfoPlugin` plugin exists in `info?.meta?.foo` and the rebuild does not fail.

### Actual Behaviour

When editing `counter.js` while `npm run build:watch` is running, the value set by the `setModuleInfoPlugin` plugin is missing in `info?.meta?.foo` and the rebuild fails.

### Additional Information

This is a reduced reproduction of https://github.com/vitejs/vite/issues/20591

ychampion:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- resolves #6441

### Description

`shouldTransformCachedModule` runs before Rollup restores cached transform metadata onto the new module info object. If a plugin calls `this.load({ id })` from that hook and then reads `this.getModuleInfo(id).meta`, it sees `{}` even though the hook argument contains the cached metadata.

This restores the cached module options before running the hook, resets back to the current module options when the hook asks to re-transform, and adds an incremental regression test for `this.load` inside the hook.

Validation:

- `npm run build:cjs`
- `npx mocha test/incremental/index.js --grep "makes cached module meta available"`
- `npx mocha test/incremental/index.js`
- `npx eslint src/Module.ts src/ModuleLoader.ts test/incremental/index.js --cache --concurrency auto`
- `env -u NO_COLOR FORCE_COLOR=1 TERM=xterm npm run test:quick`
- `git diff --check`


test/incremental/index.js
```diff
@@ -395,6 +395,39 @@ describe('incremental', () => {
...
+			foo: `export default 42`
+		};
+		let shouldTransformCachedModuleCalls = 0;
+
+		const metaPlugin = {
+			resolveId: id => id,
+			load: id => ({ code: modules[id] }),
+			shouldTransformCachedModule({ id }) {
+				shouldTransformCachedModuleCalls++;
+				this.load({ id });
```
lukastaegert, line 410: Why are we calling `this.load` in this hook? Without awaiting it, this should not have any effect, but if it were awaited, it looks to me like a guaranteed deadlock.

src/ModuleLoader.ts
```diff
@@ -314,31 +314,42 @@ export class ModuleLoader {
...
 		) {
-			if (cachedModule.transformFiles) {
-				for (const emittedFile of cachedModule.transformFiles)
-					this.pluginDriver.emitFile(emittedFile);
+			const originalModuleOptions = {
+				meta: { ...module.info.meta },
+				moduleSideEffects: module.info.moduleSideEffects,
+				syntheticNamedExports: module.info.syntheticNamedExports
+			};
+			module.updateOptions(cachedModule, { replaceExistingMeta: true });
```
lukastaegert, line 324: Here is one small problem, which is actually kind of a regression: Entry points are always freshly resolved on each run and can receive different meta on each run. If then for an entry point, we still use the cached resolution, we will lose the updated meta. Which is arguably wrong, especially in the `shouldTransformCachedModule` call. 

A simple fix is to _not_ use `replaceExistingMeta` on the first call. This will not change anything for modules where also the importer is cached as in that case, the importer's cached resolution is used, but it will have an effect on entry points.
I can push a red test to demonstrate this.

lukastaegert (commented):
Hi, this is a good catch. I have comments, but in general this is a good approach.

test/incremental/index.js
lukastaegert: I removed this now as it did not seem consequential for the reproduction.

src/ModuleLoader.ts
lukastaegert: Pushed and fixed

lukastaegert (approved):

PR #6452, opened 2026-07-19
feat: assignment literal analysis and truthy functions

cyyynthia:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

This PR addresses 2 shortcomings of Rollup currently:
- Assignment expressions are not walked through when it's a simple `=`. This unlocks analysis of e.g. `() => globalThis.value = true` as a side-effectful function returning `true`.
- Due to the TS enum bug, ObjectEntity is not considered truthy and so are functions and classes. I've made them act as truthy, which allowed Rollup to optimise core-js parts that the bundler considers always present (the optimisation gains are underwhelming due to the lack of advanced analysis of variable reassignment, which would allow Rollup to get rid of entire `requireXXX` blocks in core-js).

### DIFF
--- a/src/ast/nodes/AssignmentExpression.ts
+++ b/src/ast/nodes/AssignmentExpression.ts
@@ -14,18 +14,22 @@ import {
 	renderSystemExportFunction,
 	renderSystemExportSequenceAfterExpression
 } from '../../utils/systemJsRendering';
+import type { DeoptimizableEntity } from '../DeoptimizableEntity';
 import {
 	createHasEffectsContext,
 	type HasEffectsContext,
 	type InclusionContext
 } from '../ExecutionContext';
-import type { NodeInteraction } from '../NodeInteractions';
+import type { NodeInteraction, NodeInteractionCalled } from '../NodeInteractions';
 import { INTERACTION_ASSIGNED } from '../NodeInteractions';
+import type { EntityPathTracker } from '../utils/PathTracker';
 import { EMPTY_PATH, type ObjectPath, UNKNOWN_PATH } from '../utils/PathTracker';
 import type Variable from '../variables/Variable';
 import Identifier from './Identifier';
 import * as NodeType from './NodeType';
 import ObjectPattern from './ObjectPattern';
+import type { ExpressionEntity, LiteralValueOrUnknown } from './shared/Expression';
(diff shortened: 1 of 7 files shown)


lukastaegert:
Hmmm, isn't getting rid of polyfills in core-js actually a bad thing? After all, it is polyfilling missing prototype methods. From the diff it looks to me like your change might actually break the `copyWithin` polyfill. I believe it was for those reasons we never did the last step to mark some things as truthy (though of course negligence is likely the other reason).
At the very least what you must not do is add any logic that could give the impression that any of the assumed-present global/prototype methods actually exist, exactly not to break polyfills. It is fine to assume that objects or functions that the user declares are truthy, or methods in objects that the user declares.

cyyynthia:
> At the very least what you must not do is add any logic that could give the impression that any of the assumed-present global/prototype methods actually exist, exactly not to break polyfills. It is fine to assume that objects or functions that the user declares are truthy, or methods in objects that the user declares.

I think this is a fair point, though perhaps we can preserve this new behaviour as an opt-in config flag instead 🤔

lukastaegert (changes requested):
Hi! This looks good in general now, but I found a regression in the updates to `AssignmentExpression`. I pushed a failing test to demonstrate. Basically if we forward `AssignmentExpression.getLiteralValueAtPath` to `this.right`, then we also need to forward deoptimizations of the result to `this.right`. I managed to get this to produce bugs when using destructuring assignments. Similarly, we should also implement `deoptimizeArgumentsOnInteractionAtPath`.

lukastaegert:
> I think this is a fair point, though perhaps we can preserve this new behaviour as an opt-in config flag instead 🤔

I thought about this long in the past and I keep coming back to the fact that the only good reason why one would want to check the truthiness of a built-in prototype method is to do feature-detection, and we must not break those flows.
So there are several ways to do feature detection:
- Truthiness as a smoke test if there is _something_ at a given prototype value, this we should never break
- `typeof` checks. We are not doing anything here at the moment, but in the future, we should not blindly assume that a prototype function returns `function` for a `typeof` call
- just calling it in a try-catch. While our logic assumes many builtins have no side effects, the try-catch-deoptimization will still preserve the call in those situations

PR #6462, opened 2026-07-29
Add GitHub Actions workflow for npm package publishing

garciaerubiel745-create:
This workflow runs tests and publishes a Node.js package to GitHub Packages upon release creation.

This PR contains:

- [x] bugfix
- [x] feature
- [ ] refactor
- [x] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [ ] no


lukastaegert:
I do not know what this is supposed to be, we do have a workflow for publishing

lukastaegert:
Ah, this is a different registry. Still, this is not how I would do it, also I am not sure why we would need this. There is no motivation in you PR description, it is not clear to me why you would NOT want to publish any of the arch-dependent packages, and why this should not be directly part of the actual release process.

PR #6463, opened 2026-07-30
docs: add llms.txt documentation index for LLMs and agents

abyworkings-coder:
Adds a `llms.txt` file (served at `rollupjs.org/llms.txt`) at `docs/public/`, following the same VitePress static-passthrough convention already used for `favicon.png` and `manifest.json`.

`llms.txt` is a proposed convention ([llmstxt.org](https://llmstxt.org/)) for giving LLM tools and coding assistants a curated, low-noise index of documentation, in the same spirit as `robots.txt` or `sitemap.xml`. It's a static file that doesn't change anything about the existing site — it only adds an index that AI tools can consume directly instead of crawling/guessing through the full site.

This PR includes a curated index covering Getting Started, Core Concepts, Guides, and an Optional section, linking to the existing docs pages (introduction, tutorial, CLI, JS API, configuration options, plugin development, ES module syntax, browser usage, migration guide, troubleshooting, FAQs, other tools, REPL). All linked URLs verified to return 200.

Happy to adjust scope/structure if maintainers want a different curation.


lukastaegert (approved):
Thanks, looks good to me!

PR #6468, opened 2026-08-06
Keep the semicolon added after a replaced default export

Jaybhade:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

A semicolon-free `export default -1` produces a bundle that does not parse when `output.compact` is enabled:

```js
// main.js — note there is no trailing semicolon
export default -1
```

```js
// rollup --format cjs --compact
'use strict';var main = -1module.exports=main;
//                        ^^^^^^^^^^^ SyntaxError: Unexpected identifier 'module'
```

`export default this` fails the same way, just at runtime instead of parse time:

```js
'use strict';var main = undefinedmodule.exports=main;
// ReferenceError: undefinedmodule is not defined
```

All output formats are affected — for `es` the output is `var main = -1export{main as default};`.

#### Cause

`ExportDefaultDeclaration#renderVariableDeclaration` rewrites the statement header to `var main = ` and, when the source has no trailing semicolon, appends one at the end of the statement:

```js
code.overwrite(this.start, declarationStart, `${cnst} ${name} = `);
if (!hasTrailingSemicolon) {
    code.appendLeft(this.end, ';');
}
```

That append runs *before* `this.declaration.render(...)`. Some expressions replace themselves wholesale while rendering — `ThisExpression` overwrites its range with the module context, and `UnaryExpression` overwrites its range with a folded literal — and `overwrite` discards content appended to the left of the range's end. So whenever the replaced expression is the last thing in the statement, the semicolon that was just appended is thrown away again.

Nothing noticed until now because the finalisers put a line break after the statement, so ASI kept the output valid. `output.compact` removes that line break and the tokens run together.

The same ordering problem was already solved twice in this file: the `else` branch at the end of `render` appends its semicolon *after* rendering the declaration, and the SystemJS branch uses `appendRight` (#2529). This PR fixes the remaining branch the same way as the first: `renderVariableDeclaration` now reports whether a semicolon is still needed, and `render` appends it once the declaration has been rendered.

I kept `appendLeft` rather than switching to `appendRight` on purpose: `appendRight(this.end, ';')` also fixes the output, but it attributes the semicolon to the chunk after the statement, which shrinks the module's reported `renderedLength` by one. Appending after the render keeps that metadata unchanged.

#### Tests

Two `function` tests, since the point is that the generated code has to run:

- `export-default-folded-expression-without-semicolon` — `export default typeof 1`, which folds to a literal. Fails with `SyntaxError: Unexpected identifier 'module'` before the fix.
- `export-default-this-without-semicolon` — `export default this`. Fails with `ReferenceError: undefinedmodule is not defined` before the fix.

Both use `output.compact: true`, which is what turns the missing semicolon into a hard failure.

`npm run test:cjs` is green (5438 passing), and no existing snapshot changed.


lukastaegert (approved):
Thanks a lot, looks good to me!

PR #6473, opened 2026-08-08
fix(browser): match node's basename and extname

luantaraschi:
The browser build swaps `node:path` for `browser/src/path.ts`, so the two builds should answer the same question the same way. For a dotfile they do not, and for a path ending in a slash they do not either.

```js
extname(".htaccess")    // node:path ""    browser shim ".htaccess"
extname(".env")         // node:path ""    browser shim ".env"
extname("/a/.npmrc")    // node:path ""    browser shim ".npmrc"
extname("file.")        // node:path "."   browser shim ""
extname("file./")       // node:path "."   browser shim ""
basename("/foo/bar/")   // node:path "bar" browser shim ""
```

Two separate causes.

`/\.[^.]+$/` matched the leading dot of a dotfile as if it opened an extension, and it refused a trailing dot with nothing after it.

`basename` split on every slash and took the last piece, so a path ending in one produced `""`. Since `extname` reads the extension off `basename`, it inherited that whole class: it was handed an empty base name and had nothing to work with. The dotfile fix alone did not reach those.

This is visible through `assetFileNames`. `FileEmitter` derives all three placeholders from `extname`:

```ts
ext: () => extname(emittedName).slice(1),
extname: () => extname(emittedName),
name: () => emittedName.slice(0, Math.max(0, emittedName.length - extname(emittedName).length))
```

So an emitted asset called `.htaccess`, with the default `assets/[name]-[hash][extname]`, lands as `assets/.htaccess-<hash>` under Node and `assets/-<hash>.htaccess` in the browser. `relativeId`, `renderChunks` and `Module.basename` read the same two functions.

## The change

`extname` takes the last dot in the base name, and only when it is not the first character. `.` and `..` name a directory, so they get no extension.

`basename` drops trailing slashes before splitting, because `a/b/` names the same entry as `a/b`.

Both are node's rules restated, in the file whose whole job is to restate them.

## How it was checked

I could not run the browser suite here: it needs `npm run build:wasm`, and this machine has no Rust toolchain, so I am not claiming I ran it. I loaded `browser/src/path.ts` directly instead and compared it against `path.posix` over every string of length 0 to 6 built from `/`, `.`, `a` and `b`. 5460 inputs.

| | before | after |
|---|---|---|
| `basename` | 1359 | 0 |
| `extname` | 563 | 0 |

Unchanged before and after: `index.js` → `.js`, `file.tar.gz` → `.gz`, `a.` → `.`, `..a` → `.a`, `basename("a/b")` → `b`, `basename("/")` → `""`.

`prettier --check` and `eslint` are clean on the file.

## What I left out

`dirname` parts ways on the same trailing slash, and the trim above fixes most of it: over the same 5460 inputs it goes from 1482 disagreements to 178. The 178 that remain are all the POSIX `//` root, where `dirname("//a")` is `"//"` and this shim says `"/"`. That is a different rule with its own reasoning, so I would rather send it separately than bury it here. Say the word and I will.


lbesecker195 (approved):
I ran the new `basename`/`extname` against `path.posix` on 16 inputs including `.htaccess`, `a/b/`, `a.`, `...`, `.a.b`, `..`, `.`, `foo.tar.gz`, ``, and `/`: zero mismatches. The old regex disagreed with Node on dotfiles and trailing slashes. LGTM.

lukastaegert (approved):
Thanks for spotting this issue. I took the liberty to add the change for dirname as well in your merge request and will merge this now.

PR #6121, opened 2025-09-21
Simplify top-level await deadlock prevention

lukastaegert:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes (_bugfixes and features will not be merged without tests_)
- [ ] no

Breaking Changes?

- [ ] yes (_breaking changes will not be merged unless absolutely necessary_)
- [x] no

List any relevant issue numbers:
- resolves #6093

### Description

This very much simplifies the logic to prevent merges with top-level-await modules that create deadlocks. The proposed logic is to just assume that a file that contains a top-level await will await all dynamic imports within the module. This will prevent that any dependencies of dynamically imported modules are merged into the same chunk as the current module.

cc @TrickyPi, what do you think?

PR #6481, opened 2026-08-17
fix: avoid spurious cycle warning for dynamic imports inside functions with TLA

HosnainRafi:
## Summary

When a module uses top-level await (TLA), rollup was treating ALL dynamic imports from that module as synchronous dependencies, even if the import was inside a function. This caused false circular dependency warnings and incorrect module ordering.

See #6154 for the full reproduction.

## Root Cause

In `src/utils/executionOrder.ts`, the `analyseModuleExecution` function checks `scope.context.usesTopLevelAwait` for all dynamic imports in a module. If the module uses TLA, every dynamic import is treated as synchronous (added to the `parents` map for cycle detection). However, `usesTopLevelAwait` is a module-level flag — it doesn't distinguish between dynamic imports at the top level vs. inside functions.

## Fix

Added an `isTopLevelImport()` helper function that checks whether an ImportExpression is nested inside any function (FunctionDeclaration, FunctionExpression, or ArrowFunctionExpression). Dynamic imports are only treated as synchronous if:
1. The ImportExpression is at the top level (not inside any function), AND
2. The module uses top-level await

Dynamic imports inside functions are always async, regardless of the module's TLA status.

## Test

Added `test/form/samples/dynamic-import-tla-no-spurious-cycle/` which reproduces the issue from #6154. Without the fix, rollup produces a `CIRCULAR_DEPENDENCY` warning. With the fix, no warning is produced.

Also verified that the existing `cycles-dependency-with-TLA-await-import` test still passes, confirming that legitimate TLA cycles are still detected.

### DIFF
--- a/src/utils/executionOrder.ts
+++ b/src/utils/executionOrder.ts
@@ -19,6 +19,28 @@ export function sortByExecutionOrder(units: OrderedExecutionUnit[]): void {
 // modules are executed before or after which other modules. THen the chunking
 // would need to take care that in each chunk, all modules are always executed
 // in the same sequence.
+
+// Check if an ImportExpression is at the top level of a module
+// (not nested inside any function)
+function isTopLevelImport(node: { parent: any }): boolean {
+	let parent = node.parent;
+	while (parent) {
+		// Check if parent is a function-like node by checking the type string
+		// This covers FunctionDeclaration, FunctionExpression, ArrowFunctionExpression,
+		// and class methods
+		const type = parent.type as string;
+		if (
+			type === 'FunctionDeclaration' ||
+			type === 'FunctionExpression' ||
+			type === 'ArrowFunctionExpression'
+		) {
+			return false;
+		}
+		parent = parent.parent;
+	}
+	return true;
+}
+
 export function analyseModuleExecution(entryModules: readonly Module[]): {
 	cyclePaths: string[][];
 	orderedModules: Module[];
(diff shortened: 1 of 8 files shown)


TrickyPi:
Thanks for looking into #6154. The false cycle reported there is a real issue, but I do not think checking whether the import expression is lexically inside a function is sufficient.

The module-level `usesTopLevelAwait` check was intentional. In #6121, we deliberately made the conservative assumption that a module containing top-level await may await any dynamic import within that module. This is necessary because a dynamic import inside a function can still be awaited during module evaluation:

```js
async function load() {
	return import('./dependency.js');
}

await load();
```

With the proposed `isTopLevelImport` check, this import would be treated as an ordinary dynamic import even though it is part of the top-level-await execution path.

We already have the `top-level-await-import-3` test, which specifically covers a dynamic import wrapped in a function and invoked by top-level await. The existing `cycles-dependency-with-TLA-await-import` test mentioned in this PR only covers a direct `await import()`, so it does not protect against this case.

Also, this PR changes only `executionOrder.ts`, while `includedTopLevelAwaitingDynamicImporters` and the chunk-assignment logic still use the module-level TLA assumption. This would leave execution-order analysis and chunk assignment using different definitions.

I agree that #6154 exposes an undesirable false positive. However, fixing it probably requires analysing whether a function is reachable from top-level-await evaluation, or providing another explicit way to refine or opt out of the conservative assumption. Function nesting alone can introduce false negatives and potentially reintroduce the ordering or deadlock issues fixed by #6121.

PR #6482, opened 2026-08-20
Remove unused rendered module sources map

yoominho91:
This PR contains:

- [ ] bugfix
- [ ] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

None.

### Description

`renderedModuleSources` is populated for every non-empty rendered module but is never read. The rendered sources are already added to the chunk's `MagicStringBundle`, so this removes unused bookkeeping without changing output. It avoids allocating an unused `Map` and inserting an entry for every non-empty rendered module.

No test was added because this is an internal, behavior-preserving deletion. Rollup's output suite exercises the affected rendering path.

Validation:

- `npm run update:js`
- `npm run lint:js:nofix`
- `env -u NO_COLOR TERM=xterm npm run test:only` (5,442 passing)

### DIFF
--- a/src/Chunk.ts
+++ b/src/Chunk.ts
@@ -1340,7 +1340,6 @@ export default class Chunk {
 		const usedModules: Module[] = [];
 		let hoistedSource = '';
 		const accessedGlobals = new Set<string>();
-		const renderedModuleSources = new Map<Module, MagicString>();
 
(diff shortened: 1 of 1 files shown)


lukastaegert (approved):
Thanks, good catch!

PR #6490, opened 2026-08-28
Revert "feat: improve function return value tracking (#6065)"

### LINKED ISSUE #6487: 4.63.0 - Out of memory error - vite 4.3.9, vue 3.4.0
### Rollup Version

4.63.0

Linux - Debian & Ubuntu

v20.20.2

### Link To Reproduction

This probably only happens because of something in my repository. Bumping to 4.63.0 caused builds to fail with OOM. Pinning to 4.62.5 works. I am leaving this information, knowing that you will close it anyway for your reference.

### Expected Behaviour

Build completes

### Actual Behaviour

Out of memory

### LINKED ISSUE #6488: 4.63.0 tree-shakes side-effectful modules out of commonjs-wrapped rxjs — awaited promise never settles
### Rollup Version

4.63.0 (regression from 4.62.5)

Linux (also observed on GitHub Actions ubuntu runners)

22.14.0

### Link To Reproduction

https://gist.github.com/excavador/e7a862dba1a9ac132c9ba28ee2f5bba7

Three small files (package.json / entry.js / vite.config.js); rollup version is selected via npm `overrides`. `npm install && npx vite build && node dist/index.js`.

### Expected Behaviour

The entry

```js
import { of, lastValueFrom } from 'rxjs';
import { map } from 'rxjs/operators';

const value = await lastValueFrom(of(1).pipe(map((x) => x + 1)));
console.log('RESULT:', value);
```

bundled by vite (lib mode, rxjs pulled in through @rollup/plugin-commonjs) prints `RESULT: 2` and exits 0 — which is what happens with rollup 4.62.5 via the same pipeline.

### Actual Behaviour

With 4.63.0 the awaited promise never settles: node prints `Warning: Detected unsettled top-level await` at the `lastValueFrom` line and exits with code 13.

Diffing the sourcemap `sources` arrays between the 4.62.5 and 4.63.0 builds, 4.63.0 removes these side-effectful rxjs-internal modules that 4.62.5 keeps:

- `rxjs/dist/cjs/internal/util/errorContext.js`
- `rxjs/dist/cjs/internal/NotificationFactories.js`
- `rxjs/dist/cjs/internal/scheduler/timeoutProvider.js`

Everything else in the bundle is identical modulo chunk naming. The repro behaves the same with `build.minify: false`, so esbuild is not involved — this is the tree-shake decision itself. Given the 4.63.0 changelog, #6065 ("improve function return value tracking") looks like the most likely origin, but I have not bisected the commits.

Real-world impact: NestJS routes every request through rxjs, so server bundles built this way boot normally, register routes, and then hang on every request (including health checks) with no error — it presents as an infrastructure failure rather than a bundler one, which made this quite expensive to track down.

lukastaegert:
This PR contains:

- [ ] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [x] other

Are tests included?

- [x] no

Breaking Changes?

- [x] no

- resolves #6488
- resolves #6487

### Description
Considering the unknown nature of #6488 and #6487, this reverts #6065 for now until the issue is better understood.


lukastaegert:
cc @cyyynthia

PR #6491, opened 2026-08-28
Fix return value tracking

### LINKED ISSUE #5063: tree-shaking will not execute recursively.

lukastaegert:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

- reapplies #6065
- resolves #5063

### Description
This is an attempt at fixing the issues from #6488.


TrickyPi:
Hi Lukas, since the OOM report did not include a reproduction, I asked Codex to investigate possible causes. It found a small test case where chained calls cause nested `MultiExpression` values to grow exponentially and eventually exhaust memory.

I opened https://github.com/rollup/rollup/pull/6493 with a fix and regression tests. This may be related to the reported OOM, although I cannot confirm that it is the same issue without the original reproduction.

cyyynthia:
EDIT: seems like you also did come to this repro; I should've read the diff before commenting 😅

---

I reduced the reproduction based on rxjs to this small snippet that can be used to create a test.

```js
var a = {}
var requiredA = false
function requireA() {
  if (requiredA) return a
  requiredA = true

  a.fn = function (cb) {
    cb()
  }

  return a
}

var a_1 = /* @__PURE__ */ requireA()
a_1.fn(function () {
  console.log('doing something')
})
```

Interesting findings:
- Removing the pure annotation solves the issue. That seems to indicate that the problem stems from a missing deopt somewhere.
- Removing the `if` gate in the function also solves this issue. My guts tell me that this correlates with deopts being skipped/omitted.

I'm not sure this is the only issue as there's more things stripped away from the bundle, but these seem to be related to configuration flags now being walked by Rollup thanks to its better return value tracking allowing it to eliminate unused code-paths.

lukastaegert:
Thanks for looking into this!

@cyyynthia So the problem with pure annotated functions is that they skip their "hasEffects" call. Usually, this is when the body of a function is analyzed the first time and return expressions are analyzed. In the previous version, return expressions that were not yet reached were ignored in MultiExpression.hasEffectsOnInteractionAtPath, hence the bug. In my current rework, we just drop the `isTreeshaken` check from MultiExpression and always analyze all.

@TrickyPi I had a look at your PR. I really like the idea of using CLI tests to create artificial memory bounds for single tests. However I must admit I found the 64 return limit somewhat arbitrary, and my main worry is that functions with many returns are not the biggest problem, but instead long chains of functions with multiple returns, which are not covered by your approach.
For now, I added some somewhat complex new infrastructure that tracks return statement composition across long getReturnExpressionWhenCalledAtPath chains, even across nested getLiteralValueAtPath queries. This should fix all exponential composition problems, but it is also not entirely straightforward and also seems to have a small performance impact. I am not yet 100% sure I would like to merge this, maybe I will iterate a little more on this.

cyyynthia:
> In my current rework, we just drop the isTreeshaken check from MultiExpression and always analyze all.

That makes sense, tho I'm concerned this might cause missed treeshaking opportunities, if only one branch would yield a side-effect on interaction but that branch's dead. While putting together an example of what I had in mind, it turns out that my original code also fails in these situations anyway...... 😭

Perhaps `hasEffect` should still execute regardless of whether the function is flagged pure or not, but have the result tossed if the flag exists. This will have a negative performance impact but guarantee the body of the function is walked if necessary. Alternatively, that could be done on-demand but that'd most certainly be trickier to do and to properly invalidate on subsequent treeshake passes...

Here's the test cases I came up with:

Input code:
```js
test1: {
  function test() {
    if (false) return () => 1
    return () => (console.log('side effect'), 2)
  }

  const t = /** @__PURE__ */ test()
  if (t() === 2) console.log('kept'); else console.log('removed')
}

test2: {
  function test() {
    if (true) return () => 1
    return () => (console.log('side effect'), 2)
  }

  const t = /** @__PURE__ */ test()
  if (t() === 1) console.log('kept'); else console.log('removed')
}

test3: {
  function test() {
    if (false) return undefined
    return {}
  }

  function testImplicitPure() {
    test().readKey
  }

  if (testImplicitPure()) console.log('removed'); else console.log('kept')
}

test4: {
  function test() {
    if (true) return undefined
    return {}
  }

  function testThrows() {
    test().readKey
  }

  if (testThrows()) console.log('removed'); else console.log('kept')
}
```

Expected output:
```js
{
  function test() {
    return () => (console.log('side effect'), 2)
  }

  const t = /** @__PURE__ */ test();
  if (t() === 2) console.log('kept');
}

{
  console.log('kept');
}

{

  console.log('kept');
}

{
  function test() {
    return undefined
  }

  function testThrows() {
    test().readKey;
  }

  if (testThrows()) ; else console.log('kept');
}
```

TrickyPi:
> @TrickyPi I had a look at your PR. I really like the idea of using CLI tests to create artificial memory bounds for single tests. However I must admit I found the 64 return limit somewhat arbitrary, and my main worry is that functions with many returns are not the biggest problem, but instead long chains of functions with multiple returns, which are not covered by your approach.
> For now, I added some somewhat complex new infrastructure that tracks return statement composition across long getReturnExpressionWhenCalledAtPath chains, even across nested getLiteralValueAtPath queries. This should fix all exponential composition problems, but it is also not entirely straightforward and also seems to have a small performance impact. I am not yet 100% sure I would like to merge this, maybe I will iterate a little more on this.

Thanks for the explanation. I’m glad the idea of setting a memory limit for individual CLI tests could be useful, and I hope it helps with the tests for your eventual solution. Since you are working on a more general approach here, I’ll close #6493 in favor of this PR.

PR #6495, opened 2026-08-29
docs: add an ESM and CommonJS comparison to the introduction

MuslumYilmaz:
## Summary
Add one conceptual companion link between “The Why” and “Tree-Shaking”.

## Why
The introduction names ESM and Rollup’s CommonJS output but does not explain the semantic differences that affect static analysis and interoperability. The linked explanation provides context for why ESM is easier to tree-shake and why mixed-module packaging needs care.

## Disclosure
I’m affiliated with FrontendAtlas, which publishes the linked free resource.

## Validation
Documentation-only change; the rendered link and live destination were checked. Project CI can validate formatting and links.


lukastaegert (changes requested):
Thanks for having a look. While I agree that good external links can improve documentation, an "interview prep" page feels off to me. Most people who might want to learn more are probably not hoping to prepare for an interview. A good blog post or spec/documentation page might be more appropriate IMO.

PR #6500, opened 2026-08-31
Fix chunk hashes when sourcemaps add comments

### LINKED ISSUE #6499: [hash] does not change when output.sourcemap changes the final chunk content
### Rollup Version

4.62.5

macOS/Chrome 151

_No response_

### Link To Reproduction

https://stackblitz.com/edit/rollup-repro-qjhsxzpu?file=rollup.config.js,src%2Fmain.js,src%2Fchunks%2FchunkA.js,package.json

### Expected Behaviour

Changing output.sourcemap from true to false changes the final emitted js content because the sourceMappingURL comment is removed.
I would expect the chunk [hash] to change when the final emitted chunk content changes

### Actual Behaviour

A minimal reproduction is provided in the StackBlitz link above. It explains that changing only output.sourcemap changes the emitted JavaScript content but does not change the [hash] in the chunk filename.

For example when building with:
output: {
  chunkFileNames: 'chunks/[name]-[hash].js',
  sourcemap: true
}

Rollup generates, for example:
chunks/chunk-ABC123.js
chunks/chunk-ABC123.js.map

and the emitted JavaScript contains:
//# sourceMappingURL=chunk-ABC123.js.map

When I then change only:
sourcemap: false

Rollup correctly regenerates the JavaScript and removes the sourceMappingURL comment.

However, the generated filename remains:
chunks/chunk-ABC123.js

So the final content has changed, but the [hash] remains the same.

This becomes a problem in our system because we rely on Rollup's [hash] in the filename as a cache-busting mechanism for browser-cached chunks.

After the second build, the browser still have the previous chunks/chunk-ABC123.js cached. Which contains:
//# sourceMappingURL=chunk-ABC123.js.map

As a result, React DevTools continues attempting to fetch the .map files, even though the .map file no longer exists on disk, resulting in repeated 404 requests.

So the issue is that changing sourcemap changes the final emitted JavaScript content, but does not change the [hash] in the chunk filename and because we rely on that hash as a browser cache-busting mechanism, the browser still use the previous version of the chunk.

wangxpych:
This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [x] yes
- [ ] no

Breaking Changes?

- [ ] yes
- [x] no

- resolves #6499

### Description

Chunk content hashes were calculated before Rollup appended the sourceMappingURL comment. As a result, enabling a sourcemap could change the emitted JavaScript bytes without changing a hashed chunk file name.

Include the generated sourcemap comment in the chunk hash input, after resolving the sourcemap hash placeholder and normalizing chunk placeholders. Hidden sourcemaps still produce the same chunk hash because they do not append a comment.

The new file-hash regression test fails on the previous implementation when sourcemaps are toggled. Existing sourcemap filename snapshots are updated for the corrected hashes.

Validation:

- npm run build:js:node
- npm run lint:js:nofix
- file-hashes test suite on Node.js 24.12.0 (17 passing)
- chunking-form test suite on Node.js 24.12.0 (1,156 passing)
- sourcemaps test suite on Node.js 24.12.0 (177 passing)


wangxpych:
Closing because I’m stepping back from maintaining this contribution. Thanks for your time.

emmanuelik0147-rgb:
Link To Reproduction
https://stackblitz.com/edit/rollup-repro-qjhsxzpu?file=rollup.config.js,src%2Fmain.js,src%2Fchunks%2FchunkA.js,package.json

Expected Behaviour

emmanuelik0147-rgb:
Link To Reproduction
https://stackblitz.com/edit/rollup-repro-qjhsxzpu?file=rollup.config.js,src%2Fmain.js,src%2Fchunks%2FchunkA.js,package.json

Expected Behaviour

PR #6502, opened 2026-09-03
Escape quotes and backslashes in import attribute values

dylanpulver:
This PR contains:

- [x] bugfix

Are tests included?

- [x] yes

Breaking Changes?

- [x] no

### Description

`formatAttributes` (`src/ExternalChunk.ts:70`) interpolates the attribute value straight into a single-quoted literal. With `format: 'es'`, `externalImportAttributes: true`, `importAttributesKey: 'with'`:

- `it's` → `with { type: 'it's' }`, a SyntaxError in the bundle
- `back\slash` → `with { type: 'back\slash' }`, parses but the value becomes `backslash`
- a newline → a literal newline in the literal, SyntaxError

Plugins are the realistic vector: `resolveId` returning `attributes: { integrity: "sha256-o'Brien+/=" }` emits an unparseable bundle.

Both neighbours already guard. `getImportPath()` ten lines up wraps in `escapeId()`, and the key half of the same pair goes through `stringifyObjectKeyIfNeeded`. Only the value was raw, so this reuses `escapeId` rather than adding a second escaper.

`git log -S` traces the line to 33d8c403 "[v3.0] Basic support for import assertions (#4646)". No existing sample covers a quote, backslash or newline in a value.

`mocha test/test.js`: 5445 passing, 5444 on master — the delta is the new test. Reverting `ExternalChunk.ts` fails it; escaping only the apostrophe still fails the backslash line.

Not fixed here: `escapeId` renders a newline as a line continuation, so such a value now emits valid code but loses the newline, as ids already do.

Drafted with Claude Opus 5 (`claude-opus-5`); every claim was executed.

### DIFF
--- a/src/ExternalChunk.ts
+++ b/src/ExternalChunk.ts
@@ -68,7 +68,7 @@ export function formatAttributes(
 		return null;
 	}
 	const assertionEntries: [key: string, value: string][] = Object.entries(attributes).map(
-		([key, value]) => [key, `'${value}'`]
+		([key, value]) => [key, `'${escapeId(value)}'`]
 	);
 	if (assertionEntries.length > 0) {
 		return getObject(assertionEntries, { lineBreakIndent: null });
--- a/test/form/samples/import-attributes/escapes-attribute-values/_config.js
+++ b/test/form/samples/import-attributes/escapes-attribute-values/_config.js
@@ -1,0 +1,9 @@
+module.exports = defineTest({
+	description: 'escapes quotes and backslashes in import attribute values',
+	options: {
+		external: () => true,
+		output: {
+			importAttributesKey: 'with'
+		}
+	}
+});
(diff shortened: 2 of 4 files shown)


src/ExternalChunk.ts
```diff
@@ -68,7 +68,7 @@ export function formatAttributes(
 		return null;
 	}
 	const assertionEntries: [key: string, value: string][] = Object.entries(attributes).map(
-		([key, value]) => [key, `'${value}'`]
+		([key, value]) => [key, `'${escapeId(value)}'`]
```
lukastaegert: `escapeId` is a helper that is meant for file names and paths. In this context, my gut feeling is that `JSON.stringify` would be the more straightforward choice which also does not have many edge cases that `escapeId` has. Most notably, `escapeId` escapes line-breaks in a way that would actually swallow the line break rather than preserving it correctly. What do you think?


```suggestion
		([key, value]) => [key, JSON.stringify(value)]
```

lukastaegert (changes requested):
Hi, this is a very good catch, thank you for spotting it. See my comment, though.

src/ExternalChunk.ts
lukastaegert: I performed the change now, from my side this is fine

lukastaegert (approved):

PR #6506, opened 2026-09-05
fix(watch): await the rerun so listener failures surface as ERROR events

yoominho91:
### What

`Watcher.invalidate()` schedules a rebuild, clears the listeners for the current run, and starts the next one. That call is not awaited, so anything the run rejects with escapes the surrounding `try`/`catch`: the ERROR and END events the catch exists to emit are never sent, and the rejection surfaces as an unhandled rejection instead.

### Steps to reproduce

An `event` listener that throws during a rebuild is enough. Against 4.63.1:

```js
const watcher = watch({ input: 'src/main.js', output: { dir: 'out', format: 'es' } });
let starts = 0;
watcher.on('event', event => {
  if (event.code === 'BUNDLE_END') event.result.close();
  if (event.code === 'BUNDLE_START' && ++starts === 2) throw new Error('handler blew up');
});
// then touch src/main.js
```

Before: `UNHANDLED REJECTION: handler blew up`, and no ERROR event arrives.
After: the ERROR event is emitted with that error and the watcher carries on.

The throw travels out of the `BUNDLE_START` emit in `Task.run` — which is outside that method's own `try` — up through `Watcher.run`, past the catch that was meant to handle it.

### Change

One `await` on the rerun, which puts the failure back inside the existing catch.

### Tests

Adds `emits an error event if a listener throws while re-running` to `test/watch/index.js`. It times out waiting for the ERROR event without the change and passes with it.

Full `mocha test/test.js`: 5442 passing. The single `plugin-hook-filters` failure also fails on unmodified `master` in my environment (I built with a prebuilt native binary rather than a local Rust build), so it is unrelated to this change.

### DIFF
--- a/src/watch/watch.ts
+++ b/src/watch/watch.ts
@@ -95,7 +95,7 @@ export class Watcher {
 				this.invalidatedIds.clear();
 				await this.emitter.emit('restart');
 				this.emitter.removeListenersForCurrentRun();
-				this.run();
+				await this.run();
 			} catch (error: any) {
 				this.invalidatedIds.clear();
 				await this.emitter.emit('event', {
--- a/test/watch/index.js
+++ b/test/watch/index.js
@@ -143,6 +143,44 @@ describe('rollup.watch', function () {
 		assert.deepStrictEqual([...events], ['START', 'BUNDLE_START', 'BUNDLE_END', 'END', undefined]);
 	});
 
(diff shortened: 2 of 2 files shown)


lukastaegert (approved):
Thanks, I added some additional improvements on top, looks good to me now.

